PR #4, opened 2017-05-09
Add support for only units on the NumberWithUnit extractors

ejadib:


Microsoft.Recognizers.Text.NumberWithUnit.Tests/TestNumberWithUnitChinese.cs
```diff
@@ -16,10 +15,20 @@ private void BasicTest(IModel model, string source, string value)
...
+            var results = model.Parse(source);
+            Assert.AreEqual(values.Length, results.Count);
+            var resultsValues = results.Select(x => GetStringValue(x)).ToArray();
+            CollectionAssert.AreEqual(values, resultsValues);
+        }
+
+        private string GetStringValue(ModelResult source)
+        {
+            object value, unit;
+            source.Resolution.TryGetValue(nameof(value), out value);
```
chrimc62: Do you want to do TryGetValue here?  Wouldn't you want it to tell you if not present or do you want the null further downstream?

PR #6, opened 2017-05-18
added french lang support

matthewshim-ms:
needs code review


Microsoft.Recognizers.Text.Number/French/Parsers/FrenchNumberParserConfiguration.cs
```diff
@@ -0,0 +1,250 @@
...
+using System.Collections.Immutable;
+using System.Globalization;
+using System.Text.RegularExpressions;
+using System.Text;
+
+namespace Microsoft.Recognizers.Text.Number.French.Parsers
+{
+    public class FrenchNumberParserConfiguration : INumberParserConfiguration
+    {
+        public FrenchNumberParserConfiguration(): this(new CultureInfo("fr-FR")) { }
```
ejadib: Can you update this to use the culture from https://github.com/Microsoft/Recognizers-Text/blob/master/Microsoft.Recognizers.Text.Number/Utilities/Culture.cs? Add French there, and update the constructor as [here](https://github.com/Microsoft/Recognizers-Text/blob/master/Microsoft.Recognizers.Text.Number/English/Parsers/EnglishNumberParserConfiguration.cs#L13)

matthewshim-ms: added as per request

PR #12, opened 2017-06-05
Update namespaces

ejadib:
* Removed suffixes like Parsers/Extractors/Models/Utilities

### DIFF
--- a/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/DateExtractorChs.cs
+++ b/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/DateExtractorChs.cs
@@ -1,11 +1,9 @@
 ﻿using System.Collections.Generic;
 using System.Text.RegularExpressions;
-using Microsoft.Recognizers.Text.Number.Chinese.Extractors;
-using Microsoft.Recognizers.Text.Number.Chinese.Parsers;
-using Microsoft.Recognizers.Text.Number.Parsers;
-using Microsoft.Recognizers.Text.DateTime.Utilities;
+using Microsoft.Recognizers.Text.Number;
+using Microsoft.Recognizers.Text.Number.Chinese;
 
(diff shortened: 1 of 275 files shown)


Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/DateExtractorChs.cs
```diff
@@ -1,220 +1,218 @@
...
-        }
-    }
+﻿using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Text.Number;
+using Microsoft.Recognizers.Text.Number.Chinese;
+
+namespace Microsoft.Recognizers.Text.DateTime.Chinese
+{
+    public class DateExtractorChs : IExtractor
```
chrimc62: This is probably a separate issue, but in general classes in a namespace should not reflect the namespace itself.  In this case adding CHS to the end of the classes.

Microsoft.Recognizers.Text.DateTime/Spanish/Extractors/SpanishDateExtractorConfiguration.cs
```diff
@@ -1,151 +1,149 @@
-﻿using System.Text.RegularExpressions;
-using System.Collections.Generic;
-using Microsoft.Recognizers.Text.Number.Spanish.Extractors;
-using Microsoft.Recognizers.Text.Number.Parsers;
-using Microsoft.Recognizers.Text.Number.Spanish.Parsers;
-using Microsoft.Recognizers.Text.DateTime.Extractors;
-
-namespace Microsoft.Recognizers.Text.DateTime.Spanish.Extractors
```
chrimc62: It is kind of annoying there are all these large diffs.  Was it only space differences or something?

ejadib: Probably because due to line endings :(

PR #22, opened 2017-06-15
Change the end of date period to inclusive

rubio41:
As discussed with Borje, we should inform the user for the change before we actually make the change. Before that we should set the bool "InclusiveEndPeriod" in BaseDatePeriodParser to false.

### DIFF
--- a/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
+++ b/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
@@ -8,10 +8,12 @@ namespace Microsoft.Recognizers.Text.DateTime
     public class BaseDatePeriodParser : IDateTimeParser
     {
         public static readonly string ParserName = Constants.SYS_DATETIME_DATEPERIOD; //"DatePeriod";
-        
+
         private static readonly Calendar _cal = DateTimeFormatInfo.InvariantInfo.Calendar;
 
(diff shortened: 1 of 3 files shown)


ejadib:
@rubio41 since the BaseDatePeriodParser is also used by the Spanish recognizer (see [here](https://github.com/Microsoft/Recognizers-Text/blob/master/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishMergedParserConfiguration.cs#L27-L28)), could you also check that the Spanish tests keep passing?

Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -829,17 +848,17 @@ private DTParseResult ParseSeason(string text, DateObject referenceDate)
         private static DateObject ComputeDate(int cadinal, int weekday, int month, int year)
```
tellarin: Argument should be cardinal.

tellarin:
@rubio41, in this PR, please leave 'InclusiveEndPeriod' as false, as you've mentioned. Once the new spec is out and users are warned, we can change it to true.

rubio41:
@ejadib I will change the code in Spanish recognizer accordingly. But I cannot understand Spanish. Therefore, you should config the configuration files and add some unit tests accordingly if there are. @tellarin , is there a standard way for the language related changes?

PR #28, opened 2017-06-27
Added text project and updated number regocnizer helper class

JasonSowers:
needs to go after #27 added NumberRecognizer class with an update for French.


Microsoft.Recognizers.Text.Number/BaseNumberRecognizer.cs
```diff
@@ -1,27 +1,28 @@
...
-                }
-            }
-
-            var model = ModelInstances[culture][typeof(TModel)];
-
-            return model;
-        }
-    }
-}
+using System;
```
yiylin: In Microsoft/Recognizers-Text, there is a text project. Most of your change is deleting the old code and adding new code which is the same as the old code. I think you can abandon this pr.

JasonSowers: I was working with @mattshim for some reason his project did not have the text project.  The part that actually changed is in Microsoft.Recognizers.Text.Number/NumberRecognizer.cs Added the French to it.  For some reason it is not getting captured in the pull request, I'll close this one and resubmit

PR #36, opened 2017-07-05
Fix last half year bug

rubio41:


Microsoft.Recognizers.Text.DateTime.Tests/English/TestDurationExtractor.cs
```diff
@@ -44,6 +44,9 @@ public void TestDurationExtract()
 
             BasicTest("I'll leave for an hour", 15, 7);
             BasicTest("I'll leave for a year", 15, 6);
+
+            BasicTest("half year", 0, 9);
+            BasicTest("half an year", 0, 12);
```
yiylin: Can you add the Chinese unit test and add TODO in the Spanish unit test?

Microsoft.Recognizers.Text.DateTime.Tests/English/TestDurationParser.cs
```diff
@@ -46,6 +46,10 @@ public void TestDurationParse()
             BasicTest("I'll leave for all year", 31536000, "P1Y");
 
             BasicTest("I'll leave for an hour", 3600, "PT1H");
+
+            BasicTest("half year", 15768000, "P0.5Y");
+            BasicTest("half an year", 15768000, "P0.5Y");
+
```
yiylin: Can you add the Chinese unit test and add TODO in the Spanish unit test?

rubio41: the Chinese extractor and parser are not following the base class. It needs support for number extractor for "半". I will file a bug in VSO.

yiylin: ok, Then I will merge it firstly.

PR #45, opened 2017-07-11
Refactor the model register logic

yiylin:
In the old version, NumberRecognizer and NumberWithUnitRecognizer share the same base class. In their static constructor method, they set the same parameter. So I can not use NumberRecognizer and NumberWithUnitRecognizer at the same time. 

Now I change to  every Recognizer has its own ModelContainer to register model


Microsoft.Recognizers.Text/ModelContainer.cs
```diff
@@ -0,0 +1,87 @@
...
+
+            if (ret)
+            {
+                model = ModelInstances[key];
+            }
+
+            return ret;
+        }
+
+        public bool ContainModel<TModel>(string culture, bool fallbackToDefaultCulture = true)
```
tellarin: ContainsModel

PR #50, opened 2017-07-22
support more decimal and thousands separators;

linzijia07:


Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
```diff
@@ -74,6 +74,47 @@ public virtual List<ExtractResult> Extract(string source)
 
             return result;
         }
+
+        public Regex AddArabicNumberRegex(ArabicType type, string placeholder = @"\D|\b")
```
yiylin: Renaming this method to GenerateArabicNumberRegex is better. And this method should be protected

Microsoft.Recognizers.Text.Number.Tests/TestDecimalAndThousandsSeparators.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Globalization;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Collections.Immutable;
+using Microsoft.Recognizers.Text;
+using Microsoft.Recognizers.Text.Number;
+
+namespace Microsoft.Recognizers.Text.Number.Tests
+{
+    public class ArabicNumberTestConfiguration : INumberParserConfiguration
+    {
```
yiylin: It seems that we do not need to create a new configuration.We can use https://github.com/Microsoft/Recognizers-Text/blob/master/Microsoft.Recognizers.Text.Number/English/Parsers/EnglishNumberParserConfiguration.cs

Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
linzijia07: done

Microsoft.Recognizers.Text.Number.Tests/TestDecimalAndThousandsSeparators.cs
linzijia07: We need this new configuration to go through all potential combinations of decimal separator and non-decimal separator. EnglishNumberParserConfiguration just uses one combination.

PR #59, opened 2017-08-09
[Data Files] Extract RegExes and constant strings to YAML files for reuse in future ports

pcostantini:
The idea of this PR is to extract all common regular expressions, strings and dictionaries to a re-usable format that can be shared among .NET (C#) and future Recognizer ports (e.g.: JavaScript, Python, etc.)

1. YAML files with Regular Expressions, dictionaries and other string constants
2. T4 templates to generate CS constant files based on YAML
3. Updates to Recognizers and Parsers to use constant files

NOTE: This only covers the English language. More PRs coming soon for each supported language.


Common/Common-DateTime.yaml
```diff
@@ -0,0 +1,51 @@
+HourRegex: !simpleRegex
```
tellarin: Nit: I'd prefer to call this directory Resources/Patterns or something like it. Common implies common across languages and this is actually common resources to be compiled/transpiled into code in different programming languages.

Microsoft.Recognizers.Resources.Common/Microsoft.Recognizers.Resources.Common.csproj
```diff
@@ -0,0 +1,57 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
tellarin: Resources.Common here seems good as project name.

Microsoft.Recognizers.Resources.Common/YamlParser.cs
```diff
@@ -0,0 +1,101 @@
...
+using YamlDotNet.Core.Events;
+using YamlDotNet.RepresentationModel;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NodeDeserializers;
+
+namespace Microsoft.Recognizers.Resources.Common
+{
+    public class YamlParser
+    {
+        YamlStream yamlStream;
```
tellarin: This is not used anywhere?

Microsoft.Recognizers.Resources.Common/YamlParser.cs
```diff
@@ -0,0 +1,101 @@
...
+        Deserializer yamlDeserializer;
```
tellarin: This could be made private readonly.

Common/NumericWithUnit-English.yaml
```diff
@@ -0,0 +1,619 @@
...
+    - sek
+    - sen
+    - sol
+    - sos
+    - std
+    - try
+    - yer
+    - yen
+#DimensionExtractorConfiguration
+DimensionSuffixList: !dictionary
```
tellarin: This should be InformationSuffixList

Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
```diff
@@ -75,11 +76,11 @@ public virtual List<ExtractResult> Extract(string source)
             return result;
         }
 
-        protected Regex GenerateArabicNumberRegex(ArabicType type, string placeholder = @"\D|\b")
+        protected Regex GenerateArabicNumberRegex(ArabicType type, string placeholder = CommonNumeric.PlaceHolderDefault)
         {
             Regex addedRegex = null;
-            string integerTemplate = "(((?<!\\d+\\s*)-\\s*)|((?<=\\b)(?<!(\\d+\\.|\\d+,))))\\d{{1,3}}({0}\\d{{3}})+" + $@"(?={placeholder})";
-            string doubleTemplate = "(((?<!\\d+\\s*)-\\s*)|((?<=\\b)(?<!\\d+\\.|\\d+,)))\\d{{1,3}}({0}\\d{{3}})+{1}\\d+" + $@"(?={placeholder})";
+            string integerTemplate = CommonNumeric.IntegerTemplateRegex + $@"(?={placeholder})";
```
tellarin: Shouldn't this regex/templates also be exported to the common resources?

PR #74, opened 2017-08-24
Add datetime options

rubio41:
1. add datetime options for configuration
2. split the from to into two datetime points for DatetimeOptions.SplitFromTo only
3. support the number in text for hour minute (only for the minute be divided by 5)


Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishMergedExtractorConfiguration.cs
```diff
@@ -7,8 +7,14 @@ namespace Microsoft.Recognizers.Text.DateTime.English
 {
     public class EnglishMergedExtractorConfiguration : IMergedExtractorConfiguration
     {
-        public static readonly Regex BeforeRegex = new Regex(DateTimeDefinitions.BeforeRegex, RegexOptions.IgnoreCase | RegexOptions.Singleline);
```
tellarin: Please don't reformat regex code unless needed.

PR #80, opened 2017-09-04
fix 5 set related bugs reported by Mercury

rubio41:
- [x] 1. at 9am every Sunday, should be contacted
- [x] 2. every morning at 9am, the timex is TMO, should merge morning and 9am
- [x] 3. schedule a repeating event on mondays and wednesdays, this should not be a range, but two separate sets (related with the monday+s bug)
- [x] 4. on Mondays = every Monday
- [x] 5. 7 til 8am = 7 until 8am


Microsoft.Recognizers.Text.DateTime/Spanish/Extractors/SpanishDatePeriodExtractorConfiguration.cs
```diff
@@ -72,7 +72,7 @@ public class SpanishDatePeriodExtractorConfiguration : IDatePeriodExtractorConfi
         //TODO: modify it according to the related regex in English
         public static readonly Regex OneWordPeriodRegex =
             new Regex(
-                @"\b(((pr[oó]xim[oa]?|est[ea]|[uú]ltim[oa]?|en)\s+)?(?<month>Abril|Abr|Agosto|Ago|Diciembre|Dic|Enero|Ene|Febrero|Feb|Julio|Jul|Junio|Jun|Marzo|Mar|Mayo|May|Noviembre|Nov|Octubre|Oct|Septiembre|Setiembre|Sept|Set)|(?<=\b(del|de la|el|la)\s+)?(pr[oó]xim[oa](s)?|[uú]ltim[oa]?|est(e|a))?\s+(fin de semana|semana|mes|año)|fin de semana|(mes|años)? a la fecha)\b",
+                @"\b(((pr[oó]xim[oa]?|est[ea]|[uú]ltim[oa]?|en)\s+)?(?<month>Abril|Abr|Agosto|Ago|Diciembre|Dic|Enero|Ene|Febrero|Feb|Julio|Jul|Junio|Jun|Marzo|Mar|Mayo|May|Noviembre|Nov|Octubre|Oct|Septiembre|Setiembre|Sept|Set)|(?<=\b(del|de la|el|la)\s+)?(pr[oó]xim[oa](s)?|[uú]ltim[oa]?|est(e|a))\s+(fin de semana|semana|mes|año)|fin de semana|(mes|años)? a la fecha)\b",
```
tellarin: Why this change?

Patterns/English/English-DateTime.yaml
```diff
@@ -18,7 +18,7 @@ MonthNumRegex: !simpleRegex
 PeriodYearRegex: !simpleRegex
   def: \b(?<year>19\d{2}|20\d{2})\b
 WeekDayRegex: !simpleRegex
-  def: (?<weekday>Sunday|Monday|Tuesday|Wednesday|Thursday|Friday|Saturday|Mon|Tue|Wedn|Weds|Wed|Thurs|Thur|Thu|Fri|Sat|Sun)\b
+  def: \b(?<weekday>Sunday|Monday|Tuesday|Wednesday|Thursday|Friday|Saturday|Mon|Tue|Wedn|Weds|Wed|Thurs|Thur|Thu|Fri)\b
```
tellarin: Why remove Sat and Sun?

Patterns/English/English-DateTime.yaml
```diff
@@ -297,7 +297,9 @@ EachUnitRegex: !nestedRegex
   def: (?<each>(each|every)\s*{DurationUnitRegex})
   references: [ DurationUnitRegex ]
 EachPrefixRegex: !simpleRegex
-  def: (?<each>(each|every)\s*$)
+  def: \b(?<each>(each|(every(\s+other)?))\s*$)
```
tellarin: Where do you deal with the 'every other' case in resolution?

Microsoft.Recognizers.Text.DateTime/Spanish/Extractors/SpanishDatePeriodExtractorConfiguration.cs
rubio41: In English, this pattern is used to extract (next|past|...) (month|week|day...). We didn't want to extract single "month|week|day" but require a prefix. I made the changes in English, but not in Spanish. In addition, for the single "month|week|day", this pattern will output an extraction start with "\s" which is not expected.

Patterns/English/English-DateTime.yaml
rubio41: "Sat" and "Sun" are ambiguous words. It is not correct to extract that alone. I use a new regex for the SingleWeekdayRegex which deleted the Sat and Sun.

Patterns/English/English-DateTime.yaml
rubio41: The original plan is to add it later. Now, I treat every other week = biweekly.

Microsoft.Recognizers.Text.DateTime/Spanish/Extractors/SpanishDatePeriodExtractorConfiguration.cs
tellarin: OK

PR #81, opened 2017-09-04
Extract Portuguese Definitions

pcostantini:
Extract Regular Expressions, term lists and dictionaries into YAML for future re-use in future ports


pcostantini:
Waiting for folder re-structure PR. Will submit again later

PR #82, opened 2017-09-04
Folder structure per language

JuanAr:
Moved the .NET (C#) Recognizer solution to it own folder (`.NET`) in preparation for new language and platform implementation.
Updated T4 code to read YAML files from new relative location.
Tested and fixed NuGet package creation scripts (mainly, encoding issues)


.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
```diff
@@ -169,7 +169,7 @@ public static class DateTimeDefinitions
 		public const string ConnectorRegex = @"^(,|for|t|around)$";
 		public const string FromToRegex = @"\b(from).+(to)\b.+";
 		public const string SingleAmbiguousMonthRegex = @"^(the\s+)?(may|march)$";
-		public const string PrepositionSuffixRegex = @"\b(on|in|at|around|since|from|to)$";
```
yiylin: @rubio41 can you check this change?

rubio41: This only used for filtering. I didn't find any time related sentences in Google for "since May", "since the May", "since March", "since the March".

PR #83, opened 2017-09-04
TypeScript Recognizers

enzocanoo:
Typescript implementation of Text Recognizers with
- Numbers
- Numbers with Units
- DateTime

Including English Extractors/Parsers configuration and tests


JavaScript/src/dateTime/baseDuration.ts
```diff
@@ -0,0 +1,272 @@
...
+export class BaseDurationParser implements IDateTimeParser {
+    private readonly parserName = Constants.SYS_DATETIME_DURATION;
+    private readonly config: IDurationParserConfiguration;
+    
+    constructor(config: IDurationParserConfiguration) {
+        this.config = config;
+    }
+    
+    parse(extractorResult: ExtractResult, referenceDate?: Date): DateTimeParseResult | null {
+        if (!referenceDate) referenceDate = new Date();
```
tellarin: Nitpick: Add more blank lines throughout the code to improve readability.  Especially between ifs.

JavaScript/src/dateTime/baseDuration.ts
```diff
@@ -0,0 +1,272 @@
...
+        }
+        return result;
+    }
+    
+    private parseInExactNumberUnit(source: string): DateTimeResolutionResult {
+        let result = new DateTimeResolutionResult();
+        let match = RegExpUtility.getMatches(this.config.inExactNumberUnitRegex, source).pop();
+        if (!match) return result;
+        
+        let num = 3;
```
tellarin: Why is num fixed at 3 here? Should num be renamed to clarify what it means?

JavaScript/src/dateTime/baseHoliday.ts
```diff
@@ -0,0 +1,247 @@
...
+    abstract sanitizeHolidayToken(holiday: string): string;
+
+    constructor() {
+        this.variableHolidaysTimexDictionary = BaseDateTime.VariableHolidaysTimexDictionary;
+        this.holidayFuncDictionary = this.initHolidayFuncs();
+    }
+
+    protected initHolidayFuncs(): ReadonlyMap<string, (year: number) => Date> {
+        return new Map<string, (year: number) => Date>(
+            [
```
tellarin: We should think of a way to make this auto-generated in the future.
Can you add a TODO here?

JavaScript/src/dateTime/baseHoliday.ts
```diff
@@ -0,0 +1,247 @@
...
+                ["washingtonsbirthday", BaseHolidayParserConfiguration.WashingtonsBirthday],
+                ["canberra", BaseHolidayParserConfiguration.CanberraDay],
+                ["labour", BaseHolidayParserConfiguration.LabourDay],
+                ["columbus", BaseHolidayParserConfiguration.ColumbusDay],
+                ["memorial", BaseHolidayParserConfiguration.MemorialDay]
+            ]);
+    }
+
+    // All months are zero-based (-1)
+    private static MothersDay(year: number): Date { return new Date(year, 5 - 1, BaseHolidayParserConfiguration.getDay(year, 5 - 1, 1, DayOfWeek.Sunday)); }
```
tellarin: We should think of a way to make this auto-generated in the future.
Can you add a TODO here?

JavaScript/test/dateExtractor-english.test.js
```diff
@@ -0,0 +1,90 @@
+var describe = require('ava-spec').describe;
+var EnglishDateExtractorConfiguration = require('../compiled/dateTime/english/dateConfiguration').EnglishDateExtractorConfiguration;
+var BaseDateExtractor = require('../compiled/dateTime/baseDate').BaseDateExtractor;
+var Constants = require('../compiled/dateTime/constants').Constants;
+
+describe('Date Extractor', it => {
+    let extractor = new BaseDateExtractor(new EnglishDateExtractorConfiguration());
+
+    basicTest(it, extractor, "I'll go back on 15", 16, 2);
```
tellarin: We should also think of a way to put the unit test data for both C# and JS into data files, so unit tests don't need to be added to both and kept in sync manually.

Patterns/English/English-Numbers.yaml
```diff
@@ -23,19 +23,25 @@ PlaceHolderDefault: !simpleRegex
   def: \D|\b
 NumbersWithPlaceHolder: !paramsRegex
   def: (((?<!\d+\s*)-\s*)|(?<=\b))\d+(?!(\.\d+[a-zA-Z]))(?={placeholder})
+  def_js: (((?!\d+\s*)-\s*)|(?=\b))\d+(?!(\.\d+[a-zA-Z]))(?={placeholder})
```
tellarin: Is there an easy or suggested way to convert or test .net regexes to .js?
It would be nice to have a readme file in the repo that gives guidelines or pointers to new devs on how to do it and what to pay attention to.

Patterns/English/English-NumbersWithUnit.yaml
```diff
@@ -502,8 +502,10 @@ AmbiguousDimensionUnitList: !list
 #EnglishNumberWithUnitExtractorConfiguration
 BuildPrefix: !simpleRegex
   def: (?<=(\s|^|\W))
+  def_js: (?=(\s|^|\W)|(?!(\s|^|\W))\b)
 BuildSuffix: !simpleRegex
   def: (?=(\s|\W|$))
+  def_js: (?=(\s|\W|$))
```
tellarin: For the cases where the js regex is the same as the regular one, can't it just reference the value of the other, instead of duplicating the regex text?

PR #86, opened 2017-09-06
fix 3 bugs found from Mercury test data

rubio41:
- [x] for the 25th
- appear at the end of sentence
- [x] Thursday the 21st
- the date which above refer to is 9/21/2017
- both week day and day of month refer to a same date
- [x] 12 noon
- = noon


.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDateExtractor.cs
```diff
@@ -1,97 +1,142 @@
...
+        public void TestDateExtractForThe()
+        {
+            BasicTest("I went back for the 27", 12, 10);
+            BasicTest("I went back for the 27th", 12, 12);
+            BasicTest("I went back for the 27.", 12, 10);
+            BasicTest("I went back for the 27!", 12, 10);
+            BasicTest("I went back for the 27 .", 12, 10);
+            BasicTest("I went back for the 21st", 12, 12);
+            BasicTest("I went back for the 22nd", 12, 12);
+            BasicTest("I went back for the second", 12, 14);
```
tellarin: Can you please add a TestMethod for some negative cases?

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDateParser.cs
```diff
@@ -186,5 +248,33 @@ public void TestDateParseLuis()
...
+        public void TestDateParseForTheLuis()
+        {
+            BasicTest("I went back for the 27", "XXXX-XX-27");
+            BasicTest("I went back for the 27th", "XXXX-XX-27");
+            BasicTest("I went back for the 27.", "XXXX-XX-27");
+            BasicTest("I went back for the 27!", "XXXX-XX-27");
+            BasicTest("I went back for the 27 .", "XXXX-XX-27");
+            BasicTest("I went back for the 21st", "XXXX-XX-21");
+            BasicTest("I went back for the 22nd", "XXXX-XX-22");
+            BasicTest("I went back for the second", "XXXX-XX-02");
```
tellarin: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Extractors/SpanishDateExtractorConfiguration.cs
```diff
@@ -151,10 +162,38 @@ public SpanishDateExtractorConfiguration()
...
 
         Regex IDateExtractorConfiguration.MonthEnd => MonthEnd;
 
         Regex IDateExtractorConfiguration.DateUnitRegex => DateUnitRegex;
+
+        Regex IDateExtractorConfiguration.ForTheRegex => ForTheRegex;
+
+        Regex IDateExtractorConfiguration.WeekDayAndDayOfMothRegex => WeekDayAndDayOfMothRegex;
+
+        private static ImmutableDictionary<string, int> InitDayOfWeek()
```
tellarin: Why was this added here now?

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDateExtractor.cs
```diff
@@ -1,97 +1,142 @@
...
+                var date = new DateObject(referenceTime.Year, referenceTime.Month, dayOfMonth);
+                weekDay = date.DayOfWeek.ToString();
+            }
+
+            var sentence = "I went back " + weekDay;
+            return sentence;
+        }
+
+        [TestMethod]
+        public void TestDateExtract()
```
tellarin: Where are the TestMethods to cover cases like "Thursday the 21st"?
Please create separate TestMethods for these cases, don't add test to pre-existing TestMethods.

rubio41: They are in the TestDateExtractWeekDayAndDayOfMonth. It requires the WeekOfDay to be the same with the date mentioned. Otherwise, it should be two separate time mentions. Therefore, we use a function to generate the WeekOfDay when given a date.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDateParser.cs
rubio41: The parser will get no input if it is not extracted. Adding the negative cases in the extractors is enough.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDateExtractor.cs
rubio41: done

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Extractors/SpanishDateExtractorConfiguration.cs
rubio41: 1. The English DayOfWeek is configured in Yaml file but the Spanish are not.
2. For "the Friday the 8th", it has to check the 8th in this month is Friday or not. If yes, it will extract as a whole. If no, it will extract two entities "the Friday" and "the 8th".

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDateExtractor.cs
tellarin: Got it. Next time, please add an example in a code comment in the method. Just to make it easier to track what covers what.

PR #89, opened 2017-09-06
Fix bugs found in Mercury test data

rubio41:
- [x] tomorrow morning at 9 o'clock am
- [x] the month of September
- [x] August 1st 11 AM (previously 11 will be extracted as year and break the understanding of time)
- [x] 7pm on Sun
- [x] on the 17th
- [x] Christmas day
- [x] easter with no resolution
- [x] separate the WeekOfDay and the ordinary number if they are not matching


.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
```diff
@@ -50,7 +50,7 @@ public static class DateTimeDefinitions
 		public const string WeekOfRegex = @"(week)(\s*)(of)";
 		public const string MonthOfRegex = @"(month)(\s*)(of)";
 		public const string MonthRegex = @"(?<month>April|Apr|August|Aug|December|Dec|February|Feb|January|Jan|July|Jul|June|Jun|March|Mar|May|November|Nov|October|Oct|September|Sept|Sep)";
-		public const string DateYearRegex = @"(?<year>19\d{2}|20\d{2}|((9\d|0\d|1\d|2\d)(?!\s*\:)))";
+		public static readonly string DateYearRegex = $@"(?<year>19\d{{2}}|20\d{{2}}|((9\d|0\d|1\d|2\d)(?!(\s*((\:)|{AmDescRegex}|{PmDescRegex})))))";
```
tellarin: Why are AmDesc and PmDesc references in a regex called DateYear?

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestMergedParser.cs
```diff
@@ -37,6 +37,7 @@ public void BasicTestWithTwoResults(string text, string type1, string type2)
         [TestMethod]
         public void TestMergedParse()
         {
+            BasicTest("Set an appointment for Easter", Constants.SYS_DATETIME_DATE);
```
tellarin: Is there a test to make sure DateTime.MinValue is being filtered out from response? Please add one if not.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -257,13 +264,13 @@ private DateTimeResolutionResult ParseImplicitDate(string text, DateObject refer
...
                 int day = 0, month = referenceDate.Month, year = referenceDate.Year;
-                var dayStr = match.Groups["day"].Value.ToLower();
+                var dayStr = match.Groups["DayOfMonth"].Value.ToLower();
 
                 // create a extract result which content ordinal string of text
                 ExtractResult er = new ExtractResult();
                 er.Text = dayStr;
-                er.Start = match.Groups["day"].Index;
-                er.Length = match.Groups["day"].Length;
+                er.Start = match.Groups["DayOfMonth"].Index;
```
tellarin: Shouldn't this be a constants somewhere as it's related to some resource/data definition?

.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
rubio41: This is a looking behind assertion. In some cases we will extract two digits as year like 11/25/16, the 16 here equals to 2016. Using the assertion here to determine if is not connected with the am or pm.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestMergedParser.cs
rubio41: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
rubio41: now, it will be single number (27), ordinary number (27th), number in text (twenty seven), ordinary number in text (twenty seventh). The resource/data definition doesn't contain all of them.  Therefore, we use the number parser to get the value. Should we create a task to config the 31*4 numbers to the regex as well as the resource/date definition?

.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
tellarin: I see. Can you add this as a comment in the line immediately before this regex in the YAML file?
Comments in YAML start with #.
I think this is very relevant for people translating it to other languages to know.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
tellarin: Let's not change this now. But please open an issue to analyze this case again soon.

PR #94, opened 2017-09-08
Recognizers Chinese Definition files

pcostantini:
Extract Chinese expressions, constants and dictionaries into YAML files for re-use between platforms.
The PR includes all recognizers types: Number, NumberWithUnit and DateTime


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/DateTimeExtractorChs.cs
```diff
@@ -65,7 +61,7 @@ public List<Token> MergeDateAndTime(string text)
                 return ret;
             }
 
-            ers.Sort(delegate(ExtractResult er1, ExtractResult er2)
+            ers.Sort(delegate (ExtractResult er1, ExtractResult er2)
```
tellarin: This will conflict with the Sort change in PR #92.

pcostantini: Updated PR with master branch merge

PR #98, opened 2017-09-12
Spanish Definition files

pcostantini:
Extract Spanish Regular Expressions, constants and dictionaries into YAML files for re-use between platforms.
The PR includes all recognizers types: Number, NumberWithUnit and DateTime


.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishTimeParserConfiguration.cs
```diff
@@ -1,120 +1,121 @@
...
+                    deltaMin = int.Parse(minStr);
+                }
+                else
+                {
+                    minStr = match.Groups["deltaminnum"].Value.ToLower();
+                    Numbers.TryGetValue(minStr, out deltaMin);
+                }
+            }
+
+            if (trimedPrefix.EndsWith("pasadas") || trimedPrefix.EndsWith("pasados") ||
```
tellarin: These strings should also be moved to definitions, or become a regex there.

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishTimeParserConfiguration.cs
```diff
@@ -1,120 +1,121 @@
...
+                trimedPrefix.EndsWith("pasadas las") || trimedPrefix.EndsWith("pasados las") ||
+                trimedPrefix.EndsWith("pasadas de las") || trimedPrefix.EndsWith("pasados de las"))
+            {
+                //deltaMin it's positive
+            }
+            else if (trimedPrefix.EndsWith("para la") || trimedPrefix.EndsWith("para las") ||
```
tellarin: Same?

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishTimeParserConfiguration.cs
pcostantini: After all definitions extraction are merged, we'll create a new PR (for all languages) with these changes

PR #109, opened 2017-09-20
Fix negative lookbehind matches on javascript

enzocanoo:
Bug: Regexes with more than one negative lookbehind resolves different matches than the C# implementation.

Solution: Those regexes are evaluated in two steps, first obtaining the matches without the negative lookbehind and then dropping those that also match with the negative lookbehind groups.

### DIFF
--- a/JavaScript/src/utilities.ts
+++ b/JavaScript/src/utilities.ts
@@ -70,47 +70,43 @@ export class Match {
 
 export class RegExpUtility {
     static getMatches(regex: RegExp, source: string): Array<Match> {
-        if (source === 'I went back for the 27') {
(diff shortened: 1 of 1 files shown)


JavaScript/src/utilities.ts
```diff
@@ -70,11 +70,50 @@ export class Match {
 
 export class RegExpUtility {
     static getMatches(regex: RegExp, source: string): Array<Match> {
+        let rawRegex:string = (regex as any).xregexp.source;
```
tellarin: Can you add some comments in the code to explain the implemented logic?
It will help future people when they need to maintain/extend this.
Thanks.

PR #112, opened 2017-09-22
Move .NET Solution to Spec based approach for running unit-tests

pcostantini:
This PR contains the necessary work to migrate the current C# Test projects to a Spec based approach that can be re-used across platforms

The Spec files are automatically generated based on the assertions of the original Test projects, by running them


.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -0,0 +1,313 @@
...
+                    return new BaseDateTimePeriodExtractor(new EnglishDateTimePeriodExtractorConfiguration());
+                case "Duration":
+                    return new BaseDurationExtractor(new EnglishDurationExtractorConfiguration());
+                case "Holiday":
+                    return new BaseHolidayExtractor(new EnglishHolidayExtractorConfiguration());
+                case "Set":
+                    return new BaseSetExtractor(new EnglishSetExtractorConfiguration());
+                case "Merged":
+                    return new BaseMergedExtractor(new EnglishMergedExtractorConfiguration(), DateTimeOptions.None);
+                case "MergedSkipFromTo":
```
tellarin: These case values are now test-case-specific, right? Shouldn't we have something like a string-backed enum to keep them?
I'm not fond of the idea where devs needs to remember specific strings and if we decide to rename them, if will affect many different places.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -0,0 +1,313 @@
...
+                    return new BaseMergedExtractor(new EnglishMergedExtractorConfiguration(), DateTimeOptions.SkipFromToMerge);
+            }
+            return null;
```
tellarin: I'd rather throw an exception. It makes it easier to immediately see where the error came from. Otherwise a nullpointerexception could come from multiple places. Or null checks need to be spread throughout the test code.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    public static class TestUtils
+    {
+        public static string GetCulture(string source)
+        {
+            var langStr = source.Substring(source.LastIndexOf('_') + 1);
+            switch (langStr)
+            {
+                case "Chs":
```
tellarin: Same as above for "langStr".

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        public static bool EvaluateSpec(TestModel spec, out string message)
+        {
+            if (string.IsNullOrEmpty(spec.Input))
+            {
+                message = $"spec not found";
+                return true;
+            }
+            if (spec.IsNotSupported())
```
tellarin: Nitpick: Add blank lines between ifs.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        public static string GetModelName(string source)
+        {
+            return source;
+        }
+
+        public static string GetExtractorName(string source)
+        {
+            return source.Replace("Base", "").Replace("Extractor", "").Replace("Parser", "");
```
tellarin: Strings as constants to facilitate maintenance later.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/Spanish/TestDateParser.cs
```diff
@@ -10,6 +10,13 @@ public class TestDateParser
         readonly IDateTimeParser parser;
         readonly BaseDateExtractor extractor;
 
+        [ClassCleanup]
+        public static void ClassCleanup()
+        {
+            TestWriter.Close("Spa", typeof(BaseDateParser));
+        }
+
```
tellarin: Nit

.NET/Microsoft.Recognizers.Text.DateTime.Tests/Spanish/TestDurationParser.cs
```diff
@@ -7,15 +7,24 @@ public class TestDurationParser
     {
         readonly BaseDurationExtractor extractor = new BaseDurationExtractor(new SpanishDurationExtractorConfiguration());
         readonly IDateTimeParser parser = new BaseDurationParser(new SpanishDurationParserConfiguration(new SpanishCommonDateTimeParserConfiguration()));
+        readonly System.DateTime date = new System.DateTime(2016, 11, 7).Date;
```
tellarin: referenceDate

.NET/Microsoft.Recognizers.Text.Tests/TestWriter.cs
```diff
@@ -0,0 +1,284 @@
...
+        public string TypeName { get; set; }
+        public IDictionary<string, object> Resolution { get; set; }
+
+        public TestModelResult(ModelResult result)
+        {
+            this.Text = result.Text;
+            this.TypeName = result.TypeName;
+            Resolution = new Dictionary<string, object>();
+            object value;
+            if (result.Resolution.TryGetValue("value", out value))
```
tellarin: Nit

.NET/Microsoft.Recognizers.Text.DataDrivenTests/NumberWithUnit/TestNumberWithUnit_Chs.cs
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        [DataSource("Microsoft.VisualStudio.TestTools.DataSource.CSV", "DimensionModel-Chs.csv", "DimensionModel-Chs#csv", DataAccessMethod.Sequential)]
+        [TestMethod]
+        public void DimensionModel()
+        {
+            base.TestNumberWithUnit();
+        }
+
+        [DataSource("Microsoft.VisualStudio.TestTools.DataSource.CSV", "TemperatureModel-Chs.csv", "TemperatureModel-Chs#csv", DataAccessMethod.Sequential)]
```
tellarin: I see these are pointing to CSVs, but the PR includes a few jsons.
What's the expected flow for this PR? Just confirming.

pcostantini: The DataSource requires a CSV file to work. The TestMethod then iterates on all the CSV rows.
The TestInitialize method loads all test cases from the Spec file and creates a CSV with the same amount of test cases. The CSV is then used only as an iterator for the [TestMethod].

It uses a similar approach as to MsTestHacks, but much more focused to our problem.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
pcostantini: Updated to use Enums instead of strings

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
pcostantini: Done. All unresolved extractors/parsers or models will throw an exception instead

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
pcostantini: Updated to use Culture's cultureName instead of hardcoded CHS, SPA, ENG. The Specs folder structure was updated accordingly

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
pcostantini: Done!

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
pcostantini: Replaced all .Replace in several places by calling a single method (SanitizeSourceName:  https://github.com/southworkscom/Recognizers-Text/blob/e12d3f040fce71cb29d336b2f0703deec5c7d8b8/.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs#L337)

.NET/Microsoft.Recognizers.Text.DateTime.Tests/Spanish/TestDurationParser.cs
pcostantini: Done!

.NET/Microsoft.Recognizers.Text.Tests/TestWriter.cs
pcostantini: Done

.NET/Microsoft.Recognizers.Text.DateTime.Tests/Spanish/TestDateParser.cs
pcostantini: Moved to string constants (e.g.: TestCulture.English to represent "English")

PR #113, opened 2017-09-22
Move TypeScript/JavaScript solution to Spec based approach for running unit-tests 

pcostantini:
This PR contains the necessary work to update the TypeScript/Javascript project to use Spec based approach for running the unit-tests.

NOTE: This PR depends on merging https://github.com/Microsoft/Recognizers-Text/pull/112, which contains the Spec files.

NOTE 2: The tests are configured to run both data-driven tests (based on the Specs) AND the deprecated tests. Since we are still migrating Spanish recognizers. Future work will only use the Spec based approach and deprecated test files will be removed.


pcostantini:
This PR was updated with the same feedback from .NET PR

PR #118, opened 2017-09-27
Fix small regression bugs and "the week of"

rubio41:
- Add 649 midnight tonight. 
- ADD LUNCH AT 12:30 PM ON FRI 
- For four Monday at noon
- What do I have the week of November 30th


.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.26730.3
+# Visual Studio 14
```
tellarin: Let's not change this back to older versions.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestTimeParser.cs
```diff
@@ -214,12 +215,14 @@ public void TestTimeParseLuis()
         [TestMethod]
         public void TestTimeParseMealTimeLuis()
         {
-            BasicTest("I'll be back 12 o'clock lunchtime", "T12");
-            BasicTest("I'll be back 12 o'clock lunch", "T12");
-            BasicTest("I'll be back 8 o'clock dinner", "T20");
-            BasicTest("I'll be back 8 o'clock breakfast", "T08");
-            BasicTest("I'll be back at lunch 12 o'clock", "T12");
+            //We only support lunchtime now
+            //BasicTest("I'll be back 12 o'clock lunch", "T12");
```
tellarin: We'll add a task to potentially bring some of these back only as disambiguation.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestDatePeriodExtractor.cs
```diff
@@ -181,6 +181,7 @@ public void TestDatePeriodExtractSeason()
         public void TestDatePeriodExtractWeekOf()
         {
             //test week of and month of
+            BasicTest("the week of september.15th", 0, 26);
```
tellarin: In new unit tests we should add longer more complete sentences. That will help make sure corner cases don't happen.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -415,7 +415,6 @@
   "Context": {
     "ReferenceDateTime": "2016-11-07T00:00:00"
   },
-  "NotSupported": "javascript",
```
tellarin: Can you revert these removal changes? These tests should still be marked as not supported in JS.

Specs/NumberWithUnit/English/DimensionModel.json
```diff
@@ -570,7 +570,6 @@
 ,
 {
   "Input": "in 1995 canon introduced the first commercially available slr lens with internal image stabilization , ef 75 -300mm f / 4 - 5 . 6 is usm.",
-  "NotSupported": "javascript",
```
tellarin: Same.

.NET/Microsoft.Recognizers.Text.DateTime.Tests/English/TestTimeParser.cs
rubio41: done

PR #128, opened 2017-10-05
.NET - Remove C# Tests in favor of Spec based tests

pcostantini:
- Remove all Test projects, leaving only the Spec based test runner
- Add Spec with DateTime SplitDateAndTime
- Add Spanish Holiday Spec
- Include DecimalAndThousandsSeparators and Number ParserFactory tests (helpers not covered by Specs)

The Code Coverage with the Spec based approach is the same as with the older tests.


tellarin:
With the new French timex, this PR needs to be modified.

pcostantini:
Hi @tellarin ,
Merge conflicts were resolved. New French DateTime tests were re-created as Spec files

PR #134, opened 2017-10-10
fix an exception & add synonyms for "before"

rubio41:
fix the key not found exception while it is with before/after
add "in advance of" and "prior to" as "before".


.NET/Microsoft.Recognizers.Definitions/Spanish/DateTimeDefinitions.cs
```diff
@@ -396,5 +398,14 @@ public static class DateTimeDefinitions
 		public const string NextPrefixRegex = @"(pr[oó]xim[oa]|siguiente)\b";
 		public const string PastPrefixRegex = @"([uú]ltim[oa])\b";
 		public const string ThisPrefixRegex = @"(est[ea])\b";
+		public const string RelativeDayRegex = @"";
```
tellarin: Are these right? Isn't it undoing something from the Spanish code?

PR #138, opened 2017-10-11
Fix JavaScript parse reference date

enzocanoo:
**Issue:** JavaScript DateTime tests fails on node js v7 or prior because `Date.parse` behavior differs from the implemented on node js v8. [Related node issue (Date parse behavior change in node-v8)](https://github.com/nodejs/node/issues/13210)

**Fix:** Use a custom `ISODateString-to-dateObject` converter to get the same behavior in all node versions.

Additionally, the JavaScript build script has been modified to force build submodules.

### DIFF
--- a/JavaScript/test/runner-datetime.js
+++ b/JavaScript/test/runner-datetime.js
@@ -210,9 +210,14 @@ function getModel(config) {
     return findModel(options, cultureCode);
 }
 
+function parseISOLocal(s) {
+    var b = s.split(/\D/);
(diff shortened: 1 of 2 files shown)


JavaScript/test/runner-datetime.js
```diff
@@ -210,9 +210,14 @@ function getModel(config) {
     return findModel(options, cultureCode);
 }
 
+function parseISOLocal(s) {
```
tellarin: Can we add a comment in the code here to explain why this was needed? It will be helpful in the future.

PR #147, opened 2017-10-18
Javascript - Chinese number & number with unit recognizers 

enzocanoo:
Chinese implementation of number & number-with-unit recognizers


Patterns/Chinese/Chinese-Numbers.yaml
```diff
@@ -78,34 +78,20 @@ ZeroToNineMapChs: !dictionary
     俩: 2
     倆: 2
     仨: 3
-RoundNumberMapChs: !dictionary
```
tellarin: Was this duplicated?

enzocanoo: Yes, it was defined on [L16](https://github.com/southworkscom/Recognizers-Text/blob/28804d1cc595248f81efd9f7e158d7ba668d91fa/Patterns/Chinese/Chinese-Numbers.yaml#L16)

PR #158, opened 2017-10-27
French number with unit

matthewshim-ms:
Age model, Temp Model, and Dimension model working


tellarin:
@matthewshim-ms, can you take a look at the build failure above? It seems 22 unit tests are failing.
Thanks.

PR #160, opened 2017-10-27
French-NumberWithUnit support 

matthewshim-ms:
Re-submit for unit tests, now passing (Age, Currency, Temperature, Dimension)


Specs/NumberWithUnit/French/CurrencyModel.json
```diff
@@ -29,11 +29,11 @@
 }
 ,
 {
-  "Input": "Siegel payé national et 94.000 $ shuster pour laisser tomber toutes les réclamations.",
+  "Input": "Siegel payé national et 94000 $ shuster pour laisser tomber toutes les réclamations.",
   "NotSupportedByDesign": "python",
   "Results": [
     {
```
tellarin: If 94.000 is a valid number format in French, we should also support it. 
If it is, let's also open another issue to implement support for this and create a new unit test case for this example, ok?

Specs/NumberWithUnit/French/CurrencyModel.json
```diff
@@ -209,11 +209,11 @@
 }
 ,
 {
-  "Input": "le parc et sa famille a économisé pendant quatre ans pour acheter un appartement minuscule ici, mais a constaté que plus près ils sont arrivés à sauver 40,000 $ dont ils ont à l'origine eu besoin, plus du prix est monté.",
+  "Input": "le parc et sa famille a économisé pendant quatre ans pour acheter un appartement minuscule ici, mais a constaté que plus près ils sont arrivés à sauver 40000 $ dont ils ont à l'origine eu besoin, plus du prix est monté.",
   "NotSupportedByDesign": "python",
   "Results": [
     {
-      "Text": "40,000 $",
+      "Text": "40000 $",
```
tellarin: Similar case as above.

Specs/NumberWithUnit/French/CurrencyModel.json
```diff
@@ -624,6 +624,14 @@
         "value": "63,45",
         "unit": "Dollar"
       }
+    },
+    {
+      "Text": "cent",
```
tellarin: Cent in 8 pour cent should not be extracted in this case.
Let's disable this test with notsupported and open a new issue to fix it.

Specs/NumberWithUnit/French/CurrencyModel.json
```diff
@@ -888,7 +896,7 @@
   "NotSupportedByDesign": "python",
   "Results": [
     {
-      "Text": "12-cent",
```
tellarin: Is 12-cent a form that typically happens in written French? Formal or informal?
If a common form, we should try to support it too.

Specs/NumberWithUnit/French/CurrencyModel.json
matthewshim-ms: Most common use is 'spaces' to separate thousands, although period is formally accepted - [see source](https://french.kwiziq.com/revision/grammar/how-to-write-decimal-numbers-in-french). It should basically be the same as the spanish one. When I added a 'space' case to the spanish recognizer it did not pass. The most important is that commas denote decimal separator

Specs/NumberWithUnit/French/CurrencyModel.json
matthewshim-ms: Ideally would be "40 000" or "40.000" - it should match the spanish configuration.

PR #167, opened 2017-11-01
Fix ending number as time

rubio41:
for the case "move my 3pm appointment to 4"
the single number will be extracted as time if it is 1. at the end of the sentence. 2. meeting the pattern "[time] (appointment|meeting|conference) to number"


tellarin:
Solution needs further changes.

PR #168, opened 2017-11-01
Fix ending number as time

rubio41:
add support to "move my 3pm appointment to 4"
recognize "4" as time only 1. at the end of sentence or follow by ,?, !;2. match the pattern "[time|datetime] (appointment|meeting|call) to [number]"


.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishMergedExtractorConfiguration.cs
```diff
@@ -37,10 +40,12 @@ public class EnglishMergedExtractorConfiguration : IMergedExtractorConfiguration
 
         public IExtractor DurationExtractor { get; }
 
-        public IExtractor GetExtractor { get; }
+        public IExtractor SetExtractor { get; }
```
tellarin: Is this rename also needed in other languages?

PR #176, opened 2017-11-06
Determine the date and weekofday is the same day or not

rubio41:
For this fix, need to pass the reference time to the extractor.
Therefore, change the interface of IExtractor to IDateTimeExtractor.


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -109,14 +115,17 @@ private List<Token> NumberWithMonth(string text)
...
-                        ExtractResult erTmp = new ExtractResult();
-                        erTmp.Text = match.Groups["DayOfMonth"].Value;
-                        erTmp.Start = match.Groups["DayOfMonth"].Index;
-                        erTmp.Length = match.Groups["DayOfMonth"].Length;
-                        var day = Convert.ToInt32((double)(this.config.NumberParser.Parse(erTmp).Value ?? 0));
+                        int month = reference.Month, year = reference.Year;
+
+                        // get week of day for the ordinal number which is regarded as a date of reference month
+                        var date = DateObject.MinValue.SafeCreateFromValue(reference.Year, reference.Month, num);
+                        var date2weekdayStr = date.DayOfWeek.ToString().ToLower();
```
tellarin: numWeekDayStr

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -292,51 +292,27 @@ private DateTimeResolutionResult ParseImplicitDate(string text, DateObject refer
...
                 int month = referenceDate.Month, year = referenceDate.Year;
-
                 // create a extract result which content ordinal string of text
                 ExtractResult erTmp = new ExtractResult();
                 erTmp.Text = match.Groups["DayOfMonth"].Value.ToString();
                 erTmp.Start = match.Groups["DayOfMonth"].Index;
                 erTmp.Length = match.Groups["DayOfMonth"].Length;
-
-                // parse the day in text into number, and get week of day of this number which regarded as a day in this month
+                // parse the day in text into number
```
tellarin: Add a blank line before the comment to make code more readable.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -292,51 +292,27 @@ private DateTimeResolutionResult ParseImplicitDate(string text, DateObject refer
...
-                }
-                else
-                {
-                    //the resolution should have no value
-                    var weekdayStr = match.Groups["weekday"].Value.ToLower();
-                    var weekDay = this.config.DayOfWeek[weekdayStr];
-                    var timexDayOfWeek="XXXX-WXX-" + weekDay;
-                    var timexDay = FormatUtil.LuisDate(-1, -1, day);
-                    ret.Timex = timexDayOfWeek + "," + timexDay;
+                // the validity of the phrase is guaranteed in the Date Extractor
```
tellarin: Same

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -223,7 +223,8 @@ private DateTimeResolutionResult ParseEach(IExtractor extractor, IDateTimeParser
 
             if (success)
             {
-                var pr = parser.Parse(ers[0], DateObject.Now);
+                //var pr = parser.Parse(ers[0], DateObject.Now);
```
tellarin: Why not just delete this?

Specs/DateTime/English/DateExtractor.json
```diff
@@ -915,6 +915,10 @@
 ,
 {
   "Input": "What do I have on Saturday the second",
+  "Context": {
+    "ReferenceDateTime": "2017-09-01T00:00:00"
+  },
+  "NotSupported": "javascript",
```
tellarin: This is breaking too many javascript cases not related to the new supported case.
Can't those be fixed as well?
Why adding a ReferenceDateTime breaks them??

Specs/DateTime/English/MergedExtractor.json
```diff
@@ -488,6 +488,142 @@
 }
 ,
 {
+  "Input": "I'll change the ten am appointment to 20!",
+  "NotSupported": "javascript",
```
tellarin: New cases are fine to ignore in JS. But we shouldn't break old ones.

Specs/DateTime/English/MergedParser.json
```diff
@@ -309,29 +309,6 @@
 }
 ,
 {
-  "Input": "put make cable's wedding in my calendar for wednesday the thirty first",
```
tellarin: Why was this removed?

Specs/DateTime/English/MergedParser.json
```diff
@@ -1018,5 +995,164 @@
       }
     }
   ]
+},
+{
+  "Input": "I'll change the ten am appointment to twenty!",
```
tellarin: What if there is no punctuation char at the end?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -109,14 +115,17 @@ private List<Token> NumberWithMonth(string text)
...
 
-                        if (day == num)
+                        // get week day from text directly, compare it with the weekday generated above
```
tellarin: Add blank.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/IDateTimeExtractor.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Enable a reference time pass to the extractor
+
+using DateObject = System.DateTime;
+using System.Collections.Generic;
+
+namespace Microsoft.Recognizers.Text.DateTime
+{
+    public interface IDateTimeExtractor : IExtractor
```
tellarin: Can't this interface be an abstract class and include the default implementation of Extract?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -292,51 +292,27 @@ private DateTimeResolutionResult ParseImplicitDate(string text, DateObject refer
...
-                    ret.Success = true;
-
-                    return ret;
-                }
-                else
-                {
-                    //the resolution should have no value
-                    var weekdayStr = match.Groups["weekday"].Value.ToLower();
-                    var weekDay = this.config.DayOfWeek[weekdayStr];
-                    var timexDayOfWeek="XXXX-WXX-" + weekDay;
```
tellarin: Where did this go now?

rubio41: we determine that in the extractor. We will output two separate results instead of a whole with no resolution when the two dates are not match.

Specs/DateTime/English/MergedParser.json
rubio41: the same. We have similar test cases for that.

Specs/DateTime/English/MergedParser.json
rubio41: added back. The output is different since September doesn't contain the thirty first.

PR #191, opened 2017-11-15
Add support for the requirements of users and some improvements

rubio41:
add some supports for users

- [x] weekday with a time range, e.g. on Friday between 1pm and 4pm
- [x] weekdays + [this week/next week], e.g. friday this week or tuesday of next week
- [x] late/later, early [this week/next week/this month...]
- [x] weekday + part of day, e.g. this Friday, in the afternoon
- [x] filter the 1:1 under DatetimeOption.Calendar (the test case is not added since the current test runner doesn't support tests with options)

some improvements:

- [x] Supporting the "16. nov. 2016 10:38" format
- [x] Merge of 6 Jan 2017 - 6:37am


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -508,6 +568,19 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
...
+            var match = this.config.WeekWithWeekDayRangeRegex.Match(text);
+            string weekPrefix = null;
+            if (match.Success)
+            {
+                weekPrefix = match.Groups["week"].ToString();
+            }
+
+            if (!string.IsNullOrEmpty(weekPrefix))
+            {
+                er[0].Text = weekPrefix + " " + er[0].Text;
```
tellarin: Don't concatenate string like this. Use string interpolation for easier to read code.

cleemullins:
I find this PR puzzling. Can someone help me understand:


  1.  I see updates to the .yaml, .cs, and .json files.
  2.  I don’t see any updates to the .js files.

My understanding is that the “truth” is in Yaml, and the CS and JS files are effectively auto-generated. The lack of.js files in the PR means either:


  1.  My understanding is incorrect.
  2.  The PR is incorrect.

Can someone enlighten me?

//Chris

cleemullins:
Hi Chris,

Yes, the YAML are used to generate the .CS and .JS with the Regular Expressions. The specs/tests are in the .JSON files.


For what I see in the PR, the changes are only adding new features to C# only.

The new specs are only enabled for C# and ignored for JavaScript.


This is something discussed previously to allow to push new features to a platform first. The idea is to later apply the same changes to JS.

This allows different paces to each platform and, at the same time, disallow breaking changes. This can be seen in the PR Build status<https://ci.appveyor.com/project/tellarin/recognizers-text/build/1.0.92> where both builds are passing.


[cid:46cf85a3-6cf6-40bc-84a5-94ddae756ffc]

(This flag will skip this test in JS).


Regards,

-PabloC.

________________________________
From: Chris Mullins <cmullins@microsoft.com>
Sent: Wednesday, November 15, 2017 3:50:16 PM
To: Microsoft/Recognizers-Text
Cc: manx; Wenhao Huang; Börje Karlsson
Subject: RE: [Microsoft/Recognizers-Text] Add support for the requirements of Calendar.Help and some improvements (#191)


I find this PR puzzling. Can someone help me understand:



  1.  I see updates to the .yaml, .cs, and .json files.
  2.  I don’t see any updates to the .js files.



My understanding is that the “truth” is in Yaml, and the CS and JS files are effectively auto-generated. The lack of.js files in the PR means either:



  1.  My understanding is incorrect.
  2.  The PR is incorrect.



Can someone enlighten me?



//Chris

cleemullins:
I’m not at all comfortable with C# changes only, like this.

I would like us to revert this PR until we have a booked plan for JS as well as C#.

My reason is:

  1.  These changes are specific to Calendar.Help (which is fine).
  2.  These changes will never be ported to JS (NOT fine).
  3.  These changes are not experimentation, where – at the end – they are ported everywhere. (NOT fine).

This is the beginning of Broken Window syndrome. Everyone will look at this and say, “I need X, but I only need it in my platform”. Next thing, our library is wildly inconsistent and not useful.

The anti-pattern here is summarized as:

We've seen clean, functional systems deteriorate pretty quickly once windows start breaking. There are other factors that can contribute to software rot, and we'll touch on some of them elsewhere, but neglect accelerates the rot faster than any other factor.

This is very real example of a broken window. ☹

//Chris

PR #192, opened 2017-11-15
Ported changes in PR #176 to JS/TypeScript

JuanAr:
Mostly changes to Datetime Extractors/Parsers to _"Determine the date and weekofday is the same day or not (#176)"_


JuanAr:
I'll enable those tests, verify them and commit the changes.

Issue #185, opened 2017-11-13
datetime recognizer should return simple JavaScript objects and arrays


johnataylor:
Current the result of parse appears to contain the new JavaScript data structures, things like Set and Map.

It would be better to restrict this small result to simple JavaScript object and array structures. This would make them easier to work with: the developer could for example trivially use JSON.stringify and then use simple language operators . and []

pcostantini:
Sounds good!
We'll submit a new PR shortly so the DateTime Recognizer return simple objects instead of Sets and Maps.
Regards

johnataylor:
Thanks - much easier to deal with for the caller. _One key point is that the data structure coming back from the recognizer should be exactly the same as the data structure we get from back from a call to LUIS_.

Specifically when we call JSON.stringify(parseResult) it should be the same string that I would have otherwise got on the wire.

So that might be useful for some ad-hoc testing - to ensure we are consistent. (Note I'm not suggesting test cases need to call out to LUIS.)

While you are looking into this I also noticed a similar bug this time with Date. _The start and end values in a resolution are Date objects when the type is daterange._ This is inconsistent with the value you might get as the resolution for, say, a date.

However datetimerange does comes back with start and end as strings. So the bug here is on daterange - it should use strings not Date.

tellarin:
@johnataylor, in general I agree with you. The representation in .NET and Javascript should be the same when turned into json and it's highly preferable to return simple objects.

But there may be small differences to the actual representation in LUIS as LUIS is only one consumer of the recognizers and we can't change format here without taking into account backwards compat to others.

Regarding the start/end comment, that's in javascript, right?

PR #197, opened 2017-11-17
JS - Fixes for issue #185 - DateTime recognizer should return simple JavaScript objects and arrays

pcostantini:
This PR fixes the issues reported in https://github.com/Microsoft/Recognizers-Text/issues/185 :

- Return simple JavaScript objects and Arrays instead of Maps
- For Duration and DateRange return values as strings (instead of Number or Date)
- Samples and README's Code snippets updated to reflect changes


pcostantini:
Rebased to include latest changes

PR #200, opened 2017-11-21
Change the week definition to ISO week definition

rubio41:
Change the week definition to ISO week definition. Require corresponding changes in Javascript.


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DatePeriodParserChs.cs
```diff
@@ -502,7 +502,7 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                         var endDate = referenceDate.This(DayOfWeek.Sunday).AddDays(7*swift);
 
                         ret.Timex = beginDate.Year.ToString("D4") + "-W" +
-                                    Cal.GetWeekOfYear(beginDate, CalendarWeekRule.FirstDay, DayOfWeek.Monday)
+                                    Cal.GetWeekOfYear(beginDate, CalendarWeekRule.FirstFourDayWeek, DayOfWeek.Monday)
```
tellarin: Shouldn't this be called CalendarWeekRule.FirstThursdayWeek instead of CalendarWeekRule.FirstFourDayWeek? That's less ambiguous as "four day" depends if you count from Sunday or Monday

tellarin: I see, this is a C# name already. Let's not change then. :)

PR #235, opened 2017-12-06
Initial German language structure

flo-zuewi:
Added German structure and first part of numbers.


flo-zuewi:
GitIgnore file is apparently faulty. Will work on it to make appveyor build succeed again.

PR #247, opened 2017-12-09
Setting the correct default no-options string to TryGetModel

tellarin:
Visibility of these methods will change in 1.1.0

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.sln
+++ b/.NET/Microsoft.Recognizers.Text.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27004.2010
+VisualStudioVersion = 15.0.27004.2006
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{ED7B6456-AB0A-48CE-8F85-711FE87F09C2}"
 EndProject
--- a/.NET/Microsoft.Recognizers.Text.sln
+++ b/.NET/Microsoft.Recognizers.Text.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27004.2006
+VisualStudioVersion = 15.0.27004.2010
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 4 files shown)


.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27004.2006
+VisualStudioVersion = 15.0.27004.2010
```
yiylin: Change this?

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -10,7 +10,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		CreateAllPackages.cmd = CreateAllPackages.cmd
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Recognizers.Text.DateTime", "Microsoft.Recognizers.Text.DateTime\Microsoft.Recognizers.Text.DateTime.csproj", "{19522B81-C63F-4DD5-AB4C-674CD44D67D6}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Recognizers.Text.DateTime", "Microsoft.Recognizers.Text.DateTime\Microsoft.Recognizers.Text.DateTime.csproj", "{19522B81-C63F-4DD5-AB4C-674CD44D67D6}"
```
yiylin: do not know why has this change?

.NET/Microsoft.Recognizers.Text/Recognizer.cs
```diff
@@ -17,12 +19,12 @@ protected void RegisterModel(string culture, string options, Dictionary<Type, IM
             modelContainer.RegisterModel(culture, models, options);
         }
 
-        public IModel GetModel<TModel>(string culture, bool fallbackToDefaultCulture = true, string options = "")
+        public IModel GetModel<TModel>(string culture, bool fallbackToDefaultCulture = true, string options = NoneOptions)
```
yiylin: No related unit test?

tellarin: I'll come up with a way to add these. As they will not follow the current Spec format.

.NET/Microsoft.Recognizers.Text.sln
tellarin: No, actually. I'll revert.

Issue #11, opened 2017-06-05
Support for PT-PT or PT-BR for date-time


AndreMantas:
I'm interested in date/time for Portuguese. How hard would it be to adapt existing code for Portuguese strings? Is there any steps one could follow? 

Thanks in advance.

chrimc62:
I think Börje was already looking at this?

tellarin:
Hi @AndreMantas,

Numbers and DateTime support in PT-BR (and partial PT-PT) are coming. There should be a PR by the end of the week.

Any help expanding them would be greatly appreciated.

Thanks.

AndreMantas:
Any updates on this? Would gladly help, just don't know how.

tellarin:
Hi Andre, this was delayed due to some refactoring and bug fixes.
If you could translate the datetime unit tests from English or Spanish to Portuguese, and add Portuguese-specific test cases, that would be very helpful.
We'll publish a spec soon to help people with how and where they can contribute. Any suggestions are very welcome.

AndreMantas:
Hi tellarin. Didn't see the notification before so sorry for delay.

So, you mean the unit tests in this folder right? https://github.com/Microsoft/Recognizers-Text/tree/master/Microsoft.Recognizers.Text.DateTime.Tests/English

Any particular classes that are more important than others?

tellarin:
@AndreMantas, this was on hold for a while as many structural changes to the project were ongoing. 
Now it's much easier to add new languages and to reuse resources and tests across C# and Javascript. 
The PT timex is ongoing now and should be done soon.

PR #256, opened 2017-12-15
Enabling Portuguese in DateTime recognizers.

tellarin:
Enabling PT in timex & adding try catches.
Issue #11.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
@@ -2,8 +2,10 @@
 
 using Microsoft.Recognizers.Text.DateTime.Chinese;
 using Microsoft.Recognizers.Text.DateTime.English;
-using Microsoft.Recognizers.Text.DateTime.Spanish;
 using Microsoft.Recognizers.Text.DateTime.French;
+using Microsoft.Recognizers.Text.DateTime.Portuguese;
+using Microsoft.Recognizers.Text.DateTime.Spanish;
+
 using Microsoft.Recognizers.Text.Utilities;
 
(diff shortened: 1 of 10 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -80,24 +87,35 @@ private DateTimeRecognizer(string cultureCode, DateTimeOptions options)
...
+
                 case Culture.French:
                     RegisterModel(Culture.French, type, options.ToString(), new DateTimeModel(
                                       new BaseMergedParser(new FrenchMergedParserConfiguration(options)),
                                       new BaseMergedExtractor(new FrenchMergedExtractorConfiguration(options))
                                   ));
                     break;
+
+                case Culture.Portuguese:
+                    RegisterModel(Culture.French, type, options.ToString(), new DateTimeModel(
```
yiylin: French？

Specs/DateTime/Portuguese/DateExtractor.json
```diff
@@ -34,6 +34,20 @@
   ]
 }
 ,
+{
+  "Input": "nas ultimas 3 semanas",
+  "NotSupported": "javascript",
+  "NotSupportedByDesign": "python",
+  "Results": [  ]
```
yiylin: Cause exception, so no result?

tellarin: Nope. This should never be extracted as this type.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
tellarin: Fixed.

PR #260, opened 2017-12-21
German Numbers and NumbersWithUnits

flo-zuewi:
Worked on translation of the Numbers-section. Mostly finished NumbersWithUnits-section.


tellarin:
Started reviewing this.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/NumberWithUnitRecognizer.cs
```diff
@@ -227,6 +227,46 @@ private NumberWithUnitRecognizer(NumberWithUnitOptions options)
                             }
                             ),
             });
+
+            RegisterModel(Culture.German, options.ToString(), new Dictionary<Type, IModel>
```
tellarin: The model registration in the *Recognizer.cs files should only be added once the full entities are supported. Otherwise it will expose detection in a language that might not be ready.

tellarin: Can you remove this for now if there's still any pending units case?
Something like "twenty three euros" in German wouldn't work yest, right?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -22,6 +22,9 @@ public BaseNumberParser(INumberParserConfiguration config)
         {
             this.Config = config;
 
+            //As soon as you develop on a machine that does not use en-US as standard-culture, numbertests for the other languages fail
+            System.Threading.Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
```
tellarin: We're looking at a better solution for this.

tellarin:
Ideally the tests should not be removed, just marked as not supported in dotnet too, if that's the case.

flo-zuewi:
I already implemented NumberWithUnit and a lot of test-cases for it, while I was awaiting your answer regarding the possible solution for the number-parser-problem.
These tests normally work,  I avoided using numbers that cannot be recognized by the numberparser right now. I did not mark them as not supported in the first place, because they actually do work.

The removal of RegisterModel causes all of these tests to fail.
I did not want to waste time on marking every single working test as not supported and changing all of it back later.
I made a backup of all the test-cases, which i wanted to reintegrate later. If you want, i can of course revert the removal and mark every case as not supported.

tellarin:
Sounds good to me in this PR. But the typical flow is to always add all the tests marked as not supported, and PR by PR, mark them as supported (and add more, if needed).
This way the toolset has a sense of the remaining work to finish some entity extractor.
If the specs are not there, there is no estimate of progress/coverage.

tellarin:
I'll merge this one, but let's add the test specs from now on, ok?

flo-zuewi:
Of course. I just did this because of the extraordinary circumstance of having already a bunch of working NumberWithUnit test while not all the Numbers could be be recognized correctly.

PR #262, opened 2018-01-09
fixed FR-weds (mercredi) definition

matthewshim-ms:


tellarin:
No need to close PRs like this. If you need to make changes, just push to your PR branch.

PR #275, opened 2018-01-16
Recog german (Numbers and NumbersWithUnit)

flo-zuewi:
Fixed the Blocking issue. Numbers and NumbersWithUnit now working, apart from a very few special cases.


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -27,6 +27,12 @@ public BaseNumberParser(INumberParserConfiguration config)
                                 + GetKeyRegex(this.Config.OrdinalNumberMap.Keys);
 
             TextNumberRegex = new Regex(@"(?<=\b)(" + singleIntFrac + @")(?=\b)", RegexOptions.IgnoreCase | RegexOptions.Singleline);
+
+            //necessary for the german language because bigger numbers are not separated by whitespaces or special characters like in other languages
+            if (config.CultureInfo.Name == "de-DE") {
+                TextNumberRegex = new Regex(@"(" + singleIntFrac + @")", RegexOptions.IgnoreCase | RegexOptions.Singleline);
+            }
```
tellarin: In the next PR, lines 29 to 34 could become a method in a Utilities/NumberParsingUtil class in the Text.Number project that receives singleIntFrac and the culture name as parameter and returns the appropriate regex. That would make the code a little easier to read and to track such special behaviour.

That util can also have a map of cultures that may require it. For now, only German. Instead of comparing to a specific culture string. We could call the mapping something like NonSpaceTokenizedCultures.

flo-zuewi: Ok

Issue #240, opened 2017-12-08
[DE DateTimeV2] Implement support for German DateTimes


tellarin:

tellarin:
@flo-zuewi has an initial PR for it.

Michaelvsk:
What's the progress here? datetimev2 is still not supported by LUIS and the old datetime entity type is practically unusable for german language (way too many false positives and other bugs).

pke:
Is there some way I can help here? When will the partial DateTime made available in LUIS? @tellarin

tellarin:
Hi @pke,
you can start taking a look at the Spec json files for DateTime that are marked as "NotSupported": "dotnet". Some of them also have "Comment" tags that may give more details on the current issues.
We suggest focusing on the .NET plat first and then we can port to JS. 
Writing additional specs to make sure coverage is good are also very valuable. 
Unfortunately the German DateTime quality doesn't meet our bar to ship yet. :-(
Any help is greatly appreciated!

tellarin:
Closing this as support is now available. Thanks @n01d, @miserale, and everyone else!

pke:
Cool! Which PR brought this in?

tellarin:
It was not a specific PR, but release dotnet-v1.1.5 made it "official".

PR #292, opened 2018-01-25
Recog german (DateTime-Extractors and Bugfixes)

flo-zuewi:
Continued implementation of https://github.com/Microsoft/Recognizers-Text/issues/240


.NET/Microsoft.Recognizers.Text.DateTime/German/Extractors/GermanDateExtractorConfiguration.cs
```diff
@@ -102,6 +102,12 @@ public class GermanDateExtractorConfiguration : BaseOptionsConfiguration,IDateEx
             // 23/7
             new Regex(DateTimeDefinitions.DateExtractor9, RegexOptions.IgnoreCase | RegexOptions.Singleline),
 
+            //10
```
tellarin: Can you add examples in the code comment here that illustrate which cases would match this regex?

.NET/Microsoft.Recognizers.Text.DateTime/German/Extractors/GermanDateExtractorConfiguration.cs
```diff
@@ -102,6 +102,12 @@ public class GermanDateExtractorConfiguration : BaseOptionsConfiguration,IDateEx
...
+            new Regex(DateTimeDefinitions.DateExtractor10, RegexOptions.IgnoreCase | RegexOptions.Singleline),
+
+            //11
```
tellarin: Same

Specs/DateTime/German/MergedExtractor.json
```diff
@@ -1 +1,144 @@
...
+    "NotSupported":"javascript",
+    "Results": [
+      {
+        "Text": "4 Uhr",
+        "Type": "time"
+      }
+    ]
+  }
+  ,
+  {//should be datetime
```
tellarin: What do you mean? It should be recognized as datetime, but due to a bug falls here?

Specs/DateTime/German/MergedExtractor.json
```diff
@@ -1 +1,144 @@
...
+    "NotSupportedByDesign": "python",
+    "NotSupported":"javascript",
+    "Results": [
+      {
+        "Text": "letzte Woche",
+        "Type": "daterange"
+      }
+    ]
+  },
+  {//should be datetimerange
```
tellarin: Same question.

Specs/DateTime/German/MergedExtractor.json
```diff
@@ -1 +1,144 @@
...
+    ]
+  },
+  {
+    "Input": "Das nächste meeting wird am 16. März 2017 sein. Wie wäre es denn mit einer Diskussionsrunde heute um 2 Uhr Nachmittags?",
+    "NotSupported": "javascript",
+    "NotSupportedByDesign": "python",
+    "Results": [
+      {
+        "Text": "16. März 2017 ",
+        "Type": "set" //should be date
```
tellarin: If this is wrong, this test should not be enabled. Should be moved to the correct spec and marked as notsupported for now.

Specs/NumberWithUnit/German/AgeModel.json
```diff
@@ -142,23 +142,6 @@
   ]
 }
 ,
-{//ordinals are currently not supported in the AgeModel
```
tellarin: Why remove the test and not keep here as unsupported for now?

.NET/Microsoft.Recognizers.Text.DateTime/German/Extractors/GermanDateExtractorConfiguration.cs
flo-zuewi: Yes, I will do this with the next PR. I am still not fully done with the extractors, some of them will probably be deleted and others changed. DateExtractor 10 and 11 will most certainly be blended into the one or two of the other ones and removed after that.

Specs/DateTime/German/MergedExtractor.json
flo-zuewi: The set-extractor seems to be a little bit to "dominant",  looking further into this issue today, forgot to put the testcases back to not supported

Specs/NumberWithUnit/German/AgeModel.json
tellarin: OK

PR #294, opened 2018-01-30
Remove null JSON return value in SimpleConsole

lucinezhang:


tellarin:
Wrong approach to solve the problem. Closing this PR/

PR #296, opened 2018-01-31
Remove null JSON return value

lucinezhang:


.NET/Microsoft.Recognizers.Text.DateTime/Models/DateTimeModel.cs
```diff
@@ -55,15 +55,32 @@ public List<ModelResult> Parse(string query, System.DateTime refTime)
...
-            }).ToList();
+                Start = parsedDateTime.Start.Value,
+                End = parsedDateTime.Start.Value + parsedDateTime.Length.Value - 1,
+                TypeName = parsedDateTime.Type,
+                Resolution = parsedDateTime.Value as SortedDictionary<string, object>,
+                Text = parsedDateTime.Text
+            };
+
+            var parentText = GetParentText(parsedDateTime);
+            if (!string.IsNullOrEmpty(parentText))
```
tellarin: This shouldn't be based on the string value. This should be based on the flag for extendedtypes being enabled or not, no?

.NET/Microsoft.Recognizers.Text/Models/ExtendedTypeModelResult.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Recognizers.Text
+{
+    public class ExtendedTypesModelResult : ModelResult
```
tellarin: ExtendedModelResult

.NET/Microsoft.Recognizers.Text/Models/ExtendedTypeModelResult.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public string ParentText { get; set; }
+
+        public ExtendedTypesModelResult() { }
```
tellarin: Nit: code standard.

PR #304, opened 2018-02-01
Improve datetime extraction

lucinezhang:
github issue #289 #290 #291


.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishDatePeriodParserConfiguration.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Recognizers.Text.DateTime.English
 {
     public class EnglishDatePeriodParserConfiguration : BaseOptionsConfiguration, IDatePeriodParserConfiguration
     {
+        private const string AfterNextSuffix = "after next";
```
tellarin: Don't add as string, if needed, add as resource in YAML. Even better if it can support multiple values.

.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
```diff
@@ -204,7 +205,7 @@ public static class DateTimeDefinitions
 		public const string OrRegex = @"\s*(or)\s*";
 		public static readonly string YearPlusNumberRegex = $@"\b(Year\s+((?<year>(\d{{3,4}}))|{FullTextYearRegex}))\b";
 		public static readonly string NumberAsTimeRegex = $@"\b({EngTimeRegex}|{PeriodHourNumRegex}|{BaseDateTime.HourRegex})\b";
-		public static readonly string TimeBeforeAfterRegex = $@"\b(((?<=\b(before|after)\s+)({EngTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex}|{MidTimeRegex}))|{MidTimeRegex})\b";
+		public static readonly string TimeBeforeAfterRegex = $@"\b(((?<=\b(before|(no later (than|by))|after)\s+)({EngTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex}|{MidTimeRegex}))|{MidTimeRegex})\b";
```
tellarin: What's "no later by"?
Shouldn't it be like "by tomorrow" or "no later than tomorrow"?
Same in BeforeRegex

Specs/DateTime/English/DateTimeModelCalendarMode.json
```diff
@@ -168,23 +193,23 @@
 }
 ,
 {
-  "Input": "Schedule a meeting after three",
+  "Input": "I'll leave by tomorrow",
   "Context": {
```
tellarin: Why is this here? This is not something only in calendar mode. Can you move to the sample place as other dateranges?

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishDatePeriodParserConfiguration.cs
```diff
@@ -209,7 +207,7 @@ public bool IsLastCardinal(string text)
         public bool IsMonthOnly(string text)
         {
             var trimedText = text.Trim().ToLowerInvariant();
-            return trimedText.EndsWith("month") || trimedText.EndsWith(string.Join(" ", "month", AfterNextSuffix));
+            return trimedText.EndsWith("month") || Regex.IsMatch(text, @"\bmonth\b") && AfterNextSuffixRegex.IsMatch(trimedText);
```
tellarin: We should avoid using Regex.IsMatch if possible. Especially if there are too many uses of it.
Can you try to refactor? Please reach out if not sure.
Thanks.

Issue #302, opened 2018-02-01
[EN/CN DateTimeV2] Improve support for year ranges


tellarin:
Support cases like:

"2014-2015"
"between 2012-2016"
"2012 or above"
"2012 or after"
"from 2000 to 2020"
"from 2000 to 2020"
"until 2020"

tellarin:
CN side is next, right @lucinezhang ? ;)

PR #313, opened 2018-02-03
Support Chinese Year Range

lucinezhang:
github issue #302


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DatePeriodParserChs.cs
```diff
@@ -311,26 +311,26 @@ private static DateTimeResolutionResult ParseYearToYear(string text, DateObject
                     }
                 }
 
-                if (beginYear < 100 && beginYear >= 90)
+                if (beginYear < 100 && beginYear >= 50)
```
tellarin: This should be using constants in the YAML file. Can you fix in a next PR?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DatePeriodParserChs.cs
```diff
@@ -311,26 +311,26 @@ private static DateTimeResolutionResult ParseYearToYear(string text, DateObject
...
                 }
-                else if (endYear < 100 && endYear < 20)
+                else if (endYear < 100 && endYear < 50)
                 {
                     endYear += 2000;
                 }
 
                 var beginDay = DateObject.MinValue.SafeCreateFromValue(beginYear, 1, 1);
-                var endDay = DateObject.MinValue.SafeCreateFromValue(endYear, 12, 31);
+                var endDay = DateObject.MinValue.SafeCreateFromValue(endYear, 1, 1);
```
tellarin: Why this change? The end day of a typical year range is the last day of the last year.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedParser.cs
```diff
@@ -75,6 +63,19 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refTime)
                 er.Text = er.Text.Substring(sinceMatch.Length);
                 modStr = sinceMatch.Value;
             }
+            else if (er.Type.Equals(Constants.SYS_DATETIME_DATEPERIOD) && Config.YearRegex.Match(er.Text).Success)
```
tellarin: Why this order change? Please add a comment in the code describing the flow and need; in case this is strictly necessary. To avoid others breaking it in the future by mistake.

JavaScript/packages/recognizers-date-time/src/dateTime/chinese/datePeriodConfiguration.ts
```diff
@@ -772,23 +772,23 @@ export class ChineseDatePeriodParser extends BaseDatePeriodParser {
...
             }
         }
 
         beginYear = this.sanitizeYear(beginYear);
         endYear = this.sanitizeYear(endYear);
 
-        let beginDate = DateUtils.safeCreateFromMinValue(beginYear, 1, 1);
-        let endDate = DateUtils.safeCreateFromMinValue(endYear, 11, 31);
+        let beginDate = DateUtils.safeCreateFromMinValue(beginYear, 0, 1);
+        let endDate = DateUtils.safeCreateFromMinValue(endYear, 0, 1);
```
tellarin: Same

Specs/DateTime/Chinese/DatePeriodParser.json
```diff
@@ -467,6 +467,82 @@
...
+  "NotSupportedByDesign": "python",
+  "Results": [
+    {
+      "Text": "从2016到2018",
+      "Type": "daterange",
+      "Value": {
+        "Timex": "(2016,2018,P2Y)",
+        "FutureResolution": {
+          "startDate": "2016-01-01",
+          "endDate": "2018-01-01"
```
tellarin: Hmmm. I see the issue.

tellarin: Can you add a code comment in the Parser method description explaining our default definition of interval? I.e. [start,end)

tellarin: For resolution, I'll update this comment later if we should use the format you have here, or another.

tellarin: To standardize things, the assumption is that all intervals are [start,end) and this matches the example here. 
Interval value in NumberRange, timex, and resolutions should always be non-inclusive of the end.
There is only one exception: for time, if the end of the range is midnight it will be represented as 23:59:59. Which doesn't apply here as this is Date only, not DateTime.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DatePeriodParserChs.cs
tellarin: See comment below. "this year" should correctly be
start: 2018-01-01
end: 2019-01-01

Issue #310, opened 2018-02-02
[* Sequence] Basic support for phone numbers


tellarin:
("622 97 54 17"),
    ("632 139823"),
    ("679 124 898"),
    ("651 193 123"),
    ("652123123"),
    ("+44 (0)7123 129683"),
    ("0044 07700 900873"),
    ("004407624938475"),
    ("(96) 192834614"),
    ("(55) 3333-33333"),
    ("(21)99713-3601"),

PR #319, opened 2018-02-06
Support basic/typical phone number recognition

STayinloves:
Github issue: https://github.com/Microsoft/Recognizers-Text/issues/310

- Add extractor and parser for phone numbers, with simple parser output.
- Add tests for normal phone number patterns in Brazil, UK, Germany,
Chile and Argentina, also some invalid tests.
- Modify SimpleConsole to support phone number recognize in it.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
+++ b/.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
@@ -18,7 +18,11 @@ namespace Microsoft.Recognizers.Definitions
 	public static class BasePhoneNumbers
 	{
 		public const string NumberReplaceToken = "@builtin.phonenumber";
-		public static readonly Func<string, string, string> IntegerRegexDefinition = (placeholder, thousandsmark) => $@"(((?<!\d+\s*)-\s*)|((?<=\b)(?<!(\d+\.|\d+,))))\d{{1,3}}({thousandsmark}\d{{3}})+(?={placeholder})";
-		public const string PlaceHolderDefault = "\\D|\\b";
+		public const string NoLetterPrefixRegex = @"(?<![a-z|0-9])";
+		public const string NoLetterSuffixRegex = @"(?![a-z|0-9])";
+		public static readonly string GeneralPhoneNumberRegex = $@"({NoLetterPrefixRegex}((\d[\s-/]?){{8,}}){NoLetterSuffixRegex})";
+		public static readonly string BrazilPhoneNumberRegex = $@"({NoLetterPrefixRegex}(\(?\s?\d{{2,3}}\s?\)?\s?)\s?\d{{4,5}}-?\d{{3,4}}{NoLetterSuffixRegex})";
+		public static readonly string UkPhoneNumberRegex = $@"({NoLetterPrefixRegex}((00)|\+)?\s?\d{{2}}\s?(\(0\))?\d{{3,5}}[/-]?\s?(\d{{7,8}}|\d{{3}}\s?\d{{3}}){NoLetterSuffixRegex})";
+		public static readonly string GermanyPhoneNumberRegex = $@"({NoLetterPrefixRegex}(\d{{4}}\s?[-/]?(\s?\d){{7,8}}){NoLetterSuffixRegex})";
 	}
 }
--- a/.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
+++ b/.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
@@ -18,11 +18,9 @@ namespace Microsoft.Recognizers.Definitions
 	public static class BasePhoneNumbers
 	{
 		public const string NumberReplaceToken = "@builtin.phonenumber";
-		public const string NoLetterPrefixRegex = @"(?<![a-z|0-9])";
-		public const string NoLetterSuffixRegex = @"(?![a-z|0-9])";
-		public static readonly string GeneralPhoneNumberRegex = $@"({NoLetterPrefixRegex}((\d[\s-/]?){{8,}}){NoLetterSuffixRegex})";
-		public static readonly string BrazilPhoneNumberRegex = $@"({NoLetterPrefixRegex}(\(?\s?\d{{2,3}}\s?\)?\s?)\s?\d{{4,5}}-?\d{{3,4}}{NoLetterSuffixRegex})";
-		public static readonly string UkPhoneNumberRegex = $@"({NoLetterPrefixRegex}((00)|\+)?\s?\d{{2}}\s?(\(0\))?\d{{3,5}}[/-]?\s?(\d{{7,8}}|\d{{3}}\s?\d{{3}}){NoLetterSuffixRegex})";
-		public static readonly string GermanyPhoneNumberRegex = $@"({NoLetterPrefixRegex}(\d{{4}}\s?[-/]?(\s?\d){{7,8}}){NoLetterSuffixRegex})";
+		public const string GeneralPhoneNumberRegex = @"(\b((\d[\s-/]?){8,})\b)";
+		public const string BrazilPhoneNumberRegex = @"(((?'brac'\B\(\s?)|\b)\d{2,3}(?'-brac'\s?\))?\s?)\s?\d{4,5}-?\d{3,5}\b(?(brac)(?!))";
+		public const string UkPhoneNumberRegex = @"(((\b(00)|\B\+)\s?)?\d{2}\s?(\(0\))?\d{3,5}[/-]?\s?(\d{7,8}|\d{3}\s?\d{3})\b)";
+		public const string GermanyPhoneNumberRegex = @"(\b(\d{4}\s?[-/]?(\s?\d){7,8})\b)";
 	}
 }
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequence_English.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequence_English.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Recognizers.Text.Sequence.Tests
         [TestMethod]
         public void PhoneNumberModel()
         {
-            TestNumber();
+            TestSequence();
         }
     }
 }
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequence_English.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequence_English.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Recognizers.Text.DataDrivenTests;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.Recognizers.Text.Sequence.Tests
+{
+    [TestClass]
+    public class TestSequence_English : TestBase
+    {
+        public static TestResources TestResources { get; protected set; }
+        public static IDictionary<string, IModel> Models { get; protected set; }
+
+        [ClassInitialize]
+        public static void ClassInitialize(TestContext context)
+        {
+            TestResources = new TestResources();
+            TestResources.InitFromTestContext(context);
+            Models = new Dictionary<string, IModel>();
+        }
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            base.TestSpecInitialize(TestResources);
+            base.ModelInitialize(Models);
+        }
+
+        [DataSource("Microsoft.VisualStudio.TestTools.DataSource.CSV", "PhoneNumberModel-English.csv", "PhoneNumberModel-English#csv", DataAccessMethod.Sequential)]
+        [TestMethod]
+        public void PhoneNumberModel()
+        {
+            TestNumber();
+        }
+    }
+}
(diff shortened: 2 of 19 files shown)


Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -0,0 +1,958 @@
...
+{
+  "Input": "My phone number is ( 19 ) 38294427.",
+  "NotSupported": "javascript",
+  "NotSupportedByDesign": "python",
+  "Results": [
+    {
+      "Text": "( 19 ) 38294427",
+      "TypeName": "phonenumber",
+      "Resolution": {
+        "value": "( 19 ) 38294427"
```
yiylin: I think the format of Resolution should always the same. But in this file, some are "( xx ) xxxx", "(xx) xxxxx", "(xx)xxxx-xx"

.NET/Microsoft.Recognizers.Text.Sequence/English/Extractors/PhoneNumberExtractor.cs
```diff
@@ -0,0 +1,37 @@
...
+        internal override ImmutableDictionary<Regex, string> Regexes { get; }
+
+        protected sealed override string ExtractType { get; } = Constants.MODEL_PHONE_NUMBER;
+
+        public PhoneNumberExtractor()
+        {
+            var regexes = new Dictionary<Regex, string>
+            {
+                {
+                    new Regex(BasePhoneNumbers.BrazilPhoneNumberRegex), "BrazilPhoneNumber"
```
yiylin: Put these constant string to some constant file

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequence_English.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            base.TestSpecInitialize(TestResources);
+            base.ModelInitialize(Models);
+        }
+
+        [DataSource("Microsoft.VisualStudio.TestTools.DataSource.CSV", "PhoneNumberModel-English.csv", "PhoneNumberModel-English#csv", DataAccessMethod.Sequential)]
+        [TestMethod]
+        public void PhoneNumberModel()
+        {
+            TestNumber();
```
yiylin: why TestNumber?

Specs/Sequence/English/PhoneNumberModel.json
yiylin: @tellarin

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -11,9 +13,34 @@ private SequenceRecognizer(SequenceOptions options)
         {
             RegisterModel(Culture.English, options.ToString(), new Dictionary<Type, IModel>
             {
+                [typeof(PhoneNumberModel)] = new PhoneNumberModel(new PhoneNumberParser(), new PhoneNumberExtractor())
```
yiylin: If for different languages, it always use the same dictionary, I think we can just reuse it? not need to new so many  objects?

.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BaseSequenceExtractor.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace Microsoft.Recognizers.Text.Sequence
+{
+    public abstract class BaseSequenceExtractor : IExtractor
+    {
+        internal abstract ImmutableDictionary<Regex, string> Regexes { get; }
+
+        protected virtual string ExtractType { get; } = "";
+
+        public List<ExtractResult> Extract(string text)
```
yiylin: split big methods to small methods

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequence_English.cs
STayinloves: I think the Sequence's check process is like the Number one, so I just reuse it here. Should I create a new function with the same code or create a function to wrap it.

Specs/Sequence/English/PhoneNumberModel.json
STayinloves: @tellarin and I have discussed to leave it as the extracted result for now. And I think that parse it to some normal format need further discussion and a lot of work.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -320,6 +320,11 @@ public void TestDateTimeMergedParser()
             }
         }
 
+        public void TestSequence()
+        {
+            TestNumber();
```
yiylin: If TestNumber can be used in NumberModel,OrdinalModel,PercentModel,NumberRangeModel and PhoneNumberModel, I suggest to rename TestNumber to more general method(such as TestBasic?). And not need to create a new method TestSequence

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
```diff
@@ -18,7 +18,11 @@ namespace Microsoft.Recognizers.Definitions
 	public static class BasePhoneNumbers
 	{
 		public const string NumberReplaceToken = "@builtin.phonenumber";
-		public static readonly Func<string, string, string> IntegerRegexDefinition = (placeholder, thousandsmark) => $@"(((?<!\d+\s*)-\s*)|((?<=\b)(?<!(\d+\.|\d+,))))\d{{1,3}}({thousandsmark}\d{{3}})+(?={placeholder})";
-		public const string PlaceHolderDefault = "\\D|\\b";
+		public const string NoLetterPrefixRegex = @"(?<![a-z|0-9])";
```
yiylin: Do we need to consider "A-Z"? If so, please add the related test case.

Specs/Sequence/English/PhoneNumberModel.json
tellarin: For this version, let's not do any formatting of the value. This can come in a future PR.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
tellarin: Is NumberWithUnit also using the same TestNumber method? If so, I agree with @yiylin.

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
tellarin: Which cases actually match this? Letters are no valid as prefix, nor suffix.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
yiylin: No, but the NumberWithUnit's test code is very similar to TestNumber, and I think after adding a "if" to TestNumber , then NumberWithUnit also can use this method to test

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
tellarin: Also, can't we just restrict the boundaries of the regexes instead of making all regexes more complex by appending more patterns to their start and end?

.NET/Microsoft.Recognizers.Text.Sequence/English/Extractors/PhoneNumberExtractor.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    class PhoneNumberExtractor : BaseSequenceExtractor
+    {
+        internal override ImmutableDictionary<Regex, string> Regexes { get; }
+
+        protected sealed override string ExtractType { get; } = Constants.MODEL_PHONE_NUMBER;
+
+        public PhoneNumberExtractor()
+        {
+            var regexes = new Dictionary<Regex, string>
```
tellarin: Shouldn't these be in a BasePhoneNumberExtractor?

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
STayinloves: It's for getting rid of the invalid phone number, such as "s6666666", "6666666s" or something else. And I'm going to add the "A-Z".

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -9,10 +11,22 @@ public class SequenceRecognizer : Recognizer
 
         private SequenceRecognizer(SequenceOptions options)
         {
-            RegisterModel(Culture.English, options.ToString(), new Dictionary<Type, IModel>
+            var models = new Dictionary<Type, IModel>
```
tellarin: Wouldn't a better approach be to just override the ways to get a model to ignore culture?

Patterns/Base-PhoneNumbers.yaml
```diff
@@ -1,7 +1,19 @@
...
+GeneralPhoneNumberRegex: !nestedRegex
+  def: ({NoLetterPrefixRegex}((\d[\s-/]?){8,}){NoLetterSuffixRegex})
+  references: [ NoLetterPrefixRegex, NoLetterSuffixRegex ]
+BrazilPhoneNumberRegex: !nestedRegex
+  def: ({NoLetterPrefixRegex}(\(?\s?\d{2,3}\s?\)?\s?)\s?\d{4,5}-?\d{3,4}{NoLetterSuffixRegex})
+  references: [ NoLetterPrefixRegex, NoLetterSuffixRegex ]
+UkPhoneNumberRegex: !nestedRegex
+  def: ({NoLetterPrefixRegex}((00)|\+)?\s?\d{2}\s?(\(0\))?\d{3,5}[/-]?\s?(\d{7,8}|\d{3}\s?\d{3}){NoLetterSuffixRegex})
+  references: [ NoLetterPrefixRegex, NoLetterSuffixRegex ]
+GermanyPhoneNumberRegex: !nestedRegex
```
tellarin: If you changed the YAML file, there should be a .cs and a .ts files created from it.
Can you add the .ts file to this PR?

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
yiylin: If so, then I think it should be [^0-9]

Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -0,0 +1,958 @@
...
+        "TypeName": "phonenumber",
+        "Resolution": {
+            "value": "017612345678"
+        }
+        }
+    ]
+}
+,
+{
+    "Input": "My phone number is 622 15 31 23.",
```
tellarin: All Chile/Argentina numbers fall into the general regex?

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
yiylin: @tellarin  mention that something like "\b"

tellarin: Yep. Like https://docs.microsoft.com/en-us/dotnet/standard/base-types/anchors-in-regular-expressions#word-boundary-b

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
tellarin: Let's keep this one like this then. In the future there will be more cases where there will be differences from sequences to numbers.
But I'd rename TestSequence to TestPhone as this is specific to it.

Specs/Sequence/English/PhoneNumberModel.json
STayinloves: They do fall into it.

Patterns/Base-PhoneNumbers.yaml
STayinloves: OK, I'll add it.

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
STayinloves: The '\b' does extract "49176 12345678" from "+49176 12345678" and "55) 3333-33333" from "(55) 3333-33333", and I think it's not suitble for this situation.

.NET/Microsoft.Recognizers.Text.Sequence/English/Extractors/PhoneNumberExtractor.cs
STayinloves: Yes, I agree. But does BasePhoneNumberExtractor inherit from BaseSequenceExtractor? Because of the same extract logic they share.

Patterns/Base-PhoneNumbers.yaml
STayinloves: I think we should create a new package to contain it. For now, it doesn't exist. And the typescript part in our pattern document seems to be out of date.

.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BaseSequenceExtractor.cs
STayinloves: I think the logic here is hard to split, but I have added some comments. Hope this do help.

Issue #324, opened 2018-02-07
[* Sequence] Support for full US phone number extraction


tellarin:

STayinloves:
Should we just support the regex `@"(\+1)?\s?\(?\d{3}\)?\s?-?\s?\d{3}\s?-?\s?\d{4}\b"` and avoid the invalid situation?

tellarin:
Numbers should match in full, and only in full. Also, no leading/trailing whitespaces.

PR #325, opened 2018-02-07
Support for full US phone number extraction

STayinloves:
Github issue: #324


Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -1,4 +1,34 @@
...
+      "TypeName": "phonenumber",
+      "Resolution": {
+        "value": "1 (877) 609-2233"
+      }
+    }
+  ]
+}
+,
+{
+  "Input": "My phone number is +1 541-754-3010.",
```
tellarin: Fail case is missing.

PR #327, opened 2018-02-08
Moving post build to robocopy 

tellarin:
- Cleanup;
- A little more robust, but this needs a better solution later.


.NET/Samples/SimpleConsole/SimpleConsole.csproj
```diff
@@ -76,4 +76,8 @@
     </ProjectReference>
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <PropertyGroup>
+    <PostBuildEvent>
+</PostBuildEvent>
```
yiylin: not to change this, right?

PR #336, opened 2018-02-09
[.NET] API Signatures Refactor

pcostantini:
**Changelog**
- Remove static creation methods from Recognizers classes
- Recognizer instances wrap model options
- New static methods for parsing directly (less verbose)
- Samples & READMEs updates with new API signatures

**Usage**
_Instead of:_

```C#
var model = DateTimeRecognizer.GetInstance(DateTimeOptions.SplitDateAndTime).GetDateTimeModel(Culture.English);
var result = model.Parse("tomorrow");
```

_Use:_

```C#
var model = new DateTimeRecognizer(Culture.English, DateTimeOptions.SplitDateAndTime).GetDateTimeModel();
var result = model.Parse("tomorrow");
```

_Or in its shorter form:_
```C#
var result = DateTimeRecognizer.RecognizeDateTime("tomorrow", Culture.English, DateTimeOptions.SplitDateAndTime);

// And for the default (none) options:
var result = DateTimeRecognizer.RecognizeDateTime("tomorrow", Culture.English);
```


.NET/Microsoft.Recognizers.Text/Recognizer.cs
```diff
@@ -1,47 +1,39 @@
...
-}
+﻿using System;
+
+namespace Microsoft.Recognizers.Text
+{
+    public abstract class Recognizer<TModelOptions>
+    {
+        private readonly ModelFactory<TModelOptions> factory;
+
+        public string CurrentCulture { get; private set; }
```
tellarin: This new refactor doesn't cover all previous requirements. 
One Recognizer instance can contain models for multiple cultures, so CurrentCulture should belong in each model, not per-recognizer.
Moreover one of the key previously supported functionalities is now missing. If a recognizer class is initialized with no culture parameters, models for all supported cultures should be created and kept in its internal cache. This is a behaviour utilized by many consumers of the project that we can't remove.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -1,173 +1,69 @@
...
-    public class DateTimeRecognizer : Recognizer
+    public class DateTimeRecognizer : Recognizer<DateTimeOptions>
     {
-
-        private DateTimeOptions instanceOptions;
-
-        public static DateTimeOptions Convert(int value)
-        {
-            return EnumUtils.Convert<DateTimeOptions>(value);
-        }
```
tellarin: These enum Convert methods cannot be removed. In many cases external users of the recognizers do not have direct visibility of the Enum and need to safe cast and int to it. This behaviour is caused by the fact that these are flagged enums.

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
```diff
@@ -1,124 +1,175 @@
...
+        }
+
+        private static List<ModelResult> RecognizeByModel(Func<NumberRecognizer, IModel> getModelFunc,  string query, string culture, NumberOptions options)
+        {
+            var recognizer = new NumberRecognizer(culture, options);
+            var model = getModelFunc(recognizer);
+            return model.Parse(query);
+        }
+
+        protected override void InitializeConfiguration()
```
tellarin: Initialize can't always instantiate all supported models. There are startup and memory consumption costs associated with it. In some scenarios, consumers only want to pay the cost for a particular culture.

.NET/Microsoft.Recognizers.Text/ModelFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+                var factoryMethod = this[key];
+                var model = (T)factoryMethod(options);
+
+                // Store in cache
+                cache[cacheKey] = model;
+
+                return model;
+            }
+
+            throw new ArgumentException($"Could not find Model with the specified configuration: {culture}, {typeof(T).ToString()}");
```
tellarin: The behaviour to fallback to a default culture was also removed, but it is required.
For example, in the PhoneNumber case, any culture should return the English model as it is a base model, non culture specific.
Also, in some other cases, if a culture doesn't have it's own model, it is desirable to fallback to English to at least get basic entity recognition. For example: trying to get a Russian date model, that doesn't exist, but still recognize "2018-12-13" as a date.

.NET/Microsoft.Recognizers.Text/Recognizer.cs
tellarin: What about instead of exposing the recognizer constructors, have clear factory methods? 
Something like var recog = CreateRecognizer(culture) and CreateRecognizer(fallbackCulture). The first returns a single specific culture and the second an instance with all supported cultures and possibly a fallback culture?
Names are only to illustrate the idea. Parameters could be passed differently. And method names + code documentation would make each behaviour explicit.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumberRecognizer_English.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.Recognizers.Text.DataDrivenTests.Tests
+{
+    [TestClass]
+    public class TestNumberRecognizer_English
+    {
+        [TestMethod]
+        public void RecognizeNumber()
+        {
+            var actual = NumberRecognizer.RecognizeNumber("192", Culture.English);
```
tellarin: The purpose of these new tests is to make sure the new Recognize*() methods work, right? But could we also make them data-driven (i.e. read json specs like the others)?
I imagine we'll have more tests in the future to cover other cases, like culture fallback for example, or "en" vs "en-US" cases; and it would be good to cover all supported cultures, not just test the English impl.

tellarin: Not a major issue though. Feel free to ignore it if the effort is too high for the benefit now.

.NET/Microsoft.Recognizers.Text/Recognizer.cs
JuanAr: We updated the recognizer constructors to support the `fallbackCulture` scenario and to contain multiple culture models.

```C#
string DefaultCulture = Culture.English;
var recognizer = new DateTimeRecognizer(DefaultCulture);
Assert.AreEqual(recognizer.GetDateTimeModel(DefaultCulture), recognizer.GetDateTimeModel());
Assert.AreNotEqual(recognizer.GetDateTimeModel(Culture.Spanish), recognizer.GetDateTimeModel());
```

Also, by default the recognizer uses lazy model initialization which can be turned off using the `lazyInitialization` parameter, forcing the recognizer to initialize all registered models:

```C#
var recognizer = new DateTimeRecognizer(DefaultCulture, DateTimeOptions.None, lazyInitialization: false);
```

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
JuanAr: We included constructor overloads to allow options to be passed as `int` values to the recognizer:
```C#
protected Recognizer(string defaultCulture, int options, bool lazyInitialization = true)
    : this(defaultCulture, EnumUtils.Convert<TModelOptions>(options), lazyInitialization)
{
}
```

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
JuanAr: The `InitializeConfiguration()` method actually registers all supported culture models and stores a `Dictionary` of lamba factories to support lazy model creation.

```C#
RegisterModel<NumberModel>(
    Culture.English,
    (options) => new NumberModel(
        AgnosticNumberParserFactory.GetParser(AgnosticNumberParserType.Number, new EnglishNumberParserConfiguration()),
        English.NumberExtractor.GetInstance(NumberMode.PureNumber)));
```

.NET/Microsoft.Recognizers.Text/ModelFactory.cs
JuanAr: We changed the `Recognizer<>` initialization to specify the `defaultCulture` (i.e. fallbackCulture) and moved the `culture` parameter to the `GetModel()` method. This way, if a given `culture` does not exists we fallback to the recognizer's default as shown in the following sample:
```C#
var recognizer = new DateTimeRecognizer(DefaultCulture);
Assert.AreEqual(recognizer.GetDateTimeModel(InvalidCulture), recognizer.GetDateTimeModel());
```

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumberRecognizer_English.cs
JuanAr: Yes, actually they helped us designing the Recognize*() methods and serve as usage examples. We agree with you that we might want to convert them into data-driven tests.
We will keep the unit tests updates in our radar and we'll probably create a new branch and PR.
It would be great to discuss the best approach for testing the SDK including all the latest changes.

tellarin: Thanks. A separate PR would be great.
When you say "SDK" you mean more than just the Recognizers-Text project, correct? We can start a thread about that.

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
JuanAr: Just a minor clarification first: model _factories_ (i.e. very simple lambas) are registered always. Instances are created on first use controller by the `lazyInitialization` param.
We updated the constructor to initialize the culture specified models and maintain the previous behavior.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumberRecognizer_English.cs
JuanAr: 👍

.NET/Microsoft.Recognizers.Text/Recognizer.cs
```diff
@@ -1,47 +1,54 @@
...
+using System;
+using System.Linq;
+
+namespace Microsoft.Recognizers.Text
+{
+    public abstract class Recognizer<TModelOptions> where TModelOptions : struct
+    {
+        private readonly ModelFactory<TModelOptions> factory;
+
+        public string DefaultCulture { get; private set; }
```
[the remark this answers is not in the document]
JuanAr: We renamed the Recognizer constructor param to `targetCulture` as the "preferred" culture.
When getting a model, the user can specify a different one or let the Recognizer use the previously specified one.
If no culture is specified, models will fallback to English culture.

```C#
[TestMethod]
public void WithoutTargetCultureAndWithoutCulture_FallbackToEnglishCulture()
{
    var recognizer = new DateTimeRecognizer();
    Assert.AreEqual(recognizer.GetDateTimeModel(), recognizer.GetDateTimeModel(EnglishCulture));
}
```

.NET/Microsoft.Recognizers.Text/Recognizer.cs
```diff
@@ -1,47 +1,54 @@
...
+                throw new ArgumentException("culture", "The default culture is required");
+            }
+
+            this.DefaultCulture = defaultCulture;
+            this.Options = options;
+
+            this.factory = new ModelFactory<TModelOptions>();
+            InitializeConfiguration();
+
+            if (!lazyInitialization) this.InitializeModels(options);
```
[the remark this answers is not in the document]
JuanAr: 👍

.NET/Microsoft.Recognizers.Text/ModelFactory.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (string.IsNullOrEmpty(selectedCulture))
+            {
+                throw new ArgumentNullException("culture", "Culture is required.");
+            }
+
+            if (TryGetModel(selectedCulture, options, out T model))
+            {
+                return model;
+            }
+            else if (TryGetModel(defaultCulture, options, out model))
```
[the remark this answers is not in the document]
JuanAr: We named it `fallbackCulture = Culture.English` and it now belongs to the `ModelFactory`.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -1,173 +1,74 @@
...
-
-        public static DateTimeRecognizer GetCleanInstance()
-        {
-            return new DateTimeRecognizer();
-        }
-
-        public static DateTimeRecognizer GetInstance(DateTimeOptions options = DateTimeOptions.None)
+        public DateTimeRecognizer(string defaultCulture, DateTimeOptions options = DateTimeOptions.None, bool lazyInitialization = true)
```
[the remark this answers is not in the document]
JuanAr: We included back the `fallbackToDefaultCulture` parameter in the `Get***Model()` methods:
```C#
public DateTimeModel GetDateTimeModel(string culture = null, bool fallbackToDefaultCulture = true)
{
    return GetModel<DateTimeModel>(culture, fallbackToDefaultCulture);
}
````

File: .NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs [hunk 10]
```diff
@@ -103,37 +103,38 @@ public static IModel GetModel(this TestContext context)
         {
             var language = TestUtils.GetCulture(context.FullyQualifiedTestClassName);
```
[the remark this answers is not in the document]
JuanAr: We changed all appearances of the `language` variable to `culture`.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumberRecognizerInitialization.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class TestNumberRecognizerInitialization
+    {
+        private readonly string DefaultCulture = Culture.English;
+        private readonly string SpanishCulture = Culture.Spanish;
+        private readonly string InvalidCulture = "vo-id";
+
+        [TestMethod]
+        public void WithoutCulture_UseDefaultCulture()
+        {
+            var recognizer = new NumberRecognizer(DefaultCulture);
```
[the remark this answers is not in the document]
JuanAr: We included new tests to 1) Create a recognizer with no target culture and then 2) Get a model with no specific culture. This result in 3) an English culture model to be created.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
JuanAr: We create a generic method for convertin `int` to the corresponding options `enum`:
```C#
public static TModelOptions GetOption(int value) => EnumUtils.Convert<TModelOptions>(value);
```

.NET/Microsoft.Recognizers.Text/Recognizer.cs
```diff
@@ -1,47 +1,47 @@
...
+        }
+
+        protected abstract void InitializeConfiguration();
+
+        private void InitializeModels(string targetCulture, TModelOptions options)
+        {
+            this.factory.InitializeModels(targetCulture, options);
+        }
+
+        public static TModelOptions GetOption(int value) => EnumUtils.Convert<TModelOptions>(value);
```
tellarin: Shouldn't it be GetOptions?

.NET/Microsoft.Recognizers.Text/Recognizer.cs
```diff
@@ -1,47 +1,47 @@
...
-            return modelContainer.GetSingleModel<TModel>();
-        }
-    }
-}
+﻿using Microsoft.Recognizers.Text.Utilities;
+using System;
+
+namespace Microsoft.Recognizers.Text
+{
+    public abstract class Recognizer<TModelOptions> where TModelOptions : struct
```
tellarin: Nit: I'd prefer TRecognizerOptions as it is at the recognizer level.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -1,173 +1,84 @@
...
+                    new BaseMergedParser(new FrenchMergedParserConfiguration(options)),
+                    new BaseMergedExtractor(new FrenchMergedExtractorConfiguration(options))));
+
+            RegisterModel<DateTimeModel>(
+                Culture.Portuguese,
+                (options) => new DateTimeModel(
+                    new BaseMergedParser(new PortugueseMergedParserConfiguration(options)),
+                    new BaseMergedExtractor(new PortugueseMergedExtractorConfiguration(options))));
+
+            RegisterModel<DateTimeModel>(
```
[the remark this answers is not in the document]
JuanAr: We removed the German model registration from DateTimeRecognizer. We left all the Model, Parser, Extractor and Configuration classes and the data-driven unit tests placeholders. Also, marked all DateTime German specs as NotSupported in dotnet.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/NumberWithUnitRecognizer.cs
```diff
@@ -1,292 +1,321 @@
...
+        private static List<ModelResult> RecognizeByModel(Func<NumberWithUnitRecognizer, IModel> getModelFunc, string query, NumberWithUnitOptions options)
+        {
+            var recognizer = new NumberWithUnitRecognizer(options);
+            var model = getModelFunc(recognizer);
+            return model.Parse(query);
+        }
+
+        protected override void InitializeConfiguration()
+        {
+            RegisterModel<CurrencyModel>(
```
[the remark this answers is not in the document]
JuanAr: Added ` #region`  only for those registrations with more than one model per culture.

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -1,32 +1,67 @@
...
-        public IModel GetIpAddressModel(string culture, bool fallbackToDefaultCulture = true)
+        public IModel GetPhoneNumberModel(string culture = null, bool fallbackToDefaultCulture = true)
         {
-            return GetModel<IpAddressModel>(Culture.English, fallbackToDefaultCulture, SequenceOptions.None.ToString());
+            return GetModel<PhoneNumberModel>(culture, fallbackToDefaultCulture);
         }
 
+        public IModel GetIpAddressModel(string culture = null, bool fallbackToDefaultCulture = true)
+        {
+            return GetModel<IpAddressModel>(culture, fallbackToDefaultCulture);
```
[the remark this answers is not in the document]
JuanAr: We've left the method arguments and hardcoded to ` Culture.English` . For PhoneNumber, we'll probably add new cultures in the future. But for IPAddress, should we simply remove the culture and fallback arguments since there won't be a culture-specific IP Address?

```C#
public IModel GetIpAddressModel(string culture = null, bool fallbackToDefaultCulture = true)
{
    return GetModel<IpAddressModel>(Culture.English, fallbackToDefaultCulture);
}
```

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Sequence/TestSequenceRecognizerInitialization.cs
```diff
@@ -0,0 +1,76 @@
...
+        private readonly string InvalidCulture = "vo-id";
+
+        [TestMethod]
+        public void WithoutCulture_UseTargetCulture()
+        {
+            var recognizer = new SequenceRecognizer(EnglishCulture);
+            Assert.AreEqual(recognizer.GetPhoneNumberModel(), recognizer.GetPhoneNumberModel(EnglishCulture));
+        }
+
+        [Ignore]
```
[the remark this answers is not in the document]
JuanAr: 👍

.NET/Microsoft.Recognizers.Text/Recognizer.cs
JuanAr: Since ` TModelOptions`  can be decorated with `[Flags]` attribute. Yes, the method should be named ` GetOptions`.

.NET/Microsoft.Recognizers.Text/Recognizer.cs
JuanAr: Agree. In `ModelFactory<TModelOptions>` we let the ` TModelOptions` naming.

PR #343, opened 2018-02-12
Fix Issue 216 "Support 'within' in datetime extractor"

guom08:


.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
```diff
@@ -178,7 +178,7 @@ public static class DateTimeDefinitions
 		public const string SinceRegex = @"\b(since|after or equal to|starting (from|on|with))\b";
 		public const string AgoRegex = @"\b(ago)\b";
 		public const string LaterRegex = @"\b(later|from now)\b";
-		public const string InConnectorRegex = @"\b(in)\b";
+		public const string InConnectorRegex = @"\b(in|within)\b";
```
tellarin: This is incorrect. The interpretation of in and withing is not the same.

tellarin: "Within 2 days" is a daterange. "in 2 days" is usually not and we shouldn't change it's behaviour.

tellarin:
Please re-create the PR with fixes from a branch in your fork, ok?
I'll close this one as is.

PR #358, opened 2018-02-23
Refactor ResolutionKey to Constants

wgx998877:
Cleanup/Refactor of resolution keys into constants for better consistency and future maintainability.
Small refactoring to TestBase

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Recognizers.Text.DataDrivenTests
         public IDateTimeExtractor Extractor { get; set; }
         public IDateTimeParser DateTimeParser { get; set; }
         public TestModel TestSpec { get; set; }
-        
+
         public void TestSpecInitialize(TestResources resources)
         {
             TestSpec = resources.GetSpecForContext(TestContext);
@@ -103,10 +103,10 @@ namespace Microsoft.Recognizers.Text.DataDrivenTests
                 if (expected.End != 0) Assert.AreEqual(expected.End, actual.End, GetMessage(TestSpec));
 
                 var values = actual.Resolution as IDictionary<string, object>;
-                var listValues = values[ResolutionKey.ValueListKey] as IList<Dictionary<string, string>>;
+                var listValues = values[ResolutionKey.ValueSetKey] as IList<Dictionary<string, string>>;
                 var actualValues = listValues.FirstOrDefault();
 
-                var expectedObj = JsonConvert.DeserializeObject<IList<Dictionary<string, string>>>(expected.Resolution[ResolutionKey.ValueListKey].ToString());
+                var expectedObj = JsonConvert.DeserializeObject<IList<Dictionary<string, string>>>(expected.Resolution[ResolutionKey.ValueSetKey].ToString());
                 var expectedValues = expectedObj.FirstOrDefault();
 
                 CollectionAssert.AreEqual(expectedValues, actualValues, GetMessage(TestSpec));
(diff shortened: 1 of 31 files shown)


.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -69,75 +69,21 @@ public void ParserInitialize(IDictionary<string, IDateTimeParser> parsers)
 
         public void TestNumber()
         {
-            if (TestUtils.EvaluateSpec(TestSpec, out string message))
-            {
-                Assert.Inconclusive(message);
-            }
+            GeneralPreTest();
```
tellarin: Rename to TestPreValidation.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -334,7 +255,10 @@ public void TestIpAddress()
             {
                 Debugger.Break();
             }
+        }
 
+        private void GeneralMainTest(bool testUnit = false, bool testType = false)
```
tellarin: Rename the method to ValidateResults

tellarin: The parameter could be a list of extra keys to check, instead of multiple booleans. In the future as we have more keys, we don't need extra parameters.
By default "value" is always checked.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -324,6 +234,17 @@ public void TestDateTimeMergedParser()
         }
         
         public void TestIpAddress()
+        {
+            GeneralPreTest();
+            GeneralMainTest(testType: true);
+        }
+
+        public void TestSequence()
```
tellarin: Shouldn't this be TestPhoneNumber to match the IpAddress case?

.NET/Microsoft.Recognizers.Text/Models/ResolutionKey.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace Microsoft.Recognizers.Text
+{
+    public static class ResolutionKey
+    {
+        public const string ValueSetKey = "values";
```
tellarin: If the class is already ResolutionKey, no need to repeat the suffix Key for every key. :)

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -14,25 +14,22 @@ public static class Constants
         public const string SYS_DATETIME_SET = "set";
         public const string SYS_DATETIME_DATETIMEALT = "datetimealt";
 
-        //key
-        public const string TimexKey = "timex";
-        public const string CommentKey = "Comment";
-        public const string ModKey = "Mod";
-        public const string TypeKey = "type";
-        public const string IsLunarKey = "isLunar";
```
tellarin: Some of these should stay here. Only the ones that show up in model results should move to your new class.

tellarin: For the ones that stay here, maybe we need to rename them and follow the code standard of this file.

wgx998877: fixed

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -69,75 +69,21 @@ public void ParserInitialize(IDictionary<string, IDateTimeParser> parsers)
...
-            {
-                var expected = tuple.Item1;
-                var actual = tuple.Item2;
-
-                Assert.AreEqual(expected.TypeName, actual.TypeName, GetMessage(TestSpec));
-                Assert.AreEqual(expected.Text, actual.Text, GetMessage(TestSpec));
-
-                Assert.AreEqual(expected.Resolution["value"], actual.Resolution["value"], GetMessage(TestSpec));
-            }
+            ValidateResults(new List<string>() { });
```
tellarin: Shouldn't the default just be the empty list?

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -334,7 +256,10 @@ public void TestIpAddress()
             {
                 Debugger.Break();
             }
+        }
 
+        private void ValidateResults(List<string> TestResolutionKeys)
```
tellarin: Rename TestResolutionKeys to extraResolutionKeys

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -348,14 +273,16 @@ public void TestIpAddress()
...
-                Assert.AreEqual(expected.Resolution["value"], actual.Resolution["value"], GetMessage(TestSpec));
-                Assert.AreEqual(expected.Resolution["type"], actual.Resolution["type"], GetMessage(TestSpec));
-            }
-        }
+                Assert.AreEqual(expected.Resolution[ResolutionKey.Value], actual.Resolution[ResolutionKey.Value], GetMessage(TestSpec));
 
-        public void TestSequence()
-        {
-            TestNumber();
+                if (TestResolutionKeys != null)
```
tellarin: This would be unnecessary if the default is the empty list.

Issue #269, opened 2018-01-10
[EN Number] Support written negative numbers


tellarin:
negative one
minus thirty-two
"minus three point five"

In English and Chinese.

STayinloves:
It seems that Chinese already has its negative numbers("负三十二“ etc) in its own [parser](https://github.com/Microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.Number/Chinese/Parsers/ChineseNumberParser.cs#L419). But the English use the `BaseNumberParse` and doesn't have a negative numbers implement.
So should we implement this by creating a specific `EnglishNumberParser` for English or in the `BaseNumberParse`?

tellarin:
Edited the title as support for Chinese is already there.

tellarin:
Reopening as the original implementation had to be reverted due to perf reasons (https://github.com/Microsoft/Recognizers-Text/pull/368).

tellarin:
Perf of latest PR is much better.

PR #361, opened 2018-02-26
Support negative number for English.

STayinloves:
For issue(#269)


Specs/Number/English/NumberModel.json
```diff
@@ -1634,4 +1634,79 @@
   "NotSupportedByDesign": "python",
   "Results": []
 }
+,
+{
+  "Input": "The answer is negative ninety - five hundred fifths",
+  "NotSupportedByDesign": "python",
+  "NotSupported": "javascript",
```
tellarin: These cases should also be supported in JS.

Specs/Number/English/NumberModel.json
```diff
@@ -1634,4 +1634,79 @@
...
+      "Text": "negative ninety - five hundred fifths",
+      "TypeName": "number",
+      "Resolution": {
+        "value": "-1900"
+      }
+    }
+  ]
+}
+,
+{
```
tellarin: Can you add more tests cases like "minus 5", "minus five percent", "minus five point five" and similar?

PR #364, opened 2018-02-27
Recog german - DateTime, Bugfixes, Additional Tests

flo-zuewi:
https://github.com/Microsoft/Recognizers-Text/issues/240


Specs/DateTime/German/MergedExtractor.json
```diff
@@ -25,10 +25,10 @@
   {
     "Input": "Das ganze findet dann am Montag um 4 Uhr statt",
     "NotSupportedByDesign": "python",
-    "NotSupported":"javascript,dotnet",
+    "NotSupported": "javascript,dotnet",
```
tellarin: Will the fixes for these last cases come in a next PR?

Specs/Number/English/NumberModel.json
```diff
@@ -722,6 +722,7 @@
 {
   "Input": " two hundred point zero three",
   "NotSupportedByDesign": "python",
+  "Debug":true,
```
tellarin: Please remove the Debug tags before pushing.

Specs/DateTime/German/DatePeriodParser.json
```diff
@@ -23,26 +23,26 @@
       }
     ]
   },
-  {//Assert.Null
-    "Input": "Ich bin nächsten Monat vom 4. bis zum 23. weg",
```
tellarin: Why was "nächsten Monat" removed? Will support for relative day/month/etc come in a follow up PR?

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -72,6 +73,12 @@ protected override void InitializeConfiguration()
                 (options) => new DateTimeModel(
                     new BaseMergedParser(new PortugueseMergedParserConfiguration(options)),
                     new BaseMergedExtractor(new PortugueseMergedExtractorConfiguration(options))));
+
+            RegisterModel<DateTimeModel>(
+                Culture.German,
```
tellarin: Please don't push the registration of the German DateTime models until all tests are passing, ok?
We only want to expose them when that's true.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -22,6 +22,9 @@ public BaseNumberParser(INumberParserConfiguration config)
         {
             this.Config = config;
 
+            //As soon as you develop on a machine that does not use en-US as standard-culture, numbertests for the other languages fail
```
tellarin: I understand the final correction for this bug is not in the repo yet and this causes issues, but please don't push the code with this workaround.
Sorry for the delay to have a final fix integrated.

.NET/Microsoft.Recognizers.Text.DateTime/German/Parsers/GermanDatePeriodParserConfiguration.cs
```diff
@@ -202,7 +202,7 @@ public bool IsLastCardinal(string text)
         public bool IsMonthOnly(string text)
         {
             var trimedText = text.Trim().ToLowerInvariant();
-            return trimedText.EndsWith("monat");
+            return (trimedText.EndsWith("monat")|| trimedText.EndsWith("monate")|| trimedText.EndsWith("monaten")|| trimedText.EndsWith("monats"));
```
tellarin: Nit: Code standard: Whitespaces around operators (||)

Specs/DateTime/German/DatePeriodParser.json
flo-zuewi: The Phrase "Nächsten Monat" is fully working, just went further down (Or is now in DateTimeModel). I removed it to quickly test the DateRange without a specific month given, which did not work at first.
I can add it as separate testcase.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
flo-zuewi: Ok, I added it because most of it is working, except for very few issues which mostly seem to be related to bugs that other languages have too. More details in the report.

Specs/DateTime/German/DatePeriodParser.json
tellarin: Please add as a separate case and other similar situations.
Thanks.

Specs/DateTime/German/MergedExtractor.json
flo-zuewi: There is an issue with the Set-Extractor. Sometimes testcases are falsely detected as sets, which causes some tests to fail. I was not able to find the reason for this behaviour.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
tellarin: I see many tests marked as not supported and the DateTimeModelSplitDateAndTime and DateTimeModelCalendarMode specs are empty.
I didn't have time to look into detail.
Can you also look at the Choice recognizer and the new NumberRange type?

PR #367, opened 2018-03-01
Port C# Sequence Recognizers to JavaScript

amitstein:
Ported C# Sequence Recognizers to JavaScript

As part of porting the C# Sequence recognizers to JavaScript, a couple of regular expression definitions in the Base-PhoneNumbers.yaml file were updated:

- Implemented an equivalent regular expression for BrazilPhoneNumberRegEx with the overall same results but compatible with both C# and JavaScript (the original one didn't work quite well in JS).
- Fixed what seemed a bug in the GeneralPhoneNumberRegex definition which made the resulting capture include an undesired trailing whitespace.


amitstein:
Hi @tellarin 
Thank you for letting me know  :)

PR #372, opened 2018-03-02
Refactor Chinese GetCustomNumberModel

JuanAr:
This updates are based on the changes made to the `test-refactor` branch since tests is where the GetCustomNumberModel  methods where originally placed. This PR should be merged only after the tests refactor are merged into `master`.

In this PR the logic to create the Custom Chinese Number Model we move into the Recognizer + Configuration + Extractors _pipeline_.


.NET/Microsoft.Recognizers.Text.Number/NumberOptions.cs
```diff
@@ -6,5 +6,6 @@ namespace Microsoft.Recognizers.Text.Number
     public enum NumberOptions
     {
         None = 0,
+        ChineseNumbersWithPercent = 1
```
tellarin: I don't think this is something we should surface at the recognizer level. This is more an implementation detail of one of the Chinese number models. 
We'll do a closer analysis of the change and update the PR review soon.

JavaScript/test/runner-number.js
```diff
@@ -10,7 +10,7 @@ var modelFunctions = {
     'PercentModel': (input, culture, options) => Recognizer.recognizePercentage(input, culture, options, false),
     // TODO: Implement number range model in javascript
     'NumberRangeModel': (input, culture, options) => null,
-    'CustomNumberModel': (input, culture, options) => getCustomNumberModel(culture).parse(input)
```
tellarin: If custom model was only used to test this particular case, I think we can remove it.
The tests for the Chinese Units model will already make use of it.

.NET/Microsoft.Recognizers.Text.Number/NumberOptions.cs
tellarin: Changes in PR https://github.com/Microsoft/Recognizers-Text/pull/379

JuanAr: ACK. Please, let me know if we should solve conflicts for this PR or Close it without merging.

tellarin: Let's close this one.
I'll remove the custom model test in another PR as it only target one case, and it's tested by the unit model tests anyways. It's easier that way.

tellarin:
Closing as discussed. Custom test will be removed as it doesn't add much value.

PR #374, opened 2018-03-03
Implementing Japanese ChoiceRecognizer

AniChikage:
First pull request for Japanese version under .Net.

Progress:
1. add Japanese version for "Choice".
2. add common true/false phrase, which are
(1).true: はい|そうです|そう|よい.
(2).false: いいえ|ではありません|ではない|じゃない|じゃありません.


Patterns/Japanese/Japanese-Choice.yaml
```diff
@@ -0,0 +1,9 @@
+﻿---
+LangMarker: Eng
+#Boolean Recognizers
```
tellarin: This file was not updated with the Japanese values used to generate the ChoiceDefinitions.cs file. Please update them here.
Also, please add the generated definitions file in Typescript.

.NET/Microsoft.Recognizers.Definitions/Japanese/ChoiceDefinitions.cs
```diff
@@ -0,0 +1,25 @@
...
+// </auto-generated>
+//------------------------------------------------------------------------------
+namespace Microsoft.Recognizers.Definitions.Japanese
+{
+	using System;
+	using System.Collections.Generic;
+
+	public static class OptionsDefinitions
+	{
+		public const string LangMarker = "Jp";
```
tellarin: Please change the LangMarker in the YAML file to "Jpn" and re-generate this file.

Patterns/Japanese/Japanese-Choice.yaml
AniChikage: Got it.

Issue #363, opened 2018-02-27
EN DateTimeV2 - year's quarters are not identified 


amotzhoshen:
Hi,
For the input utterance "I'll be back during Q1" - I'd expect to have Q1 identified as a daterange entity.
This is relevant (but less common) also to H1 / H2 - indicating first/second half of the year.

Thanks!

yaronism:
Cool. we also faced this issue with LUIS. @tellarin any estimation when it'll be solved?

tiviti:
I have also encountered this issue!

tellarin:
Hi @amotzhoshen, @yaronism, and @tiviti,
preliminary support for Q1-4 has been merged. We'll add H1/H2 and it should soon enter the release train.
I'll update this issue once it's released.

yaronism:
Super cool. Thanks for your prompt support!

tellarin:
Support for H1/H2 is now also in. This will be included in the next release.

PR #380, opened 2018-03-06
Support year halves in English

STayinloves:
GitHub issue #363
Support year halves in English
- fix month overflow when creating date in "Q4"
- add tests
- remove QuarterRegexYearBack as it's unused.


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1032,7 +1082,7 @@ private DateTimeResolutionResult ParseQuarter(string text, DateObject referenceD
             }
 
             var beginDate = DateObject.MinValue.SafeCreateFromValue(year, quarterNum * 3 - 2, 1);
-            var endDate = DateObject.MinValue.SafeCreateFromValue(year, quarterNum * 3 + 1, 1);
+            var endDate = DateObject.MinValue.SafeCreateFromValue(year, quarterNum * 3, 1).AddMonths(1);
```
tellarin: It would be better to change all cases to use such methods instead of manually adding/subtracting months.
Also, please declare constants to represent 3 and 6 in such equations. 3 can be TrimesterMonthCount and 6 SemesterMonthCount or similar.
You can also simplify the calculations and place inside a helper method that safely computes start/end of the period you need.

JavaScript/packages/recognizers-date-time/src/dateTime/baseDatePeriod.ts
```diff
@@ -892,6 +896,52 @@ export class BaseDatePeriodParser implements IDateTimeParser {
         return result;
     }
 
+    // Resolve month overflow
+    private createDate(year: number, month: number, day: number): Date {
```
tellarin: Instead of doing it this way, create another method in DateUtil that safely adds the period you want to a date.

PR #384, opened 2018-03-07
Fix 4:00 - 5:00 pm interpreted as 4am to 5pm

STayinloves:
GitHub issue https://github.com/Microsoft/Recognizers-Text/issues/386

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
@@ -154,14 +154,15 @@ namespace Microsoft.Recognizers.Text.DateTime
                     else if (!string.IsNullOrEmpty(pmStr) || rightPmValid)
                     {
 
-                        if (beginHour < 12)
-                        {
-                            beginHour += 12;
-                        }
-
                         if (endHour < 12)
                         {
                             endHour += 12;
+                        }
+
+                        // Resolve case like "11 to 3pm"
+                        if (beginHour + 12 < endHour)
+                        {
+                            beginHour += 12;
                         }
 
                         isValid = true;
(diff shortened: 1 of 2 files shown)


Specs/DateTime/English/TimePeriodParser.json
```diff
@@ -1027,4 +1027,160 @@
...
+{
+  "Input": "please help me set up a meeting from 1:30 to 3pm",
+  "Context": {
+    "ReferenceDateTime": "2017-12-01T13:37:00"
+  },
+  "NotSupported": "javascript",
+  "NotSupportedByDesign": "python",
+  "Results": [
+    {
+      "Text": "from 1:30 to 3pm",
```
yiylin: do we support "from 1:30am to 3"? if we support, please add these kind of test case

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
```diff
@@ -137,15 +137,20 @@ private DateTimeResolutionResult ParseSimpleCases(string text, DateObject refere
...
+                        }
+
+                        if (beginHour >= 12 && beginHour - 12 < endHour)
                         {
                             beginHour -= 12;
                         }
 
-                        if (endHour >= 12)
+                        // Resolve case like "11 to 3am"
+                        if (beginHour < 12 && beginHour > endHour)
```
yiylin: It is better that the PR title include the Issue number. Then I can check the details of the bug

tellarin: In the description, not in the title. :)

Specs/DateTime/English/TimePeriodParser.json
```diff
@@ -899,6 +899,32 @@
   ]
 }
 ,
+{
+  "Input": "please help me set up a meeting from 1:30am to 3",
```
tellarin: Can we also have the other cases?
If possible, please write different senteces too.

Specs/DateTime/English/TimePeriodParser.json
yiylin: I prefer to add more cases. "from 11am to 2", "from 11pm to 1", "from 0:01am to 1pm", "from 0:01am to 1". Not sure if some of them has been added

STayinloves: Previously I regarded that the error only occurred when "am/pm" is a suffix, now I have fixed all cases and add tests.

tellarin:
Thanks, @STayinloves.

PR #387, opened 2018-03-07
Small cleanup/fixing in test specs

tellarin:
- Fix Japanese Choice tests;
- Removing the CustomNumberModel tests that don't add much value;
- Spreading test cases around.


tellarin:
Thanks, @wgx998877 for the help with the Chinese samples. ;-)

PR #396, opened 2018-03-11
Add Japanese-Numbers.yaml

AniChikage:
add numbers pattern for Japanese text.


Patterns/Japanese/Japanese-Numbers.yaml
```diff
@@ -0,0 +1,293 @@
+﻿---
+#JapaneseNumberParserConfiguration
+LangMarker: 'Jpn'
+DecimalSeparatorChar: !char '.'
+FractionMarkerToken: ''
```
tellarin: This doesn't apply in Japanese?

.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.csproj
```diff
@@ -17,4 +17,8 @@
     <ProjectReference Include="..\Microsoft.Recognizers.Text\Microsoft.Recognizers.Text.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Japanese\" />
```
tellarin: You don't need to add this in the project file. Please remove.

Patterns/Japanese/Japanese-Numbers.yaml
```diff
@@ -0,0 +1,293 @@
...
+    ５: '5'
+    ６: '6'
+    ７: '7'１
+    ８: '8'
+    ９: '9'
+    ／: '/'
+    －: '-'
+    ，: \'
+    Ｇ: G
+	ｇ: g
```
tellarin: This looks like some encoding issue, can you remove it?

Patterns/Japanese/Japanese-Numbers.yaml
```diff
@@ -1,11 +1,6 @@
 ﻿---
 #JapaneseNumberParserConfiguration
 LangMarker: 'Jpn'
-DecimalSeparatorChar: !char '.'
```
tellarin: It's best not to delete items from the YAML file, unless really necessary.
The code later may try to use them and break. Also, in the ideal case, the YAML files between languages should be as similar as possible.

PR #403, opened 2018-03-14
Naïve filtering of frequent false positive extractions of 'May' as month

tellarin:


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DateParser.cs
```diff
@@ -689,5 +691,10 @@ private int ConvertChineseYearToInteger(string yearChsStr)
 
             return year < 10 ? -1 : year;
         }
+
+        public List<DateTimeParseResult> FilterResults(string query, List<DateTimeParseResult> candidateResults)
+        {
+            return candidateResults;
```
wgx998877: What's that function used?

tellarin: Only the basemeged parser really needs it, but in this version it needs to exist in the interface,
This is a naïve fix.

Issue #405, opened 2018-03-14
no strong name on NeoSmart.Unicode


taranais:
> No se puede cargar el archivo o ensamblado 'NeoSmart.Unicode, Version=0.1.2.0, Culture=neutral, PublicKeyToken=null' ni una de sus dependencias. Se necesita un ensamblado con nombre seguro. (Excepción de HRESULT: 0x80131044)

https://github.com/neosmart/unicode.net/issues/1

tellarin:
@pablocastro and @JuanAr can you look into it? It happens with the nuget.org packages.

taranais:
Unicode.net in not signed and Recognizers-Text it`s

JuanAr:
Can [Delay Signing](https://docs.microsoft.com/en-us/dotnet/framework/app-domains/delay-sign-assembly) to unblock until NeoSmart.Unicode is signed?

tellarin:
@JuanAr nuget packages need to have their libs already signed. 
Users of the packages can opt for something like https://docs.microsoft.com/en-us/dotnet/framework/app-domains/how-to-disable-the-strong-name-bypass-feature, but we should fix the strong naming issue with the dependency, or change to another dependency.

tellarin:
@taranais, the dependency was removed. New packages will be out ASAP.

PR #411, opened 2018-03-16
Remove NeoSmart.Unicode (no strong named) dependency form Choice*

JuanAr:
We implemented the only piece of logic we used from the NeoSmart.Unicode package within the class.
- bool IsEmoji(string letter)
- IEnumerable<string> Letters(string text)


tellarin:
I'd prefer those two methods in a separate UnicodeUtils or something, but this looks OK to me.
Waiting for build to pass.

tellarin:
This addresses issue #405

PR #414, opened 2018-03-16
Fix ipv6 bug for single ‘:’

wgx998877:
fix ipv6 bug for ellipsis
example: sync:

### DIFF
--- a/Specs/Sequence/English/IpAddressModel.json
+++ b/Specs/Sequence/English/IpAddressModel.json
@@ -329,6 +329,41 @@
     "Results": []
   },
   {
+    "Input": "wrong IPV6 address :",
+    "NotSupportedByDesign": "python",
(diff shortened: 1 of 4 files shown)


Specs/Sequence/English/IpAddressModel.json
```diff
@@ -328,6 +328,41 @@
...
+    "NotSupportedByDesign": "python",
+    "Results": []
+  },
+  {
+    "Input": ":as you say",
+    "NotSupportedByDesign": "python",
+    "Results": []
+  },
+  {
+    "Input": ": as you say",
```
tellarin: Can you add a time case, like "the train arrives at 10:00"?

wgx998877: added

PR #420, opened 2018-03-20
Rename namespace and Timex class

johnataylor:
namespace becomes MIcrosoft.Recognizers.Text.DataTypes.TimexExpression
change is reflected everywhere... package, project, folder etc.

class Timex becomes class TimexProperties which is probably a more accurate name.


.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/Timex.cs
```diff
@@ -3,32 +3,32 @@
 
 using System.Collections.Generic;
 
-namespace Microsoft.Recognizers.DataTypes.DateTime
+namespace Microsoft.Recognizers.Text.DataTypes.TimexExpression
 {
-    public class Timex
+    public class TimexProperties
```
tellarin: I always prefer that the file name matches the class name.

.NET/CreateAllPackages.cmd
```diff
@@ -30,7 +30,7 @@ popd
 pushd Microsoft.Recognizers.Text.Choice
 call CreatePackage.cmd
 popd
-pushd Microsoft.Recognizers.DataTypes.DateTime
+pushd Microsoft.Recognizers.Text.DataTypes.TimexExpression
```
tellarin: The new namespace could be Microsoft.Recognizers.Text.DataTypes.TimeExpression, without the X in Timex as it's redundant. But I'm fine with keeping it.

.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/Timex.cs
tellarin: Shouldn't TimexProperties be singluar?

.NET/CreateAllPackages.cmd
johnataylor: should be good now

PR #429, opened 2018-03-22
Fix "please book 2 hours next week" returning either blank resolution or two entities "2 hours next" and "next week"

AniChikage:
.Net and JS bug are both fixed.
tests files have bugs, I'll fix it tomorrow and add more test cases.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedExtractor.cs
@@ -110,8 +110,17 @@ namespace Microsoft.Recognizers.Text.DateTime
                     //}
                     if (dst[i].IsOverlap(result))
                     {
+                        int nextCursor = i + 1;
+                        var isCrossOverlap = false;
+                        if (nextCursor < dst.Count)
+                        {
+                            if (dst[nextCursor].IsCrossOverlap(result))
+                            {
+                                isCrossOverlap = true;
+                            }
+                        }
                         isFound = true;
-                        if (dst[i].IsCover(result))
+                        if (dst[i].IsCover(result) && !isCrossOverlap)
                         {
                             if (firstIndex == -1)
                             {
--- a/JavaScript/packages/recognizers-text/src/extractors.ts
+++ b/JavaScript/packages/recognizers-text/src/extractors.ts
@@ -17,6 +17,14 @@ export class ExtractResult {
         return ((er2.start < er1.start) && ((er2.start + er2.length) >= (er1.start + er1.length)))
         || ((er2.start <= er1.start) && ((er2.start + er2.length) > (er1.start + er1.length)));
     }
+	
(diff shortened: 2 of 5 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedExtractor.cs
```diff
@@ -110,8 +110,17 @@ private void AddTo(List<ExtractResult> dst, List<ExtractResult> src, string text
                     //}
                     if (dst[i].IsOverlap(result))
                     {
+                        int nextCursor = i + 1;
+                        var isCrossOverlap = false;
+                        if (nextCursor < dst.Count)
+                        {
+                            if (dst[nextCursor].IsCrossOverlap(result))
```
tellarin: i don't think this should be treated as overlap. "2 hours next" is an incorrect extraction.

tellarin: Also, in one of the modes, the recognition is already correct, but with no resolution.
Wouldn't it be better to just calculate the resolution and fix why the behaviour is different across modes?

JavaScript/packages/recognizers-text/src/extractors.ts
```diff
@@ -17,6 +17,14 @@ export class ExtractResult {
         return ((er2.start < er1.start) && ((er2.start + er2.length) >= (er1.start + er1.length)))
         || ((er2.start <= er1.start) && ((er2.start + er2.length) > (er1.start + er1.length)));
     }
+	
+	static isCrossOverlap(er1: ExtractResult, er2: ExtractResult): boolean {
```
tellarin: Please always add test cases to show that the problem was fixed. Also try to add other similar cases and cases that may be ambiguous to show that the fix doesn't break other situations, ok?
Thanks.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedExtractor.cs
```diff
@@ -110,8 +110,17 @@ private void AddTo(List<ExtractResult> dst, List<ExtractResult> src, string text
...
+                        if (nextCursor < dst.Count)
+                        {
+                            if (dst[nextCursor].IsCrossOverlap(result))
+                            {
+                                isCrossOverlap = true;
+                            }
+                        }
                         isFound = true;
-                        if (dst[i].IsCover(result))
+                        if (dst[i].IsCover(result) && !isCrossOverlap)
```
tellarin: It looks like this change is also breaking other test cases.
If you look at the build fail log, this now fails:
Failed   MergedExtractor (Data Row 3)
Error Message:
 Assert.AreEqual failed. Expected:<1>. Actual:<2>. Input: "this is before tomorrow 4pm "

Please always run all test cases for all platforms (dotnet, javascript, python) before creating the PR.

Issue #431, opened 2018-03-23
[* Unit] Incorrectly identifying percentages in URL encoded strings


tellarin:
Ex: http://proquest.umi.com/pqdweb?RQT=305&SQ=issn%280024%2D9114%29%20and%20%28ti%28Using%203D%20CAD%20to%20design%20a%20dog%29%20or%20startpage%28158%29%29%20and%20volume%2872%29%20and%20issue%289%29%20and%20pdn%28%3E01%2F01%2F2000%20AND%20%3C12%2F31%2F2000%29&clientId=17859

tellarin:
If the number is preceded by a “%” it still identifies 23%
Ex: https://www.test.com/search?q=c%23%20ignore%20property%20during%20twenty%20forty%20lero. 

In fact, it identifies percentage from all the following situations that might well be a part of an encoded URL:
1.	%twenty% - identifies 20%
2.	%twenty%forty% - identifies 20%

PR #434, opened 2018-03-29
Fix incorrectly identifying percentages in URL encoded strings

STayinloves:
GitHub issue #431 

Not support javascript for its lack of support of look-behind regex.


Patterns/English/English-Numbers.yaml
```diff
@@ -110,7 +110,7 @@ CurrencyRegex: !simpleRegex
   def: (((?<=\W|^)-\s*)|(?<=\b))\d+\s*(B|b|m|t|g)(?=\b)
 #Percentage Regex
 NumberWithSuffixPercentage: !nestedRegex
-  def: ({BaseNumbers.NumberReplaceToken})(\s*)(%|per cents|per cent|cents|cent|percentage|percents|percent)
+  def: ({BaseNumbers.NumberReplaceToken})(\s*)(%(?!{BaseNumbers.NumberReplaceToken})|(per cents|per cent|cents|cent|percentage|percents|percent)\b)
```
tellarin: Isn't the lookback going to cause a big performance hit?
Would just a word boundary check solve the issue?

STayinloves: I'm afraid adding word boundaries can't solve this issue because the percentage extractor transforms the text source when extract and add special characters like `@`.

tellarin: Not sure I get it, but OK. What about the performance hit of the current solution?

PR #460, opened 2018-04-18
Add Ordinal/Percentage/Cardinal recognition for Japanese

AniChikage:
This is new PR based on former "Add Ordinal/Percentage recognition for Japanese" PR, adding Cardinal recognition.


.NET/Microsoft.Recognizers.Text.Number/Japanese/Parsers/JapaneseNumberParser.cs
```diff
@@ -0,0 +1,365 @@
...
+            else if (extra.Contains("Integer"))
+            {
+                ret = ParseInteger(getExtResult);
+            }
+            else if (extra.Contains("Ordinal"))
+            {
+                ret = ParseOrdinal(getExtResult);
+            }
+            else
+
```
tellarin: This is an "else if", right? If so, please remove the blank lines.

Specs/Number/Japanese/NumberModel.json
```diff
@@ -0,0 +1,378 @@
+[
+{
+  "Input": "西九条",
```
tellarin: Do we have the same number of sentences in the Japanese specs as in the Chinese ones? Covering the same cases?

.NET/Microsoft.Recognizers.Text.Number/Utilities/PatternUtil.cs
```diff
@@ -0,0 +1,285 @@
+﻿using Microsoft.Recognizers.Text.Number.Chinese;
+using Microsoft.Recognizers.Text.Number.Japanese;
+using System.Text;
+
+namespace Microsoft.Recognizers.Text.Number
+{
+    public class PatternUtil
```
tellarin: I don't think these methods should be here. They should be in a base CJK parser class, called from the Chinese and Japanese parsers, and use the configs in the parsers. Similar to how it works for English and other "related" languages.

.NET/Microsoft.Recognizers.Text.Number/Chinese/Parsers/ChineseNumberParser.cs
```diff
@@ -32,62 +32,48 @@ public override ParseResult Parse(ExtractResult extResultChs)
...
             }
             else if (extra.Contains("Frac"))
             {
-                ret = FracParseChs(simplifiedExtResultChs);
+                ret = FracParseChs(simplifiedExtResult);
             }
             else if (extra.Contains("Dou"))
             {
-                ret = DouParseChs(simplifiedExtResultChs);
+                ret = DouParseChs(simplifiedExtResult);
```
tellarin: Ideally methods should not have language prefixes. DouParseChs should not end with Chs.
Also, please rename the methods to be more easily readable. This one could be ParseDouble.

Specs/Number/Japanese/NumberModel.json
AniChikage: Yes, "西九条" is a Japanese place name, the "WhiteListRegex" doesn't contains "条", so it only can be recognized in "ExtractAll" mode in "IntegerExtractors.cs".
But I haven't add "四川" Chinese specs into Japanese ones.

Issue #474, opened 2018-04-23
[EN DateTimeV2.Timezone] Add coverage for missing timezone names


tellarin:
"Pacific daylight savings Time” -> UTC-7
"AWDT",  "Austrialian West Daylight Time" -> UTC+9
"COT", "Colombia Time", -> UTC-5
"HKT", "Hong Kong Time" -> UTC+8

Also, confirm that all offsets for timezone abbreviations that are not ambiguous are correct.
Ambiguous abbreviation should still map to XX.

tellarin:
This page can be used to double check the timezone offsets: https://www.timeanddate.com/time/zones/

PR #481, opened 2018-04-23
[EN DateTimeV2.Timezone] Add coverage for missing timezone names

AniChikage:


tellarin:
Partially covers https://github.com/Microsoft/Recognizers-Text/issues/474

Issue #207, opened 2017-11-23
[FR Number] The common written form "40 000" should be recognized as number 40000


tellarin:
"Input": "Siegel payé national et 94 000 $ shuster pour laisser tomber toutes les réclamations."

PR #482, opened 2018-04-24
Support common written form like "40 000" correctly recognized as number

sanxing-chen:
Github issue #207

### DIFF
--- a/Specs/Number/English/NumberModel.json
+++ b/Specs/Number/English/NumberModel.json
@@ -1698,4 +1698,30 @@
     }
   ]
 }
+,
+{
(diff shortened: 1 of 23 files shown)


Specs/Number/English/NumberModel.json
```diff
@@ -1698,4 +1698,30 @@
     }
   ]
 }
+,
+{
+  "Input": "1 234 567",
```
tellarin: Looks good. But can you add a few more full sentence examples for English?
Also, cases with regular spaces and cases with narrow-space.
Thanks.

PR #489, opened 2018-04-25
Support no break space as thousand mark in number

STayinloves:
#207

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
@@ -126,6 +126,9 @@ namespace Microsoft.Recognizers.Text.Number
         // 1 234 567
         public static LongFormatType IntegerNumBlank = new LongFormatType(' ', '\0');
 
+        // 1 234 567
+        public static LongFormatType IntegerNumNoBreakSpace = new LongFormatType(' ', '\0');
+
         // 1'234'567
         public static LongFormatType IntegerNumQuote = new LongFormatType('\'', '\0');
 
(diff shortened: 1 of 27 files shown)


.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
```diff
@@ -138,9 +141,15 @@ public class LongFormatType
...
         public static LongFormatType DoubleNumBlankComma = new LongFormatType(' ', ',');
 
+        // 1 234 567,89
+        public static LongFormatType DoubleNumNoBreakSpaceComma = new LongFormatType(' ', ',');
+
         // 1 234 567.89
         public static LongFormatType DoubleNumBlankDot = new LongFormatType(' ', '.');
 
+        // 1 234 567.89
+        public static LongFormatType DoubleNumNoBreakSpaceDot = new LongFormatType(' ', '.');
```
tellarin: It would be nice to define a constant somewhere to represent this special space char. And add some code comment for people to realize what it is. Even better if we can write the unicode code and convert, so file encoding issues don't break it.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -673,7 +673,7 @@ protected double GetDigitalValue(string digitStr, double power)
             for (var i = 0; i < digitStr.Length; i++)
             {
                 var ch = digitStr[i];
-                if (!isFrac && (ch == Config.NonDecimalSeparatorChar || ch == ' '))
+                if (!isFrac && (ch == Config.NonDecimalSeparatorChar || ch == ' ' || ch == ' '))
```
tellarin: Same constant comment applies here. Otherwise any reader of the code will be confused with the "duplicate" check.

JavaScript/packages/recognizers-number/src/number/models.ts
```diff
@@ -21,6 +21,9 @@ export class LongFormatType {
     // 1 234 567
     static readonly integerNumBlank = new LongFormatType(' ', '\0');
 
+    // 1 234 567
+    static readonly integerNumNoBreakSpace = new LongFormatType(' ', '\0');
```
tellarin: Same.

JavaScript/packages/recognizers-number/src/number/models.ts
```diff
@@ -33,9 +36,15 @@ export class LongFormatType {
     // 1 234 567,89
     static readonly doubleNumBlankComma = new LongFormatType(' ', ',');
 
+    // 1 234 567,89
+    static readonly doubleNumNoBreakSpaceComma = new LongFormatType(' ', ',');
```
tellarin: Same.

PR #510, opened 2018-05-01
Add Japanese NumberWithUnit(Age)

AniChikage:


.NET/Microsoft.Recognizers.Definitions/Japanese/NumbersWithUnitDefinitions.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.Recognizers.Definitions.Japanese
+{
+	using System;
+	using System.Collections.Generic;
+
+	public static class NumbersWithUnitDefinitions
+	{
+		public static readonly IList<string> AgeAmbiguousValues = new List<string>
+		{
+			"歳"
```
tellarin: Can you add a comment in the YAML file describing what this means in English? Or examples of the ambiguity?

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26730.12
+VisualStudioVersion = 15.0.27004.2009
```
tellarin: No need to revert it now, but don't change version unless really needed.

Issue #117, opened 2017-09-26
[EN DateTimeV2] Does not recognize "less than 24 hours"


Dillagent:
If I ask a question like  "____________ in the LAST 24 hours"  datetime entity works correctly.  However, asking "______ for less than 24 hours."

tellarin:
@Dillagent, we've added this case to the list to be supported in the near future. 
Thanks for the feedback.

tellarin:
This and similar durations should be supported as durations with modifier.

PR #514, opened 2018-05-02
Handle "more than" and "less than" modifiers in datetime

STayinloves:
Github issue #117

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
@@ -57,6 +57,9 @@ namespace Microsoft.Recognizers.Text.DateTime
         public const string MID_MOD = "mid";
         public const string LATE_MOD = "end";
 
+        public const string MORE_THAN_MOD = "more than";
+        public const string LESS_THAN_MOD = "less than";
+
         // Invalid year
         public const int InvalidYear = int.MinValue;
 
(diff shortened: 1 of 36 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -57,6 +57,9 @@ public static class Constants
         public const string MID_MOD = "mid";
         public const string LATE_MOD = "end";
 
+        public const string MORE_THAN_MOD = "more than";
```
tellarin: Wouldn't "more" and "less" be better?

tellarin: Can you also add a code comment describing in which situation they would be used?

JavaScript/packages/recognizers-date-time/src/dateTime/constants.ts
```diff
@@ -45,4 +45,6 @@ export class TimeTypeConstants {
     static readonly beforeMod: string = "before";
     static readonly afterMod: string = "after";
     static readonly sinceMod: string = "since";
+    static readonly moreThanMod: string = "more than";
```
tellarin: Same.

Specs/DateTime/English/DateTimeParser.json
```diff
@@ -1242,4 +1242,29 @@
     }
   ]
 }
+,
+{
+  "Input": "I will be busy in more than an hour, so call me later",
```
tellarin: This is not a "proper" sentence. Can you change it to something like "I'll be free in less than an hour, so call me later"?

Issue #503, opened 2018-04-28
[* DateTimeV2] Regression in handling 2-digit years in date


tellarin:
Cases like "5/5/89" or "05/05/89" are no longer  recognized.

tellarin:
This should be interpreted as 5/5/1989 if the YY part falls within a given range.

tellarin:
Or at least provide two possible resolutions.

tellarin:
Chosen solution was to go back to previous range of 2-digit years to resolve to 19XX or 20XX.

PR #516, opened 2018-05-04
Improve handling of 2-digit years in date parser

STayinloves:
GitHub issue #503 

- revise js resource generator
- move min/maxYearNum to BaseDateTime
- add tests


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -512,11 +516,11 @@ private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDat
                 if (!string.IsNullOrEmpty(yearStr))
                 {
                     year = int.Parse(yearStr);
-                    if (year < 100 && year >= 90)
+                    if (year < 100 && year >= Constants.MinTwoDigitYearPastNum)
```
tellarin: This method could be somewhere shared between the Chinese parser and the base parser. It would make it simpler to keep implementations consistent.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -572,15 +576,7 @@ private DateTimeResolutionResult ParseWeekdayOfMonth(string text, DateObject ref
...
-            int cardinal;
-            if (this.config.IsCardinalLast(cardinalStr))
-            {
-                cardinal = 5;
-            }
-            else
-            {
-                cardinal = this.config.CardinalMap[cardinalStr];
-            }
+            var cardinal = this.config.IsCardinalLast(cardinalStr) ? 5 : this.config.CardinalMap[cardinalStr];
```
tellarin: If we declare a constant for the meaning of 5 here, wouldn't it make the code easier to read?

Issue #477, opened 2018-04-23
[EN DatetimeV2] Can't recognize "between 2007 and 2012" as daterange


guom08:
Can't recognize "between 2007 and 2012" as daterange

tellarin:
"from may to june" also currently doesn't work.
But "from may 1st to june 2nd" works.

PR #519, opened 2018-05-04
Extend recognition of dateranges across date types

guom08:
#477 
Mainly focus on improve cases like
(from|between...) {DateRange|DatePoint} (till|to|and...) {DateRange|DatePoint}


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -57,17 +83,91 @@ private List<Token> MatchSimpleCases(string text)
...
+                }
+
+                var middleStr = text.Substring(middleBegin, middleEnd - middleBegin).Trim().ToLowerInvariant();
+                var match = this.config.TillRegex.Match(middleStr);
+                if (match.Success && match.Index == 0 && match.Length == middleStr.Length)
+                {
+                    var periodBegin = er[idx].Start ?? 0;
+                    var periodEnd = (er[idx + 1].Start ?? 0) + (er[idx + 1].Length ?? 0);
+
+                    // handle "desde"
```
tellarin: The comments for this if and the next need to be clearer to describe what and why this is needed here.

Also, it looks like this code is the same or very similar to other places. It would be best to refactor and re-use the code instead of copy/paste. It will make fixing bugs easier in the future.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -57,17 +83,91 @@ private List<Token> MatchSimpleCases(string text)
...
+                    var beforeStr = text.Substring(0, periodBegin).Trim().ToLowerInvariant();
+                    if (this.config.GetFromTokenIndex(beforeStr, out int fromIndex)
+                        || this.config.GetBetweenTokenIndex(beforeStr, out fromIndex))
+                    {
+                        periodBegin = fromIndex;
+                    }
+
+                    ret.Add(new Token(periodBegin, periodEnd));
+                    idx += 2;
```
tellarin: Can you add a comment everywhere "magic" numbers like this happen? The comment should describe why 2.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -157,6 +93,221 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
...
+                        {
+                            futureBegin = pastBegin;
+                        }
+
+                        if (pastEnd < pastBegin)
+                        {
+                            pastEnd = futureEnd;
+                        }
+
+                        if (isSpecificDate)
```
tellarin: Couldn't the timex generation be a method somewhere for re-use?
It would also make future maintenance easier and avoid introducing inconsistencies throughout the codebase.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -705,8 +858,8 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             {
                 pastEnd = futureEnd;
             }
-            
-            ret.Timex = $"({pr1.TimexStr},{pr2.TimexStr},P{(futureEnd - futureBegin).TotalDays}D)";
```
tellarin: Is this causing some side-effect?
Can we add some test cases for periods like 1.5 days and see if things behave as expected?

guom08: I just checked, it do have some side effects for cases like "I will stay here from today to two and a half days later". I will change this back.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
guom08: Fixed.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
guom08: Added comments to describe this "magic" number.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
guom08: This is fixed.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
guom08: Add add one test case to avoid regression in the future.

Issue #215, opened 2017-11-23
[EN DateTimeV2] Support relative week of days


tellarin:
Sample utterances: "second sunday", "next Sunday", "two sundays from now"

tellarin:
@rubio41 isn't this working?

lucinezhang:
I've tested on the latest code. For now, "next Sunday" is supported, but "second sunday" and "two sundays from now" are not.

STayinloves:
"two sundays from now" still not be recognized in C# and Javascript.

PR #523, opened 2018-05-07
Better support relative week day in English

STayinloves:
GitHub issue #215


Specs/DateTime/English/DateExtractor.json
```diff
@@ -1336,4 +1336,26 @@
     }
   ]
 }
+,
+{
+  "Input": "Are you available on two sundays from now?",
```
tellarin: "Are you available two sundays from now?"

Issue #504, opened 2018-04-28
[EN DatetimeV2] Recognize simple comparison expressions for date/datetime periods


guom08:
later than or at 1/1/2016
after or on 1/1/2016
on or earlier than 1/1/2016
later than or at 1/1/2016

tellarin:
Be careful with MOD attributes and with the behaviour under ExtendedTypes.

PR #537, opened 2018-05-09
Support simple comparison expressions for date/datetime periods

guom08:
#504
Cases like "on or earlier than xxx", "greater than or equal to"
Also fix bug when recognizing "earlier than / after Feb 2018"


.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
```diff
@@ -183,9 +183,9 @@ public static class DateTimeDefinitions
 		public const string TokenBeforeTime = "at ";
 		public const string AMTimeRegex = @"(?<am>morning)";
 		public const string PMTimeRegex = @"\b(?<pm>afternoon|evening|night)\b";
-		public const string BeforeRegex = @"\b(before|in advance of|prior to|(no later|earlier|sooner) than|ending (with|on)|by|till|til|until)\b";
-		public const string AfterRegex = @"\b(after(?!\s+or equal to)|(?<!no\s+)later than)\b";
-		public const string SinceRegex = @"\b(since|after or equal to|starting (from|on|with))\b";
+		public const string BeforeRegex = @"\b(?<include>(on|in|at)\s+or\s+)?(before|in advance of|prior to|(no later|earlier|sooner) than|ending (with|on)|by|till|til|until)(?<include>\s+or\s+(on|in|at))?\b\s*";
```
tellarin: Shouldn't the included piece be it's own regex?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/MergedDateTimeParserChs.cs
```diff
@@ -115,10 +115,10 @@ public new ParseResult Parse(ExtractResult er, DateObject refTime)
                 pr.Value = val;
             }
 
-            pr.Value = DateTimeResolution(pr, hasBefore, hasAfter, hasSince);
+            pr.Value = DateTimeResolution(pr);
 
             //change the type at last for the after or before mode
-            pr.Type = $"{ParserTypeName}.{DetermineDateTimeType(er.Type, hasBefore, hasAfter, hasSince)}";
+            pr.Type = $"{ParserTypeName}.{DetermineDateTimeType(er.Type, hasMod: hasBefore || hasAfter || hasSince)}";
```
tellarin: Please declare a var for hasModifier. It makes code easier to read.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -54,6 +54,7 @@ public static class Constants
         public const string BEFORE_MOD = "before";
```
tellarin: These four MODs should have comments like:
"after" -> To mean "following in time". I.e. Does not include the starting point. Equivalent to ">"
"before" -> To mean "preceding in time". I.e. Does not include the ending point. Equivalent to "<" 
"since" -> Same as "after", but including the start time. Equivalent to ">="
"until" -> Same as "until", but including the end time. Equivalent to "<="

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -34,6 +34,8 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refTime)
 
             // Push, save the MOD string
             bool hasBefore = false, hasAfter = false, hasSince = false, hasYearAfter = false;
+            // "Include" means MOD string like "on or later than", "earlier than or in"
+            bool hasInclude = false;
```
tellarin: A better name would be hasInclusiveModifier.

tellarin: The code comment should also describe the intent of if. Not just a couple specific examples.

tellarin: Nit: Add blank lines before code comments.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -262,15 +292,15 @@ public List<DateTimeParseResult> DateTimeResolutionForSplit(DateTimeParseResult
             }
             else
             {
-                slot.Value = DateTimeResolution(slot, false, false, false);
-                slot.Type = $"{ParserTypeName}.{DetermineDateTimeType(slot.Type, false, false, false)}";
+                slot.Value = DateTimeResolution(slot);
+                slot.Type = $"{ParserTypeName}.{DetermineDateTimeType(slot.Type, false)}";
```
tellarin: Please pass named parameters for values with no context.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -292,6 +321,7 @@ public List<DateTimeParseResult> DateTimeResolutionForSplit(DateTimeParseResult
 
             var islunar = val.IsLunar;
             var mod = val.Mod;
+            var typeOutput = DetermineDateTimeType(slot.Type, hasMod: !string.IsNullOrEmpty(mod));
```
tellarin: What's typeOutput ?

Specs/DateTime/English/MergedExtractor.json
```diff
@@ -993,4 +993,59 @@
     }
   ]
 }
+,
+{
```
tellarin: There seem to be many test cases missing. Can you add a few more situations?

.NET/Microsoft.Recognizers.Definitions/English/DateTimeDefinitions.cs
guom08: (on|in|at) could be a separate Regex, also some other Regex already exists could re-use this (on|in|at) Regex.
Do you have a good name suggestion for this new Regex?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/MergedDateTimeParserChs.cs
guom08: Fixed.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
guom08: Fixed

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
guom08: Fixed

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
guom08: Fixed.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
guom08: This is not newly added, moved this code from above.
With modifier, the output type might not be the type in the original resolution result.
For example, if the resolution type is "date", with modifier, the output type should be "daterange".
Code comment added to describe this.

tellarin: outputType then, instead of typeOutput.

Specs/DateTime/English/MergedParser.json
```diff
@@ -2380,4 +2380,574 @@
...
+  "Input": "He will arrive on 1/1/2016",
+  "Context": {
+    "ReferenceDateTime": "2016-11-07T00:00:00"
+  },
+  "NotSupported": "javascript",
+  "NotSupportedByDesign": "python",
+  "Results": [
+    {
+      "Text": "on 1/1/2016",
+      "Type": "datetimeV2.daterange",
```
tellarin: Type here is incorrect.

Specs/DateTime/English/MergedParser.json
```diff
@@ -2380,4 +2380,574 @@
...
+  "NotSupportedByDesign": "python",
+  "Results": [
+    {
+      "Text": "before or in Feb 2018",
+      "Type": "datetimeV2.daterange",
+      "Value":{
+        "values":[
+          {
+            "timex":"2018-02",
+            "Mod":"until",
```
tellarin: This case seems very weird. Mod is saying the end is inclusive, but it shouldn't include 2018-03-01 as that is not in Feb 2018.

Issue #540, opened 2018-05-10
[* Sequence PhoneNumber] "86 (40) 08423331" can't be extracted correctly


STayinloves:
my phone is 86 (40) 08423331
``` json
{
  "Text": "(40) 08423331",
  "Start": 15,
  "End": 27,
  "TypeName": "phonenumber",
  "Resolution": {
    "value": "(40) 08423331"
  }
}
```

PR #543, opened 2018-05-10
Revise phone number regex

STayinloves:
GitHub issue #540

- enable tests that currently supported
- enable tests for sequence in Portuguese

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
+++ b/.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
@@ -22,7 +22,8 @@ namespace Microsoft.Recognizers.Definitions
 		public const string BrazilPhoneNumberRegex = @"(((\B\(\s?))\d{2,3}(\s?\))|(\b\d{2,3}))\s?\d{4,5}-?\d{3,5}\b";
 		public const string UkPhoneNumberRegex = @"(((\b(00)|\B\+)\s?)?\b\d{2}\s?(\(0\))?\d{3,5}[/-]?\s?(\d{7,8}|\d{3}\s?\d{3})\b)";
 		public const string GermanyPhoneNumberRegex = @"(\b(\d{3,4}\s?[-/]?(\s?\d){7,8})\b)";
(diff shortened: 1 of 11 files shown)


Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -3330,10 +3327,10 @@
     "NotSupportedByDesign": "python",
     "Results": [
       {
-        "Text": "35831647",
```
tellarin: This interpretation was correct.

Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -3343,10 +3340,10 @@
     "NotSupportedByDesign": "python",
     "Results": [
       {
-        "Text": "19 35831647",
```
tellarin: This one I think is a little ambiguous. I'm fine if we extract the number, but I don't think the tagged text should include the mismatched '('.

.NET/Microsoft.Recognizers.Definitions/BasePhoneNumbers.cs
```diff
@@ -22,7 +22,7 @@ public static class BasePhoneNumbers
 		public const string BrazilPhoneNumberRegex = @"(((\B\(\s?))\d{2,3}(\s?\))|(\b\d{2,3}))\s?\d{4,5}-?\d{3,5}\b";
 		public const string UkPhoneNumberRegex = @"(((\b(00)|\B\+)\s?)?\b\d{2}\s?(\(0\))?\d{3,5}[/-]?\s?(\d{7,8}|\d{3}\s?\d{3})\b)";
 		public const string GermanyPhoneNumberRegex = @"(\b(\d{3,4}\s?[-/]?(\s?\d){7,8})\b)";
-		public const string USPhoneNumberRegex = @"((((\B\+)|\b)1(\s|-)?)|\b)?(\B\(|\b)\d{3}\)?\s?-?\s?\d{3}\s?-?\s?\d{4}\b";
+		public const string USPhoneNumberRegex = @"((((\B\+)|\b)\d{1,3}(\s|-)?)|\b)?(\(|\b)\d{2,3}\)?\s?-?\s?\d{3}\s?-?\s?\d{4,5}\b";
```
tellarin: The number in the referenced issue is not a US number, so I don't think a fix for it should change the US regex.
Perhaps add a new one? It's a common format in China and perhaps other places.

PR #557, opened 2018-05-16
[NL choice] Add Dutch ChoiceRecognizer in .NET

iMicknl:
The only issue I still have with the implementation is that I have two failing test, because of score that is not correct. What is the best way to generate the score in the testcases? I took the scores from the failed tests window, however it looks to be something with the rounding of the score.

@ntrogh could you check the Dutch sentences in my testcase? They look solid imo. :-)

### DIFF
--- a/Specs/Choice/Dutch/BooleanModel.json
+++ b/Specs/Choice/Dutch/BooleanModel.json
@@ -39,7 +39,7 @@
         "TypeName": "boolean",
         "Resolution": {
           "value": true,
-          "score": 0.4461538461538466
+          "score": 0.44615384615384618
         }
       }
     ]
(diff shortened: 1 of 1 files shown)


iMicknl:
```
Assert.AreEqual failed. Expected:<0.485714285714286>. Actual:<0.485714285714286>. Input: "Ik zei NEE in plaats van ja!"
```

```
Assert.AreEqual failed. Expected:<0.446153846153847>. Actual:<0.446153846153846>. Input: "Ik denk dat het wel gaat werken, dus ja ik zal het doen."
```

tellarin:
Using the real double value solves the issue. Unfortunately there is currently no better way to do it.
0.44615384615384618 and 0.48571428571428571 seem to do the trick.

PR #569, opened 2018-05-21
Support cases like "during the day" as durations in English

STayinloves:


.NET/Microsoft.Recognizers.Definitions/French/DateTimeDefinitions.cs
```diff
@@ -1,4 +1,6 @@
-﻿//------------------------------------------------------------------------------
+﻿
```
tellarin: Nit: Please don't add unnecessary blank lines.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDurationParser.cs
```diff
@@ -295,6 +295,12 @@ private DateTimeResolutionResult ParseImplicitDuration(string text, DateObject r
                 ret = result;
             }
 
+            if ((config.Options & DateTimeOptions.CalendarMode) != 0 &&
```
tellarin: Can you add a code comment that explains the purpose of the check? Like in the next if.

Specs/DateTime/English/DateTimeModelCalendarMode.json
```diff
@@ -480,5 +480,53 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I'm blocked for the day",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T16:12:00"
+    },
+    "NotSupported": "javascript,python",
```
tellarin: Is the reason for the notsupported in JS caused by another issue?

Specs/DateTime/English/DateTimeModelCalendarMode.json
```diff
@@ -480,5 +480,53 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I'm blocked for the day",
```
tellarin: Are the exact same new cases already covered in DateTimeModel? If not, please add there to make it easier to check coverage and that the behaviour there doesn't change.

STayinloves: It's because these cases only work under the CalendarMode.

Specs/DateTime/English/DateTimeModelCalendarMode.json
STayinloves: For now, most of the DateTime extractors/parsers don't support CalendarMode so this PR just support .NET. These changes can come from another PR.

PR #570, opened 2018-05-24
Filter single ambiguous word which extracted as duration under Calendar Mode

STayinloves:


Patterns/English/English-DateTime.yaml
```diff
@@ -474,6 +474,8 @@ FromToRegex: !simpleRegex
   def: \b(from).+(to)\b.+
 SingleAmbiguousMonthRegex: !simpleRegex
   def: ^(the\s+)?(may|march)$
+SingleAmbiguousTermsRegex: !simpleRegex
+  def: ^(the\s+)?(day|week|month|year)$
```
tellarin: Can you add a comment in the YAML here to explain the reason for this regex?

PR #573, opened 2018-05-28
Support date period with the "later/earlier" prefix

STayinloves:


JavaScript/packages/recognizers-date-time/src/dateTime/baseDatePeriod.ts
```diff
@@ -580,6 +603,15 @@ export class BaseDatePeriodParser implements IDateTimeParser {
                 if (latePrefix) {
                     beginDate = DateUtils.safeCreateFromMinValue(year, 6, 1);
                 }
+                if (earlierPrefix && swift === 0) {
```
tellarin: Nit: Always skip a line between ifs to make reading the code easier.

Specs/DateTime/English/DatePeriodParser.json
```diff
@@ -5121,4 +5121,179 @@
     }
   ]
 }
+,
+{
+  "Input": "Please find us a time to meet earlier in the week",
```
tellarin: Can you change this for a case where there's at least two days in the range?
Also, the sentence doesn't make sense. Maybe something like "we could have scheduled a time to meet earlier in the week"

Specs/DateTime/English/DatePeriodParser.json
```diff
@@ -5121,4 +5121,179 @@
...
+          "startDate": "2018-05-01",
+          "endDate": "2018-05-13"
+        }
+      }
+    }
+  ]
+}
+,
+{
+  "Input": "Please find us a time to meet earlier this year",
```
tellarin: Same.

PR #587, opened 2018-06-04
implement string matcher for recognizers text

wgx998877:


.NET/Microsoft.Recognizers.Text.DataDrivenTests/Matcher/StringMatcherTest.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                Assert.AreEqual(value, stringMatcher.Find(value).Single().Text);
+            }
+        }
+
+        [TestMethod]
+        public void SimpleTestWithIdsStringMatcher()
+        {
+            var values = new List<string>() { "China", "Beijing", "City" };
+            var Ids = new List<string>() { "1", "2", "3" };
```
tellarin: Nit: ids

PR #595, opened 2018-06-06
Add Dutch test specs for Number

iMicknl:
All English test specs for Number (and subtypes) are translated into Dutch. At the moment all marked as `NotSupported` for Python, C# and JavaScript.

Those can be used for testing when implementing the Number parser/recognizer for Dutch. We started on the work in a fork, but it will take quite some time possibly.
It would be nice if this could be merged in already.


tellarin:
@iMicknl it seems the Python spec loader has an issue. A fix is on the way.

PR #596, opened 2018-06-06
Support timezone as the clarification of datetime in English

STayinloves:

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
@@ -214,7 +214,7 @@ namespace Microsoft.Recognizers.Text.DataDrivenTests
                 case DateTimeExtractors.Holiday:
                     return new BaseHolidayExtractor(new EnglishHolidayExtractorConfiguration());
                 case DateTimeExtractors.TimeZone:
-                    return new BaseTimeZoneExtractor(new EnglishTimeZoneExtractorConfiguration());
+                    return new BaseTimezoneExtractor(new EnglishTimeZoneExtractorConfiguration());
                 case DateTimeExtractors.Set:
                     return new BaseSetExtractor(new EnglishSetExtractorConfiguration());
                 case DateTimeExtractors.Merged:
@@ -248,7 +248,7 @@ namespace Microsoft.Recognizers.Text.DataDrivenTests
                 case DateTimeParsers.Holiday:
                     return new BaseHolidayParser(new EnglishHolidayParserConfiguration());
                 case DateTimeParsers.TimeZone:
-                    return new BaseTimeZoneParser();
+                    return new BaseTimezoneParser();
                 case DateTimeParsers.Set:
                     return new BaseSetParser(new EnglishSetParserConfiguration(commonConfiguration));
                 case DateTimeParsers.Merged:
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
@@ -81,10 +81,12 @@ namespace Microsoft.Recognizers.Text.DateTime
         public static readonly int MaxTwoDigitYearFutureNum = int.Parse(BaseDateTime.MaxTwoDigitYearFutureNum);
         public static readonly int MinTwoDigitYearPastNum = int.Parse(BaseDateTime.MinTwoDigitYearPastNum);
 
-        // special value for timezone
+        // Special values for timezone
         public const int InvalidOffsetValue = -10000;
-        public const string UtcOffsetMinsKey = "UtcOffsetMins";
-        public const string ResolveTimeZone = "resolveTimeZone";
+        public const string UtcOffsetMinsKey = "utcOffsetMins";
+        public const string TimezoneText = "timezoneText";
+        public const string Timezone = "timezone";
+        public const string ResolveTimezone = "resolveTimeZone";
         public const int PositiveSign = 1;
         public const int NegativeSign = -1;
 
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
@@ -37,7 +37,41 @@ namespace Microsoft.Recognizers.Text.DateTime
             tokens.AddRange(BeforeAfterRegexMatch(text));
             tokens.AddRange(SpecialCasesRegexMatch(text, reference));
 
-            return Token.MergeAllTokens(tokens, text, ExtractorName);
+            var ret = Token.MergeAllTokens(tokens, text, ExtractorName);
+
+            if ((this.config.Options & DateTimeOptions.EnablePreview) != 0)
+            {
+                AddTimeZone(ret, this.config.TimeZoneExtractor.Extract(text, reference), text);
+            }
+
+            return ret;
+        }
+
+        private void AddTimeZone(List<ExtractResult> ret, List<ExtractResult> timeZoneErs, string text)
+        {
+            foreach (var er in ret)
+            {
+                foreach (var timeZoneEr in timeZoneErs)
+                {
+                    var begin = er.Start + er.Length;
+                    var end = timeZoneEr.Start;
+                    if (begin < end)
+                    {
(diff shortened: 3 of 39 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -81,10 +81,12 @@ public static class Constants
...
 
-        // special value for timezone
+        // Special values for timezone
         public const int InvalidOffsetValue = -10000;
-        public const string UtcOffsetMinsKey = "UtcOffsetMins";
-        public const string ResolveTimeZone = "resolveTimeZone";
+        public const string UtcOffsetMinsKey = "utcOffsetMins";
+        public const string TimezoneText = "timezoneText";
+        public const string Timezone = "timezone";
+        public const string ResolveTimezone = "resolveTimeZone";
```
tellarin: ResolvedTimezone?

Also, let's document exactly what goes in the single timezone entity, and what goes into the merged datetime entity with timezone.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -248,7 +248,7 @@ public static IDateTimeParser GetEnglishParser(DateTimeParsers parserName)
                 case DateTimeParsers.Holiday:
                     return new BaseHolidayParser(new EnglishHolidayParserConfiguration());
                 case DateTimeParsers.TimeZone:
-                    return new BaseTimeZoneParser();
```
tellarin: No need to rename all of these. "Time zone" is actually two words.
If we do this type of renaming, we also need to be sure the files are renamed appropriately, not only the classes.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
```diff
@@ -37,7 +37,41 @@ public virtual List<ExtractResult> Extract(string text, DateObject reference)
...
+
+            if ((this.config.Options & DateTimeOptions.EnablePreview) != 0)
+            {
+                AddTimeZone(ret, this.config.TimeZoneExtractor.Extract(text, reference), text);
+            }
+
+            return ret;
+        }
+
+        private void AddTimeZone(List<ExtractResult> ret, List<ExtractResult> timeZoneErs, string text)
```
tellarin: It's better to return the modified list, than to rely on modifying a parameter.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
```diff
@@ -37,7 +37,41 @@ public virtual List<ExtractResult> Extract(string text, DateObject reference)
             tokens.AddRange(BeforeAfterRegexMatch(text));
             tokens.AddRange(SpecialCasesRegexMatch(text, reference));
 
-            return Token.MergeAllTokens(tokens, text, ExtractorName);
+            var ret = Token.MergeAllTokens(tokens, text, ExtractorName);
```
tellarin: ret could be renamed to make clear what is in there.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
tellarin: This is a "merge", right? So the method name could be MergeTimeZones.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
tellarin: This rename is also inconsistent with some of the new methods in this PR.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
```diff
@@ -37,7 +37,41 @@ public virtual List<ExtractResult> Extract(string text, DateObject reference)
...
+                foreach (var timeZoneEr in timeZoneErs)
+                {
+                    var begin = er.Start + er.Length;
+                    var end = timeZoneEr.Start;
+                    if (begin < end)
+                    {
+                        var midString = text.Substring((int) begin, (int) (end - begin));
+                        if (string.IsNullOrWhiteSpace(midString))
+                        {
+                            var length = (int) (timeZoneEr.Start + timeZoneEr.Length - er.Start);
```
tellarin: newLength instead of length would be more clear. Or mergedLength.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
```diff
@@ -37,7 +37,41 @@ public virtual List<ExtractResult> Extract(string text, DateObject reference)
...
+        {
+            foreach (var er in ret)
+            {
+                foreach (var timeZoneEr in timeZoneErs)
+                {
+                    var begin = er.Start + er.Length;
+                    var end = timeZoneEr.Start;
+                    if (begin < end)
+                    {
+                        var midString = text.Substring((int) begin, (int) (end - begin));
```
tellarin: Wouldn't gapText be clearer than midString?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -338,6 +338,30 @@ public List<DateTimeParseResult> DateTimeResolutionForSplit(DateTimeParseResult
             AddResolutionFields(res, ResolutionKey.Type, typeOutput);
             AddResolutionFields(res, DateTimeResolutionKey.IsLunar, islunar? islunar.ToString():string.Empty);
 
+
```
tellarin: Nit

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -338,6 +338,30 @@ public List<DateTimeParseResult> DateTimeResolutionForSplit(DateTimeParseResult
...
+            var hasTimezone = false;
+            // Add timezone resolution
```
tellarin: Nit: Always add a blank line before a comment.

Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -2054,74 +2054,80 @@
...
-          {
-            "timex": "2017-04",
-            "type": "daterange",
-            "start": "2017-04-01",
-            "end": "2017-05-01"
-          }
-        ]
-      }
-    }]
+    "Results": [
```
tellarin: Are these all just linebreak changes? The format in the repo should be all unix-style line breaks.
Perhaps we need a cleanup PR to standardize all files?

Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -2194,5 +2200,105 @@
...
+        "Start": 43,
+        "End": 72,
+        "TypeName": "datetimeV2.timerange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(T15:30,T18:30,PT3H)",
+              "type": "timerange",
+              "timezone": "UTC+10:00",
+              "timezoneText": "aest",
```
tellarin: Was the test code modified to check if these are correct? I didn't see it.

tellarin: We also need to be sure these new texts *don't* appear in the standalone timezone entity recognitions.

STayinloves: The test code already supports this checking. The new texts don't appear in the standalone timezone resolution, but it seems we can't do this checking in the test code.

Specs/DateTime/English/DateTimeModelComplexCalendar.json
STayinloves: Yes, I agree.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
STayinloves: Added in BaseMergedDateTimeParser.

PR #600, opened 2018-06-07
Korean Number / translated and feature added json files

baemins:
Korean Number / translated and feature added json files

6 json files


Specs/Number/Korean/NumberModel.code-workspace
```diff
@@ -0,0 +1,1919 @@
+[{
```
tellarin: You don't need to include the *code-workspace in the PRs. Just the .json files.

PR #601, opened 2018-06-07
Support <place>-time timezone pattern in English

STayinloves:


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeZoneExtractor.cs
```diff
@@ -24,9 +24,29 @@ public List<ExtractResult> Extract(string text, DateObject reference)
...
             var tokens = new List<Token>();
             tokens.AddRange(TimeZoneMatch(text));
+            tokens.AddRange(CityTimeMatch(text));
             return Token.MergeAllTokens(tokens, text, ExtractorName);
         }
 
+        private IEnumerable<Token> CityTimeMatch(string text)
+        {
+            var ret = new List<Token>();
+            var cityMatchResult = config.CityStringMatcher.Find(text.ToLowerInvariant());
```
tellarin: Should try to match suffix first, then list. Otherwise we pay the cost of matching the full list all the time.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishTimeZoneExtractorConfiguration.cs
```diff
@@ -31,8 +33,14 @@ public class EnglishTimeZoneExtractorConfiguration : BaseOptionsConfiguration, I
 
         public EnglishTimeZoneExtractorConfiguration() : base(DateTimeOptions.None)
         {
+            CityStringMatcher.Build();
```
tellarin: This build cost should only be paid if EnablePreview is on. How are we controlling it? I didn't notice it in the PR.

PR #602, opened 2018-06-07
Korean Number json update

baemins:
Korean Number json update

6 json files


tellarin:
Thanks for the PR, @baemins!
It seems the Python spec loader has an issue. We'll fix it ASAP. Don't worry about the build break for now.

baemins:
I cancelled my previous PR and made new PR without commentary files.
Thank you for your explanation!

BR

PR #608, opened 2018-06-08
Revise URL extractor

STayinloves:


.NET/Microsoft.Recognizers.Definitions/BaseURL.cs
```diff
@@ -17,6 +17,370 @@ namespace Microsoft.Recognizers.Definitions
 
 	public static class BaseURL
 	{
-		public static readonly string URLRegex = $@"(?<=\s|[\'""\(\[]|^)((((https?|ftp):\/\/)?(((www\.)?[-a-zA-Z0-9:%._\+~#=]{{2,256}}\.[a-zA-Z]{{2,6}})|localhost))|(((https?|ftp):\/\/){BaseIp.Ipv4Regex}))\b(:\d{{1,5}})?([/#][-a-zA-Z0-9:%_\+.~#?!&//=]*)?";
+		public const string SchemeRegex = @"((https?|ftp):\/\/)";
+		public const string PortRegex = @"(:\d{1,5})";
+		public const string UrlPrefixRestrictRegex = @"(?<=\s|[\'""\(\[:.]|^)";
```
tellarin: ExtractionRestrictionRegex

.NET/Microsoft.Recognizers.Definitions/BaseURL.cs
```diff
@@ -17,6 +17,370 @@ namespace Microsoft.Recognizers.Definitions
...
+		public static readonly string UrlPrefixRegex = $@"{UrlPrefixRestrictRegex}{SchemeRegex}?[-a-zA-Z0-9:%._\+~#=]{{1,256}}\.";
+		public static readonly string UrlSuffixRegex = $@"{PortRegex}?([/#][-a-zA-Z0-9:%_\+.~#?!&//=]*)?(?!:)";
+		public static readonly string UrlRegex = $@"{UrlPrefixRegex}[a-zA-Z]{{2,18}}{UrlSuffixRegex}";
```
tellarin: We should match this first. If it matches, than check restriction and TldList.

.NET/Microsoft.Recognizers.Definitions/BaseURL.cs
```diff
@@ -17,6 +17,370 @@ namespace Microsoft.Recognizers.Definitions
 
 	public static class BaseURL
 	{
-		public static readonly string URLRegex = $@"(?<=\s|[\'""\(\[]|^)((((https?|ftp):\/\/)?(((www\.)?[-a-zA-Z0-9:%._\+~#=]{{2,256}}\.[a-zA-Z]{{2,6}})|localhost))|(((https?|ftp):\/\/){BaseIp.Ipv4Regex}))\b(:\d{{1,5}})?([/#][-a-zA-Z0-9:%_\+.~#?!&//=]*)?";
+		public const string SchemeRegex = @"((https?|ftp):\/\/)";
+		public const string PortRegex = @"(:\d{1,5})";
+		public const string UrlPrefixRestrictRegex = @"(?<=\s|[\'""\(\[:.]|^)";
+		public static readonly string UrlPrefixRegex = $@"{UrlPrefixRestrictRegex}{SchemeRegex}?[-a-zA-Z0-9:%._\+~#=]{{1,256}}\.";
+		public static readonly string UrlSuffixRegex = $@"{PortRegex}?([/#][-a-zA-Z0-9:%_\+.~#?!&//=]*)?(?!:)";
```
tellarin: Let's also add a comment on the YAML file to say that Suffix here meas port+path+query
And prefix means restrictions+protocol+hostname (without its tld)

.NET/Microsoft.Recognizers.Definitions/BaseURL.cs
tellarin: This will avoid matching the whole list all the time and also allow JS to stop here as there is no matcher there yet.

PR #614, opened 2018-06-12
Fix issues in DateTime model

STayinloves:
- fix "Sydney timezone" can't be extracted correctly
- fix timezone with diacritics charecter
- improve datetimealt extractor
- improve datetime period with timezone


.NET/Microsoft.Recognizers.Text.DateTime/Utilities/StringUtil.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Globalization;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Recognizers.Text.DateTime.Utilities
+{
+    internal class StringUtil
+    {
+        public static string RemoveDiacritics(string str)
```
tellarin: In a next PR, can you move this to the base Text package. Such methods will be useful not only in DateTime.

Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -2300,5 +2300,237 @@
...
+              "end": "2019-05-22 15:00:00"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input":
+      "Cortana, can you please schedule a meeting with Maria either Monday or Tuesday next week between 1pm and 2pm CST",
```
tellarin: Ideally entities here should be extracted as datetimealt. Will this come in a next PR?

PR #618, opened 2018-06-13
Support filter for single ambiguous timezone word

STayinloves:
- minor: move RemoveDiacritics to Text package


Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -2532,5 +2532,14 @@
...
       }
     ]
+  },
+  {
+    "Input": "I want to get the most from this deal.",
+    "Comment": "This test case is for testing the ambiguous single timezone word filter.",
+    "Context": {
+      "ReferenceDateTime": "2018-06-12T00:00:00"
+    },
+    "NotSupportedByDesign": "python,javascript",
```
tellarin: Can you add a case that shows the merge happens? "Let's meet 3PM get at the book shop".

PR #623, opened 2018-06-14
Add Dutch test specs for NumberWithUnit

iMicknl:
3 of the English test specs for Number (and subtypes) are translated into Dutch. At the moment all marked as NotSupported for Python, C#, JavaScript and Java.
Those can be used for testing when implementing the Number with Units parser/recognizer for Dutch. 

Thanks to: Matthijs van der Linde, Lilian van der Lugt & Marc de Droog (without Github)


iMicknl:
@tellarin great, thanks! I wrote a small Node script which added all the unsupported tags programmatically, but looks like it missed only one. 👍

Issue #597, opened 2018-06-06
[EN DateTime Duration] Wrong resolution for "more than 4 days and less than 1 week"


linzijia07:
"find records last for more than 4 days and less than 1 week"

{
    "Text": "more than 4 days and less than 1 week",
    "Start": 22,
    "End": 58,
    "TypeName": "datetimeV2.duration",
    "Resolution": {
      "values": [
        {
          "timex": "P1W4D",
          "type": "duration",
          "value": "950400"
        }
      ]
    }
  }

Actually, "more than 4 days and less than 1 week" should be a range rather than a number.

PR #636, opened 2018-06-20
Fix wrong extraction: more than 4 days and less than 1 week

guom08:
#597 
Expected results:
"more than 4 days", "less than 1 week", separately.


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -35,13 +34,14 @@ public List<ExtractResult> Extract(string text, DateObject reference)
...
 
-            ResolveMoreThanOrLessThanPrefix(text, rets);
-
+            // First MergeMultipleDuration then ResolveMoreThanOrLessThanPrefix so cases like "more than 4 days and less than 1 week" will not be merged into one "multipleDuration"
             if (this.merge)
             {
                 rets = MergeMultipleDuration(rets, text);
             }
 
+            ResolveMoreThanOrLessThanPrefix(text, rets);
```
tellarin: Let's keep the order of parameters the same across methods, otherwise it's harder to understand the code and potentially easier to make mistakes.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -154,12 +154,7 @@ private static List<Token> GetTokenFromRegex(Regex regex, string text)
 
         private List<ExtractResult> MergeMultipleDuration(List<ExtractResult> extractorResults, string text)
```
tellarin: MergeMultipleDuration -> MergeMultipleDurations

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -34,13 +34,14 @@ public List<ExtractResult> Extract(string text, DateObject reference)
...
 
-            ResolveMoreThanOrLessThanPrefix(text, rets);
-
+            // First MergeMultipleDuration then ResolveMoreThanOrLessThanPrefix so cases like "more than 4 days and less than 1 week" will not be merged into one "multipleDuration"
             if (this.merge)
             {
                 rets = MergeMultipleDuration(rets, text);
             }
 
+            ResolveMoreThanOrLessThanPrefix(text, rets);
```
tellarin: What does "Resolve" means in this case? Maybe just call this method by a better name?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -151,7 +154,7 @@ private static List<Token> GetTokenFromRegex(Regex regex, string text)
             return ret;
         }
 
-        private List<ExtractResult> MergeMultipleDuration(List<ExtractResult> extractorResults, string text)
+        private List<ExtractResult> MergeMultipleDuration(string text, List<ExtractResult> extractorResults)
```
tellarin: What's "text"?

Issue #590, opened 2018-06-06
[EN DateTime] Cannot recognize "from 2010 – 2020"


linzijia07:
"show me sales from 2010 – 2020 and not 2015"
The recognizer cannot recognize "from 2010 – 2020", but it can recognize "from 2010 - 2020".

PR #642, opened 2018-06-21
Fix Issue: Cannot recognize "from 2010 – 2020"

guom08:
#590
Simple changes to add two more special charactor for "to/till"


JavaScript/packages/recognizers-date-time/src/resources/englishDateTime.ts
```diff
@@ -8,8 +8,8 @@
 
 import { BaseDateTime } from "./baseDateTime";
 export namespace EnglishDateTime {
-	export const TillRegex = `(?<till>to|till|til|until|thru|through|--|-|—|——|~)`;
-	export const RangeConnectorRegex = `(?<and>and|through|to|--|-|—|——)`;
+	export const TillRegex = `(?<till>to|till|til|until|thru|through|--|-|—|——|~|–)`;
```
tellarin: This is an 'n-dash', a common char for ranges. Can you test if using \u2013 is enough to make the JS port work?
The Boolean recognizers had a similar issue with some unicode chars..

tellarin: BTW, we should probably add it to the equivalent number range case.

PR #645, opened 2018-06-22
Fix bugs in DateTime's set extractor

STayinloves:
- minor: refine QuarterTermRegex
- add test cases

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -4018,4 +4018,47 @@
     }
   }]
 }
+,
+{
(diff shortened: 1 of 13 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -4018,4 +4018,47 @@
...
+        {
+          "timex": "XXXX-05-09",
+          "type": "date",
+          "value": "2019-05-09"
+        }
+      ]
+    }
+  },
+  {
+    "Text": "nights",
```
tellarin: This should extract "2 nights". Let's open an issue for a better recognition in this case.

PR #650, opened 2018-06-25
Filter unspecific date periods

STayinloves:


Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -3326,14 +3326,6 @@
       }
     ]
   },
-  {
```
STayinloves: This is a duplicated test case.

Issue #670, opened 2018-06-27
[EN Number] Ambiguous recognition for "more than 30000 in 2009" 


guom08:
It can be recognized as "more than 30000" and "2009" is a year
And it can also be recognized as "more than (30000/2009)".

As the second form is not that common, the first one should be the default behavior (left/right order).

PR #673, opened 2018-06-27
Fix cases for "Show me records more than 3000 in 2009"

guom08:
#670


Specs/Number/English/NumberRangeModel.json
```diff
@@ -805,4 +805,60 @@
...
+        "value": "[0.4,)"
+      }
+    }
+  ]
+}
+,
+{
+  "Input": "How about more than 2 in 5",
+  "NotSupported": "javascript, python, java",
+  "Results": [
```
tellarin: Add a "Comment" attribute describing this is the decision.
"As this can be ambiguous, by design the interpretation is left-to-right and the range is extracted first."

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -143,9 +145,19 @@ private void GetMatchedStartAndLength(Match match, string type, string source, o
             else
             {
                 var numberStr = string.IsNullOrEmpty(numberStr1) ? numberStr2 : numberStr1;
-
+                
                 var extractNumList = ExtractNumberAndOrdinalFromStr(numberStr);
 
+                // If cases falls into "more/less than NUM" or "NUM or more/less"
+                if (extractNumList != null && (type == NumberRangeConstants.MORE || type == NumberRangeConstants.LESS))
```
tellarin: These two ifs should be combined into a method that determining if the special behaviour needs to be applied. If so, then it should tell ExtractNumberAndOrdinalFromStr which way to work.

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -210,6 +222,31 @@ private List<ExtractResult> ExtractNumberAndOrdinalFromStr(string numberStr)
             extractNumber = extractNumber.OrderByDescending(num => num.Length).ThenByDescending(num => num.Start).ToList();
             return extractNumber;
         }
+
+        // Fractions like "2 in 2009", "in" is ambiguous connector in some of the context
+        // This function is to process and remain only the part before ambiguous connector
+        private List<ExtractResult> ProcessAmbiguousFractions(List<ExtractResult> ers)
```
tellarin: This could should probably go inside ExtractNumberAndOrdinalFromStr

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -143,9 +145,19 @@ private void GetMatchedStartAndLength(Match match, string type, string source, o
             else
             {
                 var numberStr = string.IsNullOrEmpty(numberStr1) ? numberStr2 : numberStr1;
-
+                
                 var extractNumList = ExtractNumberAndOrdinalFromStr(numberStr);
```
tellarin: ExtractNumberAndOrdinalFromStr should not be in the NumberRangeExtractor as it doesn't handle duration concepts. Let's at least at a @TODO in the code to refactor and move it to a better place. Doesn't need to be in this PR.

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -143,9 +145,19 @@ private void GetMatchedStartAndLength(Match match, string type, string source, o
             else
             {
                 var numberStr = string.IsNullOrEmpty(numberStr1) ? numberStr2 : numberStr1;
-
+
```
tellarin: Nit: Please trim trailing spaces before pushing. :)

guom08: Fixed.

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
guom08: Added.

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
guom08: Fixed.

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
guom08: Fixed.

PR #683, opened 2018-06-29
[Java] Chinese Number models

pcostantini:
This PR includes the following Chinese Number models:
- NumberModel
- OrdinalModel
- PercentModel
- NumberRangeModel

(Tests NotSupported)
image


Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/parsers/NumberFormatUtility.java
```diff
@@ -27,7 +27,7 @@ private NumberFormatUtility() {
         supportedCultures.put(Culture.Portuguese, LongFormatType.DoubleNumDotComma);
         supportedCultures.put(Culture.French, LongFormatType.DoubleNumDotComma);
         supportedCultures.put(Culture.German, LongFormatType.DoubleNumDotComma);
-        supportedCultures.put(Culture.Chinese, LongFormatType.DoubleNumDotComma);
+        supportedCultures.put(Culture.Chinese, null);
```
tellarin: I’d prefer if we define a special value for None.

Issue #684, opened 2018-07-02
"show me sales for year 2018"


guom08:
"year 2018" should be recognized as daterange, from 2018-01-01 to 2019-01-01

Note that "show me sales for year 2018" is currently supported

PR #685, opened 2018-07-02
Fix cases like "show me sales in the year 2008"

guom08:
#684


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -4610,5 +4610,130 @@
...
+              "start": "2008-01-01",
+              "end": "2009-01-01"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Show sales in the year",
```
yiylin: image

yiylin: not sure if this is a right example

Issue #706, opened 2018-07-10
[EN DatetimeV2] Fix cases like "in the year of 2008"


guom08:
"in the year of 2008" should extract "2008" or "the year of 2008".
Currently we extract two entities "the year", "2008" separately.

PR #709, opened 2018-07-11
Fix cases like "in the year of 2008"

guom08:
#706


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -5285,5 +5285,30 @@
...
+            {
+              "timex": "2008",
+              "type": "daterange",
+              "start": "2008-01-01",
+              "end": "2009-01-01"
+            }
+          ]
+        }
+      }
+    ]
```
yiylin: if these test data are the same, maybe you can put them in another new common json file.

PR #722, opened 2018-07-14
Korean test specs added for Age and Date

zzxcv1314:
korean test spec added 
* NumberWithUnit - AgeModal
* DateTime - DateExtractor
* DateTime - DateParser


AniChikage:
Currently Korean is not supported in all platforms, so please add "Not supported":"javascript, python, dotnet, java" in your modified json files. 
Please run tests in all platforms to make sure there are no errors before making a new PR.

Issue #691, opened 2018-07-03
[C#][CN DateTime] Bug in Chinese merged extractor - CheckBlackList method


enzocanoo:
The [`CheckBlackList`](https://github.com/Microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/MergedDateTimeExtractor.cs#L64-L87) method in Chinese merged extractor seems to be used to modify the `dst` array but it is wrong implemented (`dst` passed by value)

```C#
private void CheckBlackList(List<ExtractResult> dst, string text)
{
	var ret = new List<ExtractResult>();
	var regex = new Regex(@"^\d{1,2}号", RegexOptions.IgnoreCase);
	foreach (var d in dst)
	{
...
	}
	dst = ret;
}
```

PR #723, opened 2018-07-16
Fix Bug in Chinese merged extractor, enable CheckBlackList 

STayinloves:
For GitHub issue #691


Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -1836,5 +1815,30 @@
         "End": 11
       }
     ]
+  },
+  {
+    "Input": "我今年18周岁了",
+    "Context": {
+      "ReferenceDateTime": "2018-07-16T16:00:00"
+    },
+    "NotSupported": "python",
```
STayinloves: This test case isn't supported in Python because of some other issues.

STayinloves: A small bug in Python's date period parser, should be fixed after 3d30f66

Issue #541, opened 2018-05-10
[CN DateTimeV2] Single year resolution is incorrect (starts at "*-02-01") in JS


STayinloves:
"2018"

``` json
{
        "start": 0,
        "end": 3,
        "resolution": {
                "values": [
                        {
                                "timex": "2018",
                                "type": "daterange",
                                "start": "2018-02-01",
                                "end": "2019-02-01"
                        }
                ]
        },
        "text": "2018",
        "typeName": "datetimeV2.daterange"
}
```

Sothan:
There is nothing to fix since the resolution of the single year in JS is correct now.
```
{
	"start": 0,
	"end": 3,
	"resolution": {
		"values": [
			{
				"timex": "2018",
				"type": "daterange",
				"start": "2018-01-01",
				"end": "2019-01-01"
			}
		]
	},
	"text": "2018",
	"typeName": "datetimeV2.daterange"
}
```

tellarin:
Thanks for the feedback, @Sothan. I'll close this issue then.

tellarin:
Thanks for the feedback, @Sothan. I'll close this issue then.

tellarin:
Thanks for the feedback, @Sothan. I'll close this issue then.

PR #728, opened 2018-07-18
Fix "0000" being treated as number zero in .NET

STayinloves:
- enable test cases failed because of inconsistency behavior previously(#489)
- enable test cases failed in #541 as it's supported now
- port previous changes in resources to py/ts


tellarin:
Closing for now as this solution needs to be further discussed.

PR #754, opened 2018-07-31
Basic support for Number in Korean + dotNet 

hongman:
Not all tests were passed, because some CJKParser code was not suitable at Korean.

I will update continuously Other Test Specs(NumberModelPercent, NumberRange….).


hongman:
Thank you very much for your reply. 
I modified the code, yaml for 'Number'

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -477,7 +477,15 @@ private double GetIntValue(string intStr)
             if (Config.NegativeNumberSignRegex.IsMatch(intStr))
             {
                 isNegative = true;
-                intStr = intStr.Substring(1);
+                //I'm so sorry thie code because i want to use 'NegativeNumberTermsRegex' but it is not in the CJKInterface
```
tellarin: You can add it to a config class like EnglishNumberParserConfiguration.cs and here you can get from the config class.
All expressions need to be defined in the source YAML files for that language.

tellarin: This if/else could also be its own method in BaseCJKNumberParser.cs for ease of reading.

Patterns/Korean/Korean-Numbers.yaml
```diff
@@ -40,15 +38,6 @@ ZeroToNineMap: !dictionary
     팔: 8
     구: 9
     빵: 0
-    하나: 1
```
tellarin: Are these really incorrect values? Or will they come back in future refinements of the code?

Specs/Number/Japanese/PercentModel.json
```diff
@@ -15,7 +15,7 @@
 ,
 {
   "Input": "この車は19パーセントの確率が遅れた",
-  "NotSupported": "javascript, python",
+  "NotSupported": "dotNet, javascript, python",
```
tellarin: Please also add "java" to the list of NotSupported platforms.
I recommend adding NotSupported for dotnet only, as this means "currently not supported, but implementation ongoing"
While the others should be NotSupportedByDesign, a separate tag that means "an implementation in these platforms doesn't exist yet and the effort for that has not started".

Specs/Number/Korean/NumberModel.json
```diff
@@ -1,6 +1,20 @@
...
+        "value": "-10000"
+      }
+    }
+  ]
+}
+,
+  {
   "Input": "192.",
-  "NotSupportedByDesign": "dotNet,javascript,python,java,",
+  "NotSupported": "javascript,python,java",
```
tellarin: If you keep the ...ByDesign, it will avoid warnings being thrown when these don't run.

Specs/Number/Korean/NumberModel.json
```diff
@@ -488,23 +502,9 @@
   ]
 }
 ,
-{
-  "Input": "-일만",
```
tellarin: Why was this case removed?
If it should be correct, but currently doesn't work in dotnet, just mark as NotSupported and it will be skipped with a warning.

AniChikage:
Besides, if this PR is designed to just support Number for dotNet, I suggest the title changed to "Basic support for Number in Korean + dotNet "

Specs/Number/Korean/NumberModel.json
hongman: This is because the expression '-일만' is not used in CJK (Chinese, Japanese, Korean).
It also does not exist in 'test Specs' in Chinese or Japanese (like "-一萬")
So I removed that test case.

Specs/Number/Japanese/PercentModel.json
hongman: oh I see.. I modified the json an hour ago.

Patterns/Korean/Korean-Numbers.yaml
hongman: In Korean, "하나둘~~~아홉" used to indicate the number of objects rather than "Number 1". 

To represent a simple number, we use "일이삼사오육칠팔구" rather than "하나둘~~~아홉"

So I removed that expression in yaml.

PR #762, opened 2018-08-01
Normalize existing testcases for Dutch language

iMicknl:
Changed ```NotSupported``` to ```NotSupportedByDesign``` for javascript, python and java. 

I kept the dotNet on NotSupported for now, should I also change them to NotSupportedByDesign since there is no implementation yet? Or keep them on NotSupported since implementation is planned.


tellarin:
Keep dotnet as NotSupported as it will be the 1st platform to come and we want to measure progress.
Thanks!

PR #763, opened 2018-08-01
Add Dutch DateTime specs

iMicknl:
Add beginning of the Dutch DateTime specs. If you could keep this open for a few days, then I will have some Dutch people proofreading it since I already spotted some small mistakes. Will push the fixes asap this week.

(created by the Dutch MicrosoftNL employees 👍)

### DIFF
--- a/Specs/DateTime/Dutch/DateExtractor.json
+++ b/Specs/DateTime/Dutch/DateExtractor.json
@@ -882,7 +882,7 @@
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
-        "Text": "zaterdag de 23ste ",
+        "Text": "zaterdag de 23ste",
         "Type": "date"
       }
     ],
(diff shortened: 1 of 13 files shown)


tellarin:
Sure. Thanks for adding the WIP tag.

iMicknl:
@tellarin should have full parity with English, only a few test cases removed which are not relevant for Dutch language. There are not to many specific Dutch scenarios added yet, but it is a start.

In the end I should have wrote a script which could convert the Parser files to Extractor files, since that is doing 'double work' as it is almost the same. I think those could be exactly the same testcases, just the difference that the ```Extractor``` doesn't contain results, just extraction.

tellarin:
@iMicknl, do you have an estimate of how much more refinement is needed before we can review/merge?

iMicknl:
@tellarin it is more work than I hoped. I will try to finalise them end of this week. At least they don't have syntax errors anymore, however there are still some missing translations / wrong translations.

iMicknl:
@tellarin if all tests pass, this one can be merged. I double checked most of the specs and quality looks good. Final revision can be done during implementation imo.

Issue #736, opened 2018-07-21
[JP Number] Sometimes, the percent model may incorrectly parse a number


levi106:
When I run the following unit test, some tests were failed. It is 100% reproducible in my environment, but I think that it depends on the environment.

```cmd
"c:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\CommonExtensions\Microsoft\TestWindow\vstest.console.exe" /Parallel  "C:\r\Recognizers-Text.me\.NET\Microsoft.Recognizers.Text.DataDrivenTests\bin\Release\Microsoft.Recognizers.Text.DataDrivenTests.dll" "C:\r\Recognizers-Text.me\.NET\Microsoft.Recognizers.Text.DataTypes.DataDrivenTests\bin\Release\Microsoft.Recognizers.Text.DataTypes.DataDrivenTests.dll"
```

```
失敗   PercentModel (データ行 1)
エラー メッセージ:
 Assert.AreEqual に失敗しました。<19%> が必要ですが、<9%> が指定されています。Input: "この車は19 パーセントの確率が遅れた"
スタック トレース:
   場所 Microsoft.Recognizers.Text.DataDrivenTests.TestBase.ValidateResults(IEnumerable`1 testResolutionKeys) 場所 C:\r\Recognizers-Text\.NET\Microsoft.Recognizers.Text.DataDrivenTests\TestBase.cs:行 330
   場所 Microsoft.Recognizers.Text.DataDrivenTests.TestBase.TestNumber() 場所 C:\r\Recognizers-Text\.NET\Microsoft.Recognizers.Text.DataDrivenTests\TestBase.cs:行 60
   場所 Microsoft.Recognizers.Text.Number.Tests.TestNumber_Japanese.PercentModel() 場所 C:\r\Recognizers-Text\.NET\Microsoft.Recognizers.Text.DataDrivenTests\Number\TestNumber_Japanese.cs:行 42

失敗   PercentModel (データ行 2)
エラー メッセージ:
 Assert.AreEqual に失敗しました。<50%> が必要ですが、<0%> が指定されています。Input: "彼女は50パ ーセントこのことはできません"
スタック トレース:
[...]

PR #766, opened 2018-08-03
Revise JP number regexes

STayinloves:
This PR address GitHub issue #736.


tellarin:
It seems the original code is actually relying on an undefined behaviour of hashtables. I'll close this PR as we analyze it further and modify the code appropriately.
Thanks @STayinloves for tracking this!

Issue #769, opened 2018-08-04
[ResourceGenerator] Output result is inconsistent between C# and JavaScript


levi106:
**C#**

```cs
#r "C:\r\Recognizers-Text.me\.NET\Microsoft.Recognizers.Definitions.Common\bin\Debug\net462\YamlDotNet.dll"
#r "C:\r\Recognizers-Text.me\.NET\Microsoft.Recognizers.Definitions.Common\bin\Debug\net462\Microsoft.Recognizers.Definitions.Common.dll"
using Microsoft.Recognizers.Definitions.Common;
var p = new YamlParser();
var s = @"CurrencyAmbiguousValues: !list
   types: [string]
   entries:
     - 円
     - 銭
     - \\";
var r = p.Deserialize(s);
var c = r["CurrencyAmbiguousValues"] as List;
foreach (var e in c.Entries) { WriteLine($"{e}"); }
円
銭
\\    <------ Two backslashes
```

**JavaScript**

```js
import * as YamlParser from "./yaml-parser";
import { GenerateCode, CodeWriter } from "./code-writers";
var s = `CurrencyAmbiguousValues: !list
   types: [string]
   entries:
     - 円
     - 銭
     - \\`;
let o = YamlParser.parse(s);
o
{ CurrencyAmbiguousValues: List { entries: [ '円', '銭', '\\' ] } }
let c = GenerateCode(o);
c
[ ArrayWriter {
    name: 'CurrencyAmbiguousValues',
    entries: [ '\'円\'', '\'銭\'', '\'\\\\\'' ],    <------ Four backslashes
    valueType: 'string' } ]
```

It seems that JSON.stringify function outputs one backslash as \\\\.

```js
JSON.stringify('\\')
'"\\\\"'
```

**Note:**

https://en.m.wikipedia.org/wiki/Yen_sign
> Japanese-language locales of Microsoft operating systems use the code page 932 character encoding, which is a variant of Shift JIS. Hence, 0x5C is displayed as a yen sign in Japanese-locale fonts on Windows.

PR #770, opened 2018-08-06
[JavaScript] Japanese Currency model

levi106:
- Add Currency recognizer for Japanese
  - A yen sign is not yet supported (#769)
- Does not remove leading and trailing whitespaces as in C#


JavaScript/packages/recognizers-number/src/number/extractors.ts
```diff
@@ -52,7 +52,7 @@ export abstract class BaseNumberExtractor implements IExtractor {
                 if (i + 1 === source.length || !matched[i + 1]) {
                     let start = last + 1;
                     let length = i - last;
-                    let substr = source.substring(start, start + length).trim();
```
tellarin: why remove the tim()?

tellarin: Makes sense. Thanks for the quick reply.

PR #780, opened 2018-08-10
Support culture codes for other English cultures/locales

STayinloves:
- Add tests
- modify date extractor and add EnableMdy DateTimeOptions
- Refactors in datetime extractor's configurator
- Fix some false test cases in ES and PT

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishDateExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishDateExtractorConfiguration.cs
@@ -1,8 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Collections.Immutable;
-using System.Linq;
-using Microsoft.Recognizers.Definitions;
+
 using Microsoft.Recognizers.Text.DateTime.English.Utilities;
 using Microsoft.Recognizers.Text.DateTime.Utilities;
 using Microsoft.Recognizers.Definitions.English;
@@ -22,7 +21,7 @@ namespace Microsoft.Recognizers.Text.DateTime.English
         public static readonly Regex MonthNumRegex =
             new Regex(DateTimeDefinitions.MonthNumRegex, RegexOptions.IgnoreCase | RegexOptions.Singleline);
 
-        public static readonly Regex YearRegex = 
+        public static readonly Regex YearRegex =
             new Regex(DateTimeDefinitions.YearRegex, RegexOptions.IgnoreCase | RegexOptions.Singleline);
 
         public static readonly Regex WeekDayRegex =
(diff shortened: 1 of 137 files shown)


.NET/Microsoft.Recognizers.Text/Parsers/IParser.cs
```diff
@@ -21,8 +21,6 @@ public ParseResult(ExtractResult er)
             Metadata = er.Metadata;
         }
 
-        public Metadata Metadata { get; set; } = null;
```
STayinloves: Remove this line because it overrides the base class but makes no differences.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishDateExtractorConfiguration.cs
```diff
@@ -178,8 +138,55 @@ public EnglishDateExtractorConfiguration(DateTimeOptions options = DateTimeOptio
             }
 
             ImplicitDateList = implicitDateList;
+
+            var enableMdy = (Options & DateTimeOptions.EnableMdy) != 0;
```
tellarin: Shouldn't this be based on the YAML config + the new config flag from culture initialization?

.NET/Microsoft.Recognizers.Text/Culture.cs
```diff
@@ -14,10 +15,12 @@ public sealed class Culture
         public const string Dutch = "nl-nl";
         public const string Korean = "ko-kr";
 
-        public string CultureName;
-        public string CultureCode;
+        public readonly string CultureName;
+        public readonly string CultureCode;
 
+        // The order of this array is important for GetCultureCodeWithFallback
```
tellarin: Fallback is always to English-US. Fallback should only be triggered if some culture is really not supported and the enableFallback flag is used.

.NET/Microsoft.Recognizers.Text/Culture.cs
```diff
@@ -29,15 +32,34 @@ public sealed class Culture
...
             this.CultureName = cultureName;
             this.CultureCode = cultureCode;
         }
 
-        public static string[] GetSupportedCultureCodes()
+        // If the input culture code isn't in the supported cultures list,
+        // the first culture with the same prefix as the input one will be returned.
+        // Otherwise, the original input culture code will be returned.
+        // e.g. "en-029"->"en-*" "vo-id"->"vo-id" "en-us"->"en-us"
+        public static string GetCultureCodeWithFallback(string cultureCode)
```
tellarin: I prefer not to change the method name. It's goal is still the same. 
We may even specify that fr-, es-, pt- are also *.

.NET/Microsoft.Recognizers.Text/ModelFactory.cs
```diff
@@ -35,12 +36,12 @@ public void InitializeModels(string targetCulture, TModelOptions options)
 
         private void InitializeModel(Type modelType, string culture, TModelOptions options)
         {
-            this.TryGetModel(modelType, culture, options, out IModel model);
+            this.TryGetModel(modelType, culture, options, out _);
```
tellarin: For readability, I recommend we keep the previous code.

.NET/Microsoft.Recognizers.Text/ModelFactory.cs
```diff
@@ -35,12 +36,12 @@ public void InitializeModels(string targetCulture, TModelOptions options)
...
         private void InitializeModel(Type modelType, string culture, TModelOptions options)
         {
-            this.TryGetModel(modelType, culture, options, out IModel model);
+            this.TryGetModel(modelType, culture, options, out _);
         }
 
         private bool TryGetModel<T>(string culture, TModelOptions options, out T model) where T : IModel
         {
-            var result = this.TryGetModel(typeof(T), culture, options, out IModel outModel);
+            var result = this.TryGetModel(typeof(T), culture, options, out var outModel);
```
tellarin: Same.

PR #783, opened 2018-08-11
[dotNet+JavaScript]fix wrong output values like "五百五" in Japanese Number model

AniChikage:
"五百五" should output "505", not "550"


hongman:
It seems to be suitable for Korean. I will update soon..

PR #790, opened 2018-08-15
[* Sequence] Extend the .NET model to score phone number recognition results

godlovesun:
- Override the basePhoneNumberExtractor, BaseSequenceParser and PhoneNumerModel to support scoring

- Add Score attribute to the ExtractResult class

- Reformat the yaml to provide information about different patterns of phone numbers and scoring details

### DIFF
--- a/Specs/Sequence/English/PhoneNumberModel.json
+++ b/Specs/Sequence/English/PhoneNumberModel.json
@@ -3514,6 +3514,56 @@
     ]
   },
   {
+    "Input": "Do you know bazinga?",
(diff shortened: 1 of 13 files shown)


Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -3296,56 +3513,6 @@
       }
     ]
   },
-  {
-    "Input": "Do you know bazinga?",
```
tellarin: All these negative tests should not be removed. Can you please add them back?

PR #793, opened 2018-08-16
Added German-Choice

cloudguy-pro:
I added the German expressions for "Choice" in a separate .yaml file...


tellarin:
Thanks, @cloudguy-pro. Do you think you can add the related .Net or JS code? It should be very similar to the structure for other languages.

cloudguy-pro:
Yeah, I'll do that as well!

PR #804, opened 2018-08-21
 Support business days/hours in English .NET

STayinloves:
For cases like "next 4 business days" and "business hours".


.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishTimePeriodParserConfiguration.cs
```diff
@@ -90,6 +90,12 @@ public bool GetMatchedTimexRange(string text, out string timex, out int beginHou
                 endHour = 23;
                 endMin = 59;
             }
+            else if (trimedText.Contains("business") && trimedText.Contains("hour"))
```
tellarin: Ideally all such strings should go into YAML files.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1260,27 +1242,55 @@ private DateTimeResolutionResult ParseDuration(string text, DateObject reference
...
                 ret.Success = true;
-
-                return ret;
             }
 
             return ret;
         }
 
-        private void GetModAndDate(ref DateObject beginDate, ref DateObject endDate, DateObject referenceDate, string timex, bool future, out string mod)
+        private static void GetModAndDate(out DateObject beginDate, ref DateObject endDate, DateObject referenceDate,
```
tellarin: Ideally this method signature should be modified. It looks like it's doing too many things and returning too many side-effects.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDurationParser.cs
```diff
@@ -208,33 +207,34 @@ private DateTimeResolutionResult ParseNumberCombinedUnit(string text)
...
-                    ret.Timex = "P" + (IsLessThanDay(unitStr) ? "T" : "") + numStr + unitStr[0];
+                    ret.Timex = TimexUtility.GenerateDurationTimex(numVal, unitStr, IsLessThanDay(unitStr));
                     ret.FutureValue = ret.PastValue = double.Parse(numStr) * this.config.UnitValueMap[srcUnit];
                     ret.Success = true;
-
-                    return ret;
+                }
+                else if (match.Groups["business"].Success)
+                {
+                    ret.Timex = TimexUtility.GenerateDurationTimex(numVal, "BD", false);
```
tellarin: We should not pass BD as a string into it. Parameters should define the input needed to generate the appropriate regex. The actual substring to use is an internal implementation detail of the Gen method.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDurationParser.cs
```diff
@@ -208,33 +207,34 @@ private DateTimeResolutionResult ParseNumberCombinedUnit(string text)
...
+                    ret.FutureValue = ret.PastValue = numVal * this.config.UnitValueMap[srcUnit.Split()[1]];
```
tellarin: Is this split always safe?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -420,6 +420,13 @@ public List<DateTimeParseResult> DateTimeResolutionForSplit(DateTimeParseResult
 
             var islunar = val.IsLunar;
             var mod = val.Mod;
+            List<string> list = null;
+
+            // Resolve dates list for date periods
+            if (slot.Type.Equals(Constants.SYS_DATETIME_DATEPERIOD) && val.List != null)
```
tellarin: We only need to add these for BD, no?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -514,14 +521,15 @@ public List<DateTimeParseResult> DateTimeResolutionForSplit(DateTimeParseResult
 
             foreach (var p in res)
             {
-                if (p.Value is Dictionary<string, string>)
+                if (p.Value is Dictionary<string, string> dictionary)
                 {
-                    var value = new Dictionary<string, string>();
+                    var value = new Dictionary<string, object>();
```
tellarin: Using object is probably not the best design. Can we use a concrete class in here?

tellarin: and elsewhere...

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/DurationParsingUtil.cs
```diff
@@ -97,13 +98,57 @@ private static DateObject GetShiftResult(IImmutableDictionary<string, double> ti
...
             var ret = new Dictionary<string, double>();
-            // resolve duration timex, such as P21DT2H(21 days 2 hours)
+
+            // Resolve duration timex, such as P21DT2H(21 days 2 hours)
             var durationStr = timexStr.Replace("P", "");
             var numberStart = 0;
             var isTime = false;
+
+            // Resolve business days
+            if (durationStr.EndsWith("BD"))
```
tellarin: It's better to define a constant for this term. It will make it much easier to change in the future.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
STayinloves: Yes, the list only added to "BD" cases, for now, this line is able to do this check.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
STayinloves: I agree that the `object` isn't the optimal solution here, but since changing this will lead to a great effort to a big refactor, I suggest to do this as a further improvement.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
STayinloves: This method has a lot of parameter in the signature, but I think it's doing a simple work about getting the date range and modifier. I have tried to split this method but causing lots of side effects, so I prefer to not change it here.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDurationParser.cs
STayinloves: It's safe I think.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
tellarin: OK. Let's add a TODO and prioritize this refactoring for another PR.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
tellarin: Add this to the same work item as that signature refactor, ok?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDurationParser.cs
```diff
@@ -267,9 +269,9 @@ private DateTimeResolutionResult ParseInexactNumberUnit(string text)
                     ret.FutureValue = ret.PastValue = double.Parse(numStr) * this.config.UnitValueMap[srcUnit];
                     ret.Success = true;
                 }
-                else if (match.Groups["business"].Success)
+                else if (match.Groups[Constants.BusinessDayGroupName].Success)
                 {
-                    ret.Timex = TimexUtility.GenerateDurationTimex(numVal, "BD", false);
+                    ret.Timex = TimexUtility.GenerateDurationTimex(numVal, Constants.TimexBusinessDay, false);
```
tellarin: This is still not ideal. The method parameters should be the info needed to generate a timex. The actual string value is an implementation detail internal to the method. The string itself should not be received as a parameter. Let's add this change to the refactor item too, ok?

Issue #796, opened 2018-08-16
[CN|DE|FR Choice] Boolean missing support for German, Chinese and French


johnataylor:
The ChoiceRecognizer RecognizeBoolean needs support for German, Chinese and French.
The priority is implementation in .NET and JavaScript.

tellarin:
@matthewshim-ms, could you help with French?

tellarin:
@johnataylor PR #812 merged in French support and I've added the missing German pieces. It would benefit from more tests, but I'll close this issue as done for now. :)

PR #812, opened 2018-08-28
French choice recognizer

matthewshim-ms:
@tellarin initial pass for now. You are correct it's quite simple so this shouldn't take very long.


.NET/French-Choice.yaml
```diff
@@ -0,0 +1,9 @@
+---
+LangMarker: Eng
```
tellarin: This file is in the wrong place. It should be in Patterns\French

tellarin:
This PR will cover the French part of #796

tellarin:
@matthewshim-ms, I went ahead and fixed the French .NET code. All tests pass now. Please take a look.
I've also added German support and all .NET specs also pass. 
But locally here JS is breaking. I'll debug later.

tellarin:
All pass now.

PR #817, opened 2018-08-29
Use lerna 2.11.0 as default

enzocanoo:
Force `lerna@2.11.0` due to an issue with the license of future versions.


iMicknl:
FYI https://github.com/lerna/lerna/pull/1633

PR #833, opened 2018-09-04
Add Japanese/DateTimeParser.json

c-nova:
Add a DateTimeParser.json in Japanese.

### DIFF
--- a/Specs/DateTime/Japanese/DateTimeParser.json
+++ b/Specs/DateTime/Japanese/DateTimeParser.json
@@ -1,0 +1,1323 @@
+[
+  {
+    "Input": "今から帰ります",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupportedByDesign": "javascript,python,java,",
+    "Results": [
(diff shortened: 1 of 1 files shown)


AniChikage:
Hi, masayoshi-san
Since this PR is just translation, I think you may make the "NotSupportedByDesign" like this: "NotSupportedByDesign" : "dotnet, javascript, python, java", it should be fine.

Also, before making a PR, you could run "build.cmd" to ensure no errors. :)

tellarin:
Hi Masayoshi-san (@c-nova), 

I went ahead and fixed the NotSupported tags as Zhishuai (@AniChikage) mentioned.
New specs should always be marked as **NotSupportedByDesign** in all platforms initially, then, when the first dotNet code comes in, marked as **NotSupported** in dotnet only (keep ByDesign in other plats), and slowly remove the tags once test cases pass, ok?

I also noticed that the spec file was encoded in **Shift_JIS**, but we need to ensure encoding is **UTF-8**. I've transcoded it.

c-nova:
Thank you!

Issue #800, opened 2018-08-17
[EN Dimension] 500 MB / mB / mb isn't recognized correctly


iMicknl:
While working on Dutch NumberWithUnits and creating more tests, I found out that Megabyte / Megabit isn't recognised correctly in all cultures. (and it doesn't have tests in English specs).

```json
{
  "Text": "500 mB",
  "Start": 0,
  "End": 5,
  "TypeName": "dimension",
  "Resolution": {
    "value": null
  }
}
```

```json
{
  "Text": "500 kB",
  "Start": 0,
  "End": 5,
  "TypeName": "dimension",
  "Resolution": {
    "unit": "Kilobit",
    "value": "500"
  }
}
```

iMicknl:
@tellarin @STayinloves could this maybe be reopened? I just checked it and Megabit is indeed working now, however Megabyte isn't recognized.

Input: 500 mB, output should be 500 Megabyte according to the defined Pattern


```json
{
  "Text": "500 mB",
  "Start": 0,
  "End": 5,
  "TypeName": "dimension",
  "Resolution": {
    "unit": "Megabit",
    "value": "500"
  }
}
```

``` Megabyte: -megabyte|-megabytes|megabyte|mB|megabytes|mega byte|mega bytes|MByte```

Also, the symbol for [Megabyte](https://en.wikipedia.org/wiki/Megabyte) is MB and not mB. Should I do a PR to change this or is this something that could be fixed while fixing this issue?

tellarin:
@iMicknl thanks for the catch. Should be covered now.

PR #836, opened 2018-09-05
[NL NumberWithUnit] Fix Kilobyte definition and enable more tests 

iMicknl:
- Bugfix: Change KB to kB
- Activate disabled tests

Still one test failing, which is logged in issue #800


iMicknl:
I agree. I added KB back and added a testcase for it. However, this is something that should be fixed in the English patterns and testcases also. The variants used there are incorrect at the moment.

See issue #800

PR #837, opened 2018-09-05
[NL Numbers] Support Dutch OrdinalModel in .NET

iMicknl:
- Add support for Dutch OrdinalModel in .NET
- Two failing testcases marked as not supported, need to have a discussion about the specific testcases

`three trillionth` is supported in English OrdinalModel, but `a trillionth` should return an empty result. Could someone elaborate this case? I am not sure why this is and maybe this should be different in Dutch due to language differences.


tellarin:
@iMicknl ```a trillionth``` in English is treated as a fraction (1/1000...) if I recall correctly, so it wouldn't be treated as Ordinal.
No idea how it should be in Dutch. :)

iMicknl:
Ok, I get it. In Dutch this is quite a special case. `een triljoenste` can be `one trillionth` or `an trillionth`, however I think we only use the first in Dutch which means it should be treated as fraction.

Is it ok to keep the two cases NotSupported for now, while I think about a fix?
@tellarin `CI of Microsoft.Recognizers-BotSys` is failing, however I don't have access to that CI step.

tellarin:
@iMicknl yes, it's fine to keep as unsupported by now. 
The CI failure is a VSO bug, we've reported to them and disabled that build for PRs. The AppVeyor build is enough for now.
I'll trigger a re-build.

Issue #656, opened 2018-06-25
[CN DatetimeV2] Fix cases like "2007年与2009年(之)间"


guom08:
The output should be a date range

Currently the output is two separate date range "2007年" and "2009年"

PR #838, opened 2018-09-06
Fix cases like '2007年与2009年(之)间'(#656)

Sothan:
Github #656


Sothan:
Sorry, there are some issues in the first submission, JS build went wrong and I've fixed that.

PR #839, opened 2018-09-06
Support cases like "year greater than 2012", "year 2012 or greater"

guom08:


.NET/Samples/SimpleConsole/Program.cs
```diff
@@ -68,7 +68,7 @@ private static IEnumerable<ModelResult> ParseAll(string query, string culture)
 
                 // Number Range recognizer will find any cardinal or ordinal number range
                 // E.g. "between 2 and 5" will return "(2,5)"
-                NumberRecognizer.RecognizeNumberRange(query, culture),
+                NumberRecognizer.RecognizeNumberRange(query, culture, NumberOptions.ExperimentalMode),
```
tellarin: Please revert this change. Program.cs is a demo of standard behaviour.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -6600,5 +6600,155 @@
...
+              "type": "daterange",
+              "start": "2016-01-01"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "You can only leave on 1/1/2016 and after",
```
tellarin: I'd prefer if these "and after" cases can only work at the end of a sentence.
Can you add a case that shows one sentence with two uses of dates. For example:
"I can only leave on 1/1/2016 and after my work item is done"
and
"I can only leave on 1/1/2016 and after 6PM"

guom08: Thanks for catching these failed cases.
Fixed.

PR #842, opened 2018-09-06
[NL Numbers] Support Dutch PercentModel in .NET

iMicknl:
- Added support for Dutch PercentModel in .NET
- Marked one case as NotSupported, since 'de' (the) is not a supported article yet

### DIFF
--- a/Specs/Number/Dutch/PercentModel.json
+++ b/Specs/Number/Dutch/PercentModel.json
@@ -1,7 +1,6 @@
 [
   {
     "Input": "100%",
-    "NotSupported": "dotNet",
     "Results": [
       {
         "Text": "100%",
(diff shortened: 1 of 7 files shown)


Specs/Number/Dutch/PercentModel.json
```diff
@@ -84,11 +78,10 @@
     "NotSupportedByDesign": "javascript,python,java"
   },
   {
-    "Input": "percentage van 30 procent",
```
tellarin: Isn't this a possible form? If so, we should tag all, right?

iMicknl: I am not 100% sure. I translated this one initially, but changed it since `percentage van 30 procent` is double. It should be `percentage van 30` or `30 procent` in my opinion. I couldn't find any reference for this edge case. Should I add this one as NotSupported for now?

iMicknl: I added it for now on our internal planner board, where we can review it with more native Dutch people.

PR #846, opened 2018-09-07
Refinements in Experimental mode (may be removed)

guom08:
Filter out special cases related to Fraction extraction


.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -232,14 +233,28 @@ private List<ExtractResult> ExtractNumberAndOrdinalFromStr(string numberStr, boo
                 ret = ret.OrderByDescending(num => num.Length).ThenByDescending(num => num.Start).ToList();
             }
 
-            if (ret != null && isAmbiguousRangeOrFraction)
+            var removeFractionWithInConnector = false;
+
+            if (Options == NumberOptions.ExperimentalMode)
```
tellarin: This is not the correct way to check for enum flags.

tellarin: Can you also move the logic that sets removeFractionWithInConnector to its own method? This is likely to change in the future and would make the code easier to read.

Issue #544, opened 2018-05-10
[CN Percent] Incorrect recognized entity in utterance "百分之七十百"


sinbai:
The utterance "百分之七十" in  sentence ”股票十天收盘价的百分之七十百分位大于1“ is analyzed "7000%" in JSON data.

![image](https://user-images.githubusercontent.com/39109137/39862069-3c079d48-5475-11e8-98be-88e70e828c00.png)

PR #873, opened 2018-09-20
fix wrong parse results like '百分十七十百', '五十千'

Sothan:
- Github Issue #544 
- Fix wrong parse results like “十百”, “十千”, "百千" etc.


Sothan:
It seems that there are some inner issues at regex of JavaScript, because the same regex in .NET, Python and JS will extract different entities and JS will extract wrong entities.

tellarin:
@Sothan, this type of thing is expected. Please add notes on the PR or in the spec with the reason for the difference.

Issue #883, opened 2018-09-21
[EN.DatetimeV2] Support cases like "Monday at 2.30", "Let's meet before 2.30pm"


guom08:
**Describe the bug**
Currently cases like "Monday at 2.30", "Let's meet before 2.30pm" can't be recognized as Time or DateTime entity

**Expected behavior**
"at 2.30" should be recognized as Time entity
"2.30pm" should be recognized as Time entity
"2.30" should not be recognized as Time entity

PR #884, opened 2018-09-21
Support cases like "Monday at 2.30", "Let's meet before 2.30pm"

guom08:
#883


tellarin:
.NET version merged in. Javascript pending, @guom08.

PR #893, opened 2018-09-28
Add 元 as a supported symbol for reminbi in the English yaml

Sothan:


iMicknl:
@Sothan @tellarin shouldn't this be something that should be merged to all languages which support the CurrencyModel? Since currency symbols are mostly universal and not inherited from Base-Currency.

PR #895, opened 2018-09-29
Fix wrong parse result for '2:00 PM', '00 PM' is not dimension

Sothan:
- Github Issue #605


Patterns/English/English-Numbers.yaml
```diff
@@ -27,7 +27,7 @@ PlaceHolderPureNumber: !simpleRegex
 PlaceHolderDefault: !simpleRegex
   def: \D|\b
 NumbersWithPlaceHolder: !paramsRegex
-  def: (((?<!\d+\s*)-\s*)|(?<=\b))\d+(?!(\.\d+[a-zA-Z]))(?={placeholder})
+  def: (?<!:\s*)(((?<!\d+\s*)-\s*)|(?<=\b))\d+(?!(\.\d+[a-zA-Z]))(?={placeholder})
```
tellarin: Shouldn't this change be applied to the units package? "00" is a number.
Also, the fix should affect all languages, not only English.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -259,6 +259,28 @@ public List<ExtractResult> Extract(string source)
                         /* Relative position will be used in Parser */
                         number.Start = start - er.Start;
                         er.Data = number;
+
+                        //Special treatment, handle cases like '2:00 pm', '00 pm' is not dimension
```
tellarin: This could a be a method to make the code cleaner and easy to modify. Also, the solution could be more generic to allow handling other potentially ambiguous terms.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/INumberWithUnitExtractorConfiguration.cs
```diff
@@ -26,5 +26,7 @@ public interface INumberWithUnitExtractorConfiguration
         string ConnectorToken { get; }
 
         Regex CompoundUnitConnectorRegex { get; }
+
+        Regex SpecialTimeRegex { get; }
```
tellarin: The name here should be more descriptive that this is to filter potentially ambiguous cases. Not some special case of time extraction.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -327,6 +349,27 @@ public void ExtractSeparateUnits(string source, List<ExtractResult> numDependRes
                                 matchResult[j] = true;
                             }
 
+                            //Special treatment, handle cases like '2:00 pm', both '00 pm' and 'pm' are not dimension
+                            var isDimensionFallsInTime = false;
```
tellarin: Why is this repeated here? If really needed, it should be a call to the same method.

PR #896, opened 2018-10-01
Remove German word from dutch choices (nein is not a valid dutch word)

NaamloosDT:
I noticed the German word "nein" being part of the dutch choices for "no". Differently from "yes" and "no", "nein" is never really used.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions/Dutch/ChoiceDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions/Dutch/ChoiceDefinitions.cs
@@ -20,6 +20,6 @@ namespace Microsoft.Recognizers.Definitions.Dutch
 		public const string LangMarker = "Nl";
 		public const string TokenizerRegex = @"[^\w\d]";
 		public const string TrueRegex = @"\b(ja|jawel|jazeker|natuurlijk|prima|jep|yes|yep|y|ok|oke)\b|(\uD83D\uDC4D|\uD83D\uDC4C)";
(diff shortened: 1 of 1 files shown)


iMicknl:
Thanks for your contribution! 

In my opinion it wouldn't hurt to keep `nein` in the Recognizer for Dutch, since it won't impact the performance that much and won't cause false positives. However I can understand why you would remove it, since it is not native Dutch and it is way less used than `yes` and `no`.

I added `nein` to the Dutch BooleanRecognizer a while ago, but I agree that this is rarely used in the Netherlands. I had a discussion with some other Dutchies about it and they thought this could still be used sometimes in spoken or informal language. 
What do you think about it?

NaamloosDT:
Personally I only encounter this in a jokingly manner. Though I see the point in keeping it. There's possible scenarios where people jokingly interact with software using this lib (e.g. I'm planning on using it in a Discord bot, and Discord isn't exactly the most formal way to interact with others)

Thank you for your response.

iMicknl:
I agree that this wouldn't be used in formal writings. However in chatbots where this package is commonly used, it could be useful. By the way, I am not a codeowner of this project, just a contributor. @tellarin, what would your opinion be on this?

If you would like to contribute to extending Dutch language support, let me know. There are still quite some cases not supported where we definitely could use help. (also check [#hacktoberfest](https://open.microsoft.com/2018/09/30/join-hacktoberfest-2018-celebration-microsoft/) 😉).

tellarin:
I'm also not the "owner" here, just one of the maintainers. 😉
I tend to agree with @iMicknl. If it's a term that could appear in common usages of the recognizers and there is no side effect of keeping it, I suggest we keep it. 
I'll close this issue for now as we seem to all agree.

PR #898, opened 2018-10-01
[NL] Optimize BooleanModel & NumberWithUnitsModel

iMicknl:
- Add support for `vanzelfsprekend|zeker|akkoord` for BooleanRecognizer
- Add support for `nooit` for BooleanRecognizer
- Add support for `x jarige` and `x-jarige` for NumberWithUnitsRecognizer (frequently seen in Dutch news articles)
- Add support for more dimension cases like `vierkante km`
- Add support for `millimicron`
- Add support for different cases in size definitions (mb/MB/Mb)
- Add support for `imperial pint`
- Bugfix rename `PM` to `PB`
- Add support for 元 as symbol for Chinese yuan (#893)

Included tests for most missing cases


iMicknl:
The only case that would be open for discussion is how we should handle `mb`, `gb`, `tb` and `pb`. In English this is mapped to `Megabit` (and equivalents), however I don't agree with this.

In Dutch most uses of full lowercase variants (like `mb`) would be mapped to Megabyte instead of Megabit. `mb/s` and `mbps` would be mapped to `Megabit per second` which is not supported at all yet. Megabit is mostly used when measuring internetspeed, which includes per second.

Please review this change thoroughly, since this could be breaking between languages. Open for discussion about this.

iMicknl:
@tellarin how can I add the start / end indexes easily to my tests? It looks like the Python test suite is failing on it, but I wasn’t planning on adding it manually. (I don’t understand why those are running at all?).

guom08:
> @tellarin how can I add the start / end indexes easily to my tests? It looks like the Python test suite is failing on it, but I wasn’t planning on adding it manually. (I don’t understand why those are running at all?).

Hi. If the format is the same as the C# model output format, you may paste the content from the C# console window. If the format is different, my suggestion is to unify the format the same as the C# output so next time you want to add more specs, it would be much easier:)

iMicknl:
@guom08 do you have any idea why the Python tests are failing after my changes? My testcases have `python` marked as `NotSupportedByDesign` and `AgeModel` has valid Start / End keys.

`C:\projects\recognizers-text\Python\tests\test_runner_number_with_unit.py:22: KeyError: 'Start'
26536`

iMicknl:
@guam08 I found the culprit maybe. The Dutch `DimensionModel` and `CurrencyModel` don't have `Start` and `End` keys yet. Apparently this is introduced (and enforced) after the Dutch NumberWithUnit model has been merged in.

What would be an easy fix to get this working? At the moment Python and Javascript are marked as 'NotSupportedByDesign', so the testrunner shouldn't even be running on it. I will have a look the coming days if I can setup the Python test suite locally to check if the error is because of one of the Dutch testcases.

iMicknl:
@guom08 @tellarin any idea how I could fix this? (without adding Start/End tags manually to 100+ testcases). I would like to get it merged in asap, before I have to solve many conflicts.

However the [`mb`](https://github.com/Microsoft/Recognizers-Text/pull/898#issuecomment-425968681) discussion is also something which needs to be discussed.

iMicknl:
@guom08 @tellarin any idea how I could fix this? (without adding Start/End tags manually to 100+ testcases). I would like to get it merged in asap, before I have to solve many conflicts.

However the [`mb`](https://github.com/Microsoft/Recognizers-Text/pull/898#issuecomment-425968681) discussion is also something which needs to be discussed.

iMicknl:
@guom08 @tellarin any idea how I could fix this? (without adding Start/End tags manually to 100+ testcases). I would like to get it merged in asap, before I have to solve many conflicts.

However the [`mb`](https://github.com/Microsoft/Recognizers-Text/pull/898#issuecomment-425968681) discussion is also something which needs to be discussed.

tellarin:
I'll finish reviewing and merge soon. But I suggest we leave the "mb" for another PR.

tellarin:
I'll finish reviewing and merge soon. But I suggest we leave the "mb" for another PR.

iMicknl:
@tellarin yes, in the end it was just my mistake. I forgot to mark one testcase as `NotSupportedByDesign`, just found out when running the testcases locally in a more verbose option.

PR #900, opened 2018-10-08
[CN DateTime] Increase the recognition for "後|之後"

felaray:


felaray:
ok, I'll try , thanks!

PR #901, opened 2018-10-08
[CN DateTime] Increase the recognition for "後|之後"

felaray:


tellarin:
@felaray, PRs that change the YAML resource files should also contain the generated code files in the different platforms. Please add the files to PR #900.
Thanks.

PR #902, opened 2018-10-08
 [CN DateTime] Extend the recognition for traditional Chinese

felaray:
Revised direction 
1. 后=>后|後
2. 之后=>之后|之後
3. 以后=>以后|以後


tellarin:
@felaray Thanks! Next time, no need to close the PR (line in #900). You can just push the fixes to the same PR number.

PR #905, opened 2018-10-10
Improve recognition of Url and Email

guom08:


.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BaseURLExtractor.cs
```diff
@@ -38,7 +41,20 @@ public BaseURLExtractor()
         public override List<ExtractResult> Extract(string text)
         {
             var ret = base.Extract(text);
-            var urlMatches = UrlRegex.Matches(text);
+            ret.AddRange(ExtractUrl(text, UrlRegex));
```
tellarin: I think this flow is a little unusual. Isn't it better to check all regexes as in base and then filter?

Issue #746, opened 2018-07-29
[CN DatetimeV2] No extract result for "上午"


jiongjiongli:
>>> results = Recognizers.recognize_datetime('上午', Culture.Chinese)
>>> for result in results:
...     print(json.dumps(result, default=lambda o: o.__dict__, indent='\t', ensure_ascii=False))
... 
>>> 

Actual Result is empty

felaray:
I think you should join the date. 
For example: 今日上午

tellarin:
For consistency with English, @Sothan has added this across platforms. Thanks, folks.

PR #907, opened 2018-10-11
Support extract cases like '上午', '下午' in Chinese DateTime

Sothan:
Github issue #746

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
             if (extra != null)
             {
                 // Handle special case like '上午', '下午'
-                var parseResult = ParseNight(er.Text, referenceTime);
+                var parseResult = ParseTimeOfDay(er.Text, referenceTime);
 
                 if (!parseResult.Success)
                 {
@@ -83,7 +83,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
             return null;
         }
 
-        private DateTimeResolutionResult ParseNight(string text, DateObject referenceTime)
+        private DateTimeResolutionResult ParseTimeOfDay(string text, DateObject referenceTime)
         {
             int day = referenceTime.Day,
                 month = referenceTime.Month,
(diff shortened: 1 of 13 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
```diff
@@ -76,6 +83,76 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
             return null;
         }
 
+        private DateTimeResolutionResult ParseNight(string text, DateObject referenceTime)
```
tellarin: This should be renamed to ParseTimeOfDay, no? Maybe also in other languages.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
```diff
@@ -76,6 +83,76 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
...
+        }
+        
+        private bool GetMatchedTimexRange(string text, out string timex, out int beginHour, out int endHour, out int endMin)
+        {
+            var trimmedText = text.Trim();
+            beginHour = 0;
+            endHour = 0;
+            endMin = 0;
+            
+            if (trimmedText.EndsWith("上午"))
```
tellarin: Ideally all these hardcoded strings should be defined in the YAML file. Can you move them there?

Sothan: Sorry, I don't know how to achieve this since there are the similar hardcoded strings in English.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
Sothan: Ok, updated.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
tellarin: You can take a look at the English YAML file. DateTokenPrefix shows an example of declaring a string there.
Ideally the string hardcoded in other languages should also move. This could be a separate PR.
Thanks.

PR #911, opened 2018-10-16
Fix year wrongly recognized inside a GUID

Sothan:
- Fix wrong extract single year which is surrounded by '-' at both sides.
- For example, for case '-1687-', '1687' shouldn't be extracted as a single year.


Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -13,6 +13,8 @@ MonthNumRegex: !simpleRegex
 TwoNumYear: '50'
 YearNumRegex: !simpleRegex
   def: (?<year>((1[5-9]|20)\d{2})|2100)
+IllegalYearRegex: !simpleRegex
```
tellarin: If you want to declare a new regex for this, it should reference IllegalYearRegex, not re-declare it.
And to avoid it being declared everywhere, ideally it should be somewhere like Base-DateTime.yaml

Patterns/English/English-DateTime.yaml
```diff
@@ -34,6 +34,8 @@ WrittenNumRegex: !simpleRegex
 FullTextYearRegex: !nestedRegex
   def: \b((?<firsttwoyearnum>{CenturyRegex})\s+(?<lasttwoyearnum>((zero|twenty|thirty|forty|fourty|fifty|sixty|seventy|eighty|ninety)\s+{WrittenNumRegex})|{WrittenNumRegex}))\b|\b(?<firsttwoyearnum>{CenturyRegex})\b
   references: [ CenturyRegex, WrittenNumRegex ]
+IllegalYearRegex: !simpleRegex
```
tellarin: Same and so on.

Python/libraries/recognizers-date-time/recognizers_date_time/date_time/base_dateperiod.py
```diff
@@ -147,9 +152,14 @@ def match_simple_cases(self, source: str) -> List[ExtractResult]:
                     if not year_str:
                         year = self.__get_year_from_text(match_year)
 
-                        if not (year >= 1500 and year <= 2000):
+                        if not (year >= 1500 and year <= 2100):
```
tellarin: Ideally these boundaries should reference values defined in the YAML resource files too.

tellarin: Exactly to avoid cases like this where some value was not updated when it changed in another place.

PR #918, opened 2018-10-18
[Mirror of PR#895] Add python support to fix dimension wrongly recognized inside time

Sothan:
Mirror of PR #895


Specs/NumberWithUnit/German/DimensionModel.json
```diff
@@ -380,7 +380,7 @@
   {
     "Input": "2:00 pm",
     "NotSupportedByDesign": "python",
```
tellarin: Shouldn't this be enabled now?

PR #932, opened 2018-10-23
Refactor hardcoded strings and renames

Sothan:
- Refactor hardcoded strings in parse TimeOfDay
- Rename 'isDimensionFallsInTime' to 'isNotUnit'


.NET/Microsoft.Recognizers.Text.DateTime/Utilities/DateTimeResolutionResult.cs
```diff
@@ -42,4 +42,13 @@ public class TimeZoneResolutionResult
 
         public string TimeZoneText { get; set; }
     }
+
+    public class TimeOfDayResolutionResult
```
tellarin: Timex should be part of it.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -207,5 +207,42 @@ public static DateObject OffsetDateObject(DateObject date, int offset, DatePerio
 
             return ret;
         }
+
+        public static TimeOfDayResolutionResult ParseTimeOfDay(string timex)
```
tellarin: Rename the parameter to tod. It can remain a string, or be an enum.
timex should be a property of the response. It's fine if it's the same string value as passed in. But it may not always be the case.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/TimePeriodParserChs.cs
```diff
@@ -112,44 +112,40 @@ private bool GetMatchedTimexRange(string text, out string timex, out int beginHo
             beginHour = 0;
             endHour = 0;
             endMin = 0;
-            
-            if (trimmedText.EndsWith(Constants.CN_MORNING))
+
+            var timeOfDay = "";
+            if (trimmedText.EndsWith(DateTimeDefinitions.Morning))
```
tellarin: In this code to determine tod, timeOfDay should receive values representing the tod concept (maybe even directly like TMO), not a string in a given language.

PR #934, opened 2018-10-24
Fix "cent" wrongly extracted as currency from "pour cent" 

tellarin:
Issue #206. 
Moved existing filter code.


Patterns/Portuguese/Portuguese-DateTime.yaml
```diff
@@ -819,4 +819,8 @@ SpecialDecadeCases: !dictionary
     '': 0
 DefaultLanguageFallback: DMY
 DurationDateRestrictions: []
+AmbiguityFiltersDict: !dictionary
+  types: [ string, string ]
+  entries:
+    'null': 'null'
```
tellarin: Currently we don't have support for empty dictionaries.

PR #947, opened 2018-11-01
Fix incorrect cases like "05-Aug-2015" and "Monday morning 10am to 12pm"

guom08:
- Adding support for date formats like: "05-Aug-2015";
- Bug fix for "Monday morning 10am to 12pm" not correctly recognizing "am" time (regression).


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
```diff
@@ -473,75 +448,143 @@ private DateTimeResolutionResult ParseSimpleCases(string text, DateObject refere
...
-                var pmStr = match.Groups[Constants.PmGroupName].Value;
-                var amStr = match.Groups[Constants.AmGroupName].Value;
-                var descStr = match.Groups[Constants.DescGroupName].Value;
-                if (!string.IsNullOrEmpty(amStr) || !string.IsNullOrEmpty(descStr) && descStr.StartsWith("a"))
-                {
-                    if (beginHour >= Constants.HalfDayHourCount)
-                    {
-                        beginHour -= Constants.HalfDayHourCount;
-                    }
+                var pastHours = endHour - beginHour;
```
tellarin: All timex generation code should be in timex generation, not spread throughout the code.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
```diff
@@ -473,75 +448,143 @@ private DateTimeResolutionResult ParseSimpleCases(string text, DateObject refere
...
+                    DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, beginHour, 0, 0),
+                    DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, endHour, 0, 0));
+
+                ret.Success = true;
+            }
+
+            return ret;
+        }
+
+        private string ParseTimePeriod(List<ExtractResult> ers, Match match, out int beginHour, out int endHour)
```
tellarin: Thanks for extracting these methods! :)

Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -6404,5 +6404,60 @@
...
+              "type": "date",
+              "value": "2016-08-05"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Are you available on Monday morning from 10am to 12pm",
```
tellarin: Does "Monday afternoon from 4 to 6" work?

guom08: This specific case is not supported in this PR as the extraction part also needs to refine to support this case. This support would come in the upcoming PR.
But this PR enable the differentiate between "Monday 4-6" and "Monday afternoon 4-6".

Issue #818, opened 2018-08-30
[EN DateTimeV2] "the day after" gives today.


dlwh:
**To Reproduce**

* in the python repl, try out `the day after`

**Expected behavior**

Should either be tomorrow or `P1D`

**Platform (please complete the following information):**
Python 3.7, master

PR #956, opened 2018-11-08
Support cases like "the day after" and "the day before"

guom08:
#818


Python/libraries/recognizers-date-time/recognizers_date_time/date_time/english/date_parser_config.py
```diff
@@ -175,6 +175,10 @@ def get_swift_day(self, source: str) -> int:
             swift = 2
         elif trimmed_text.endswith('day before yesterday'):
             swift = -2
+        elif trimmed_text.endswith('day after') or trimmed_text.endswith('day after tmr'):
```
tellarin: This swift calculation for "tmr" seems wrong.
BTW, why is Python different from the others platforms? Shouldn't all handle this case?

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishDateParserConfiguration.cs
```diff
@@ -145,6 +145,14 @@ public int GetSwiftDay(string text)
             {
                 swift = -2;
             }
+            else if (trimmedText.EndsWith("day after"))
```
tellarin: It's best to move these hardcoded strings to the YAML files. It will avoid repeating them for every platform.
And ideally this should be a common method, that each language config could reuse or re-implement.

PR #949, opened 2018-11-02
Improve the performance of currency

Sothan:
Improve the performance of currency
- Using Matcher to replace Regex in NumberWithUnit models.
- Now, we need about 7 seconds to run all of the NumberWithUnit specs. At the same time, we need about 8 seconds to run all of the Number specs. The complexity of NumberWithUnit is subject to the complexity of Number since we need to extract numbers before units.

PR #968, opened 2018-11-12
Improve the performance of currency

Sothan:
- Add some changes to PR #949 
- Add few specs into CurrencyModel
- Improve about 21% in init
- Improve about 40% in parsing


.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -132,209 +146,154 @@ public bool ValidateUnit(string source)
...
-                    var match = ambiguousMultiplierRegex.Matches(number.Text);
-                    if (match.Count == 1)
-                    {
-                        var newLength = number.Text.Length - match[0].Length;
-                        number.Text = number.Text.Substring(0, newLength);
-                        number.Length = newLength;
-                    }
-                }
-            }
+            var nonUnitMatch = this.config.PmNonUnitRegex.Matches(source);
```
wgx998877: set this null as default, and do the matches when use it.

.NET/Microsoft.Recognizers.Text/Matcher/NumberWithUnitTokenizer.cs
```diff
@@ -0,0 +1,51 @@
...
+                return tokens;
+            }
+
+            bool inToken = false;
+            int tokenStart = 0;
+            var chars = input.ToCharArray();
+            for (int i = 0; i < chars.Length; i++)
+            {
+                var c = chars[i];
+                if (!char.IsLetter(c) || IsCjk(c))
```
wgx998877: Please add more unit with symbol cases in the test files?

.NET/Microsoft.Recognizers.Text/Matcher/NumberWithUnitTokenizer.cs
```diff
@@ -0,0 +1,94 @@
...
+                var c = chars[i];
+                if (char.IsWhiteSpace(c))
+                {
+                    if (inToken)
+                    {
+                        tokens.Add(new Token(tokenStart, i - tokenStart, input.Substring(tokenStart, i - tokenStart)));
+                        inToken = false;
+                    }
+                }
+                // Let 's$', 'us$' etc. as a token
```
tellarin: Nit: Comments should be to the side of the if or inside it.

.NET/Microsoft.Recognizers.Text/Matcher/NumberWithUnitTokenizer.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            if (inToken)
+            {
+                tokens.Add(new Token(tokenStart, chars.Length - tokenStart, input.Substring(tokenStart, chars.Length - tokenStart)));
+            }
+
+            return tokens;
+        }
+
+        private bool CaseByContext(char curChar, char preChar)
```
tellarin: I think this method can get a better name. It's not clear what "context" means here.

.NET/Microsoft.Recognizers.Text/Matcher/NumberWithUnitTokenizer.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            // Split tokens like '200ml' to two separate tokens, i.e., '200' and 'ml'
+            if ((char.IsLetter(curChar) && char.IsDigit(preChar)) || (char.IsDigit(curChar) && char.IsLetter(preChar)))
+            {
+                return true;
+            }
+
+            // For cases like '$100' or '$100' shouldn't be a token
+            if ((char.IsDigit(curChar) && preChar == '$') || (curChar == '$' && char.IsDigit(preChar)))
```
tellarin: This doesn't seem to make sense. Isn't $ a prefix and 100 a number, in a similar way to 200 as number and ml as suffix above?

tellarin: Seems only the comment is wrong.

PR #975, opened 2018-11-14
[dotnet][Datetime][Japanese] Initial .net support for datetime in japanese

gustavocalheiros:
### Status
- All files based on Chinese implementation
- Recognition of Range of dates (from....to) like: １１月１日から１２月６日まで
- Recognition of exact dates like "２０１６年１２月１日" 
- Recognition of the Japanese expressions for "today", "yesterday", months (january->december) and days of the week (monday->sunday).
- Enabled 95 out of the 300 tests that were already implemented and disabled.
- Added 13 new tests for DateRange
- Total of 108 green tests for [dotNet].

### Issues:
- Still 205 tests that are disabled for [dotNet], due to lack specific implementation to make them pass.
- On some existant tests, symbols like "昨年" (which mean "last year") are considered as "date". Maybe it should be considered as daterange. Same for "last month", "next year", etc. These tests remain disabled.


.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Extractors/DatePeriodExtractorJap.cs
```diff
@@ -0,0 +1,241 @@
...
+        public static readonly Regex ZeroToNineIntegerRegexJap = new Regex(DateTimeDefinitions.ZeroToNineIntegerRegexJap, RegexOptions.IgnoreCase | RegexOptions.Singleline);
+
+        public static readonly Regex YearInJapaneseRegex = new Regex(DateTimeDefinitions.DatePeriodYearInJapaneseRegex, RegexOptions.IgnoreCase | RegexOptions.Singleline);
+
+        public static readonly Regex MonthSuffixRegex = new Regex(DateTimeDefinitions.MonthSuffixRegex, RegexOptions.IgnoreCase | RegexOptions.Singleline);
+
+        // for case "(从)?(2017年)?一月十日到十二日" 
+        public static readonly Regex SimpleCasesRegex = new Regex(DateTimeDefinitions.SimpleCasesRegex, RegexOptions.IgnoreCase | RegexOptions.Singleline);
+
+        public static readonly Regex YearAndMonth = new Regex(DateTimeDefinitions.YearAndMonth, RegexOptions.IgnoreCase | RegexOptions.Singleline);
```
tellarin: You should be able to remove all RegexOptions.IgnoreCase from regexes in the DateTime classes.

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Extractors/DateTimeExtractorJap.cs
```diff
@@ -0,0 +1,179 @@
+﻿using System.Linq;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions.Japanese;
+using DateObject = System.DateTime;
+using Microsoft.Recognizers.Text.Number;
+
+namespace Microsoft.Recognizers.Text.DateTime.Japanese
+{
+    public class DateTimeExtractorJap : IDateTimeExtractor
```
tellarin: There is no need to add the Jap suffix. Could you rename classes to remove the suffixes?
I see that some classes in Chinese use it. We'll remove those too.

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Extractors/DurationExtractorJap.cs
```diff
@@ -0,0 +1,77 @@
...
+using Microsoft.Recognizers.Definitions.Japanese;
+using Microsoft.Recognizers.Text.Number;
+using Microsoft.Recognizers.Text.NumberWithUnit.Japanese;
+using Microsoft.Recognizers.Text.NumberWithUnit;
+
+using DateObject = System.DateTime;
+
+namespace Microsoft.Recognizers.Text.DateTime.Japanese
+{
+    public class DurationExtractorJap : BaseDateTimeExtractor<DurationType>
```
tellarin: If most of the logic of extractor/parser classes are shared between Chinese and Japanese, it would be best to call them in a pattern like CJKDurationExtractor and put them in the Microsoft.Recognizers.Text.DateTime.Extractor and Microsoft.Recognizers.Text.DateTime.Parser
But I'm fine with merging this way, and refactoring to unify later.

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/HolidayParserJap.cs
```diff
@@ -0,0 +1,414 @@
...
+        private static readonly IExtractor IntegerExtractor = new IntegerExtractor();
+
+        private static readonly IParser IntegerParser = new BaseCJKNumberParser(new JapaneseNumberParserConfiguration());
+
+        public static readonly Dictionary<string, Func<int, DateObject>> FixedHolidaysDict = new Dictionary<string, Func<int, DateObject>>
+        {
+
+            #region Fixed Date Holidays
+
+            {"元旦", NewYear},
```
tellarin: Ideally such lists and dictionaries should be moved to the target language YAML file.

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,168 +1,169 @@
-﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
```
tellarin: Nitpick: Be careful to not commit files where the only changes are line break differences (\n vs \r\n). It makes reviewing more complicated than necessary.

Specs/DateTime/Japanese/DateExtractor.json
```diff
@@ -1,1473 +1,1473 @@
-[
```
tellarin: I assume most changes in this file and the corresponding parser one are just removing the notsupported in dotnet tags and fixing a few entity boundaries?
If so, can you push a new version without the line break changes?

Specs/DateTime/Japanese/DatePeriodParser.json
```diff
@@ -0,0 +1,325 @@
+[
```
tellarin: Would you be able to help translating the other json specs? We recommend starting from the most basic types.
Other than that, the ones ending in Model are the most important ones.
Any case that still doesn't work can remain marked as notsupported. This is very valuable to track progress of proper support in a given language.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -95,6 +96,12 @@ protected override void InitializeConfiguration()
                     new BaseMergedDateTimeParser(
                         new GermanMergedParserConfiguration(new BaseOptionsConfiguration(options))),
                     new BaseMergedDateTimeExtractor(new GermanMergedExtractorConfiguration(options))));
-        }
+
+            RegisterModel<DateTimeModel>(
```
tellarin: We should only register the model to expose it to the outside once all tests for all specs pass (or the majority do and only corner cases or less common ones are left). Can you comment this one out?
Tests of all extractors/parsers are used to measure coverage. Once all look good, we add final model tests and enable it here.

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/HolidayParserJap.cs
gustavocalheiros: Sorry for this late response. About this modification, There was a mistake from my part. In this file I should have removed all the holidays, because they are related to the Chinese calendar and have nothing to do with Japanese. I'll make a new pool request to fix this.

PR #978, opened 2018-11-15
Support cases like "next Monday or next Tuesday" in ComplexCalendar mode

guom08:
Javascript and Python support are not part of this PR as this feature doesn't exist on other platforms.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
@@ -2,6 +2,7 @@ using System.Linq;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using DateObject = System.DateTime;
+using System;
 
 namespace Microsoft.Recognizers.Text.DateTime
 {
@@ -57,7 +58,7 @@ namespace Microsoft.Recognizers.Text.DateTime
                 }
                 else
                 {
-                    i = j;
+                    i++;
                 }
             }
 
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
```diff
@@ -367,42 +352,118 @@ private bool IsConnectorOrWhiteSpace(int start, int end, string text)
...
+                        i = j;
                     }
                 }
             }
-            
+
+            return isValidContext;
+        }
+
+        private Dictionary<string, object> GenerateAltData(string subType, string parentText, ExtractResult contextEr = null)
```
tellarin: I don't get what this means.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
```diff
@@ -367,42 +352,118 @@ private bool IsConnectorOrWhiteSpace(int start, int end, string text)
...
+            if (contextEr != null)
+            {
+                data.Add(Constants.Context, contextEr);
+            }
+
             return data;
         }
 
+
+        private ExtractResult ExtractRelativePrefixContext(string text)
```
tellarin: Which text is this? The whole query? Can you rename for clarity?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
```diff
@@ -367,42 +352,118 @@ private bool IsConnectorOrWhiteSpace(int start, int end, string text)
...
+
+                // DatePeriod as Context: such as "next week on Tuesday or Thursday"
+                if (datePeriodErs.Count == 1)
                 {
-                    data.Add(Constants.Context, ers[0]);
-                    data.Add(Constants.SubType, Constants.SYS_DATETIME_DATE);
+                    extractResults[0].Data = GenerateAltData(extractResults[0].Type, parentText, contextEr: null);
+                    extractResults[0].Type = ExtractorName;
+
+                    for (int i = 1; i < extractResults.Count; i++)
```
tellarin: Can you add some comments in this loop? I don't follow what is happening.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
```diff
@@ -335,7 +300,27 @@ private bool IsConnectorOrWhiteSpace(int start, int end, string text)
                 data = ExtractDate_DateRange(former, latter);
             }
 
-            return data;
+            if (data.Count == 0)
```
tellarin: Based on our offline chat, let's make sure this whole method follows the same principle. At the very lease, invert the check below so the method can be split in two separate behaviours.
Thanks.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
guom08: Changed name of this method to "CreateMetadata"

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
guom08: Change this parameter name to "entityText"

PR #979, opened 2018-11-15
Improve the performance of currency

Sothan:
- Add some changes to PR #949 
- Add few specs into CurrencyModel
- Improve about 21% in init
- Improve about 40% in parsing


tellarin:
Closing this one, so we can track the reviews in the original PR.

PR #992, opened 2018-11-19
Update German DateTime definitions, specs, Add holiday parser for German holidays

n01d:
- Add missing definitions (regex) for German DateTime Recognizers

- Correct German DateTime specs, add new DateTime specs for German

- Add an implementation for a German holiday parser based on the Chinese implementation

- Add an implementation for calculating the date of Easter for a specific year, and all public Holidays based on Easter
- Make a small change to BaseDateExtractor to allow matching for weekdays for non-English languages

- Add invariant culture to double parsing in BaseDurationParser to avoid parsing errors on Windows Systems set to a number format that uses a ',' as decimal separator


Specs/DateTime/German/DatePeriodExtractor.json
```diff
@@ -1,12 +1,12 @@
 [
   {
-    "Input": "Ich werde diesen Januar nicht hier sein.",
+    "Input": "Diesen Januar werde nicht hier sein.",
```
tellarin: Same.

Specs/DateTime/German/DatePeriodExtractor.json
```diff
@@ -1,12 +1,12 @@
...
     "NotSupported": "javascript, python",
     "Results": [
       {
-        "Text": "diesen Januar",
+        "Text": "Diesen Januar",
```
tellarin: All expected output should be in lowercase as sentence case is not preserved internally in the recognizers.

Specs/DateTime/German/DatePeriodExtractor.json
```diff
@@ -564,13 +624,13 @@
     ]
   },
   {
-    "Input": "Ich würde 2016-05 nicht hier",
+    "Input": "Ich war im 05.2016 nicht hier",
```
tellarin: Same. Here for example, my understanding is that government docs use YYYY-MM-DD and YYYY-MM, don't they?

Specs/DateTime/German/DatePeriodParser.json
```diff
@@ -925,7 +925,7 @@
     "NotSupported": "javascript, python",
     "Results": [
       {
-        "Text": "zweitausender",
+        "Text": "Die zweitausender",
```
tellarin: Lowercase

Specs/DateTime/German/DatePeriodParser.json
```diff
@@ -1069,19 +1121,19 @@
...
         },
         "Start": 10,
-        "Length": 10
+        "Length": 9
       }
     ]
   },
   {
-    "Input": "Ich würde 2016  /05 nicht hier",
+    "Input": "Ich würde 2016 . 05 nicht hier",
```
tellarin: Same

Specs/DateTime/German/DateTimeExtractor.json
```diff
@@ -36,7 +36,7 @@
     ]
   },
   {
-    "Input": "Ich gehe am 15. November um 8 Uhr weg.",
+    "Input": "Ich gehe am 15. November um 8 Uhr los.",
```
tellarin: I've seen "8 Uhr weg" and "8-Uhr weg" before. Aren't they also common/informal/regional/etc?

Specs/DateTime/German/DateTimeExtractor.json
```diff
@@ -96,7 +96,7 @@
     ]
   },
   {
-    "Input": "Das ganze ist am 14. Oktober um 8:00:00 Uhr",
+    "Input": "Das ganze ist am 14. Oktober um 8:00:00 Uhr vorbei",
```
tellarin: Don't these two sentences mean different things?

Specs/DateTime/German/DateTimeExtractor.json
```diff
@@ -120,26 +120,26 @@
     ]
   },
   {
-    "Input": "Um 8 Uhr diesen Morgen",
+    "Input": "Um 8 Uhr heute Morgen",
```
tellarin: Same

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -145,17 +175,17 @@
     ]
   },
   {
-    "Input": "Ich komme dann Freitag zurück",
+    "Input": "Ich komme am Freitag zurück",
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "NotSupported": "dotNet, javascript, python",
+    "NotSupported": "javascript, python, dotNet",
```
tellarin: No need to modify the order. Unfortunately this will break some scripts we use to eval the specs. They should be robust to ordering here, but currently they are not.

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -145,17 +175,17 @@
...
     "Results": [
       {
-        "Text": "Freitag",
-        "Start": 0,
-        "End": 0,
-        "TypeName": "datetimeV2.date",
+        "Text": "am freitag ",
+        "Start": 10,
+        "End": 20,
+        "TypeName": "datetimeV2.set",
```
tellarin: Should "am Freitag" really be a set? Set means a recurrent date, like "every monday".

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -399,16 +429,16 @@
     ]
   },
   {
-    "Input": "Ich werde am 22. April von 5 bis 6 Nachmittags nicht im Büro sein",
+    "Input": "Ich werde am 22. April von 5 bis 6 Uhr Nachmittags nicht im Büro sein",
```
tellarin: Shouldn't both work if the first one is just informal?

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -434,7 +464,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T16:12:00"
     },
-    "NotSupported": "dotNet, javascript, python",
+    "NotSupported": "javascript, python, dotNet",
```
tellarin: Same

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -446,50 +476,52 @@
...
+              "end": "2016-11-07 22:00:00"
             }
           ]
         }
       }
     ]
   },
   {
-    "Input": "Das ist dann nächsten Montag Nachmittag.",
+    "Input": "Das ist dann nächsten Montagnachmittag.",
```
tellarin: Is the two-token form not commonly used?

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -599,7 +631,7 @@
             {
               "timex": "PT123.45S",
               "type": "duration",
-              "value": "123.45"
+              "value": "123,45"
```
tellarin: There is a known issue with representing the separator here. Currently it's inconsistent.
Ideally value should be a canonical form and there should be another field for the locally-formatted value.
We can check the current Dutch to see it's usage. This will be refactored later.

Specs/DateTime/German/DateTimeParser.json
```diff
@@ -90,20 +90,20 @@
...
         },
-        "Start": 23,
+        "Start": 13,
         "Length": 21
       }
     ]
   },
   {
-    "Input": "Am 21.04.2016, 16:00 werde ich zurück sein.",
+    "Input": "Am 21.04.2016 um 16:00 Uhr werde ich zurück sein.",
```
tellarin: Same

Specs/DateTime/German/DateTimeParser.json
```diff
@@ -210,20 +210,20 @@
...
         },
-        "Start": 12,
+        "Start": 13,
         "Length": 22
       }
     ]
   },
   {
-    "Input": "Ich gehe am 5. Mai 2016 um 20 Minuten nach 8 Abends nach Hause",
+    "Input": "Ich gehe am 5. Mai 2016 um 20 Minuten nach 8 Uhr Abends nach Hause",
```
tellarin: Same

Specs/DateTime/German/DateTimeParser.json
```diff
@@ -288,14 +288,14 @@
     ]
   },
   {
-    "Input": "am 15 um 8:00:24",
+    "Input": "am 15. um 8:00:24",
```
tellarin: Same

Specs/DateTime/German/DateTimeParser.json
```diff
@@ -355,19 +355,20 @@
...
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "NotSupported": "javascript, python",
+    "Comment": "Not extracted, possibly due to a conflict between heute and morgen.",
+    "NotSupported": "dotNet, javascript, python",
     "Results": [
       {
-        "Text": "diesen Morgen um 7",
+        "Text": "heute Morgen um 7",
```
tellarin: Same

Specs/DateTime/German/DateTimePeriodExtractor.json
```diff
@@ -12,14 +12,14 @@
     ]
   },
   {
-    "Input": "Ich werde heute von 5 Uhr bis 7 Uhr weg sein",
+    "Input": "Ich werde heute von 5 - 7 Uhr weg sein",
```
tellarin: Same

Specs/DateTime/German/DateTimePeriodExtractor.json
```diff
@@ -108,25 +132,25 @@
     ]
   },
   {
-    "Input": "Ich werde wohl irgendwann diesen Nachmittag zurück gehen.",
+    "Input": "Ich werde wohl irgendwann heute Nachmittag zurück kommen.",
```
tellarin: Same

Specs/DateTime/German/DateTimePeriodParser.json
```diff
@@ -259,40 +285,40 @@
...
-        "Start": 9,
-        "Length": 12
+        "Start": 10,
+        "Length": 11
       }
     ]
   },
   {
-    "Input": "An diesem Morgen",
+    "Input": "An diesem Vormittag",
```
tellarin: Same

Specs/DateTime/German/HolidayExtractor.json
```diff
@@ -10,12 +1210,12 @@
         "Text": "Ostern",
         "Type": "date",
         "Value": {
-          "Timex": "",
+          "Timex": "XXXX",
```
tellarin: Is this the same timex returned elsewhere for non-fixed holidays?

Specs/DateTime/German/MergedExtractor.json
```diff
@@ -86,26 +86,26 @@
     ]
   },
   {
-    "Input": "06/06 12:15",
+    "Input": "6.6. 12:15",
```
tellarin: Same

Specs/DateTime/German/TimePeriodParser.json
```diff
@@ -104,14 +104,14 @@
     ]
   },
   {
-    "Input": "Ich bin von 4 Uhr Abends bis 5 Uhr Abends da",
+    "Input": "Ich bin von 4 Uhr Nachmittags bis 5 Uhr Nachmittags da",
```
tellarin: Same

Specs/DateTime/German/TimePeriodParser.json
```diff
@@ -260,7 +260,7 @@
     ]
   },
   {
-    "Input": "Das Treffen geht über den Morgen",
+    "Input": "Das Treffen geht den ganzen den Morgen",
```
tellarin: Same. I thought I've seen the first form from when I studied German :)

Specs/DateTime/German/DateTimePeriodExtractor.json
```diff
@@ -60,23 +72,35 @@
     ]
   },
   {
-    "Input": "Das Projekt ging vom 7/3/2017 15:00 bis zum 8/4/2017 16:30.",
+    "Input": "Das Projekt ging vom 7.3.2017 15:00 bis zum 8.4.2017 16:30.",
```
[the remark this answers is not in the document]
n01d: I will add this back in. We made the decision to remove / change everything that would never be written / said like this. But if you want to make sure to catch this anyway, it makes sense to add it back.

Specs/DateTime/German/DateExtractor.json
```diff
@@ -1,72 +1,87 @@
 [
   {
-    "Input": "Ich gehe am 22. April zurück",
+    "Input": "Ich komme am 22. April zurück",
```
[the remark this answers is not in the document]
n01d: "Ich gehe zurück" seems to be the simple translation "I go back". I have never heart or read anything like this in German, so we decided to change this to "Ich komme zurück".

Specs/DateTime/German/DateExtractor.json
```diff
@@ -1,72 +1,87 @@
...
         "Type": "date",
-        "Start": 12,
+        "Start": 13,
         "Length": 10
       }
     ]
   },
   {
-    "Input": "Ich gehe am 22/04 zurück",
+    "Input": "Ich komme am 22.04 zurück",
```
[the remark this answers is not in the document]
n01d: In this case I would remove it from the spec. The "/" is never used for day/year combination in German and could even lead to ambiguity is cases like "10/04" as this form is sometimes used for month/year in German.

Specs/DateTime/German/DateParser.json
```diff
@@ -42,13 +42,61 @@
...
         },
-        "Start": 12,
+        "Start": 13,
         "Length": 10
       }
     ]
   },
   {
-    "Input": "Ich gehe am 15. Januar 2011 zurück.",
+    "Input": "Ich komme am 15.1. zurück.",
```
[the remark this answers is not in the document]
n01d: The "15. Januar 2011" case is still covered by another spec. "Ich gehe zurück" might be grammatically correct but does not make sense semantically.  As the other date format is still covered I think it is ok to remove that "ich gehe" form.

Specs/DateTime/German/DatePeriodExtractor.json
n01d: I don't know why this was removed. I no have both forms in the spec and fixed a mistake in the new sentence.

Specs/DateTime/German/DatePeriodExtractor.json
n01d: In this case it was a mistake as even the input sentence hat the lover case word. But I just checked the English specs and in general the expected output text matches the case of the input here.

  {
    "Input": "I'll be out month of May",
    "Results": [
      {
        "Text": "month of May",
        "Type": "daterange",
        "Start": 12,
        "Length": 12
      }
    ]
  },

Specs/DateTime/German/DatePeriodExtractor.json
n01d: Not that I am aware of, but I added a spec for 2016-05 back in. But the sentence was still incorrectly translated and was grammatically incorrect. I added a correct sentece with 2016-05 now.

Specs/DateTime/German/DatePeriodParser.json
n01d: Both the . and / form are in the spec.

Specs/DateTime/German/DatePeriodParser.json
n01d: I believe it is case sensitive for English as well. Could you please check? See this for example:

 {
    "Input": "I'll be out from 2 to 7 Feb, two thousand and eighteen",
    "Context": {
    },
    "NotSupported": "javascript, python",
    "Results": [
      {
        "Text": "from 2 to 7 Feb, two thousand and eighteen",
        "Type": "daterange",
        "Value": {
          "Timex": "(2018-02-02,2018-02-07,P5D)",
          "FutureResolution": {
            "startDate": "2018-02-02",
            "endDate": "2018-02-07"
          },
          "PastResolution": {
            "startDate": "2018-02-02",
            "endDate": "2018-02-07"
          }
        },
        "Start": 12,
        "Length": 42
      }
    ]
  },

Specs/DateTime/German/DateTimeExtractor.json
n01d: Both are correct, both mean slightly different things. I can remember why it was changed but it should make a difference from an extraction standpoint. Both are not the correct translations for the English term in the spec anyway. ;)

Specs/DateTime/German/DateTimeExtractor.json
n01d: The initial one is not a complete sentence, so we added this word. The old sentence actually doesn't have a meaning as it lacks word(s).

Specs/DateTime/German/DateTimeExtractor.json
n01d: "diesen Morgen" is grammatically correct and understandable, but I never heart anyone say it. I will add it back for coverage.

Specs/DateTime/German/DateTimeModel.json
n01d: It commonly used, but incorrect. But as many people will use the incorrect form I now added this back in, but also left the correct form in there.

Specs/DateTime/German/DateTimeParser.json
n01d: Do we also want to catch typos or incorrect forms? Without the "." it is incorrect and should not be written that way.

Specs/DateTime/German/DateTimePeriodParser.json
n01d: Someone translated the English spec "I'll go back this morning" to just "An diesem Morgen" rather than "Ich komme an diesem Morgen zurück". And "An diesem Morgen" or "An diesem Vormittag" would both be translated to "On this Morning" in English, but mean different things in German. I will just add several variations.

Specs/DateTime/German/HolidayExtractor.json
n01d: We return XXXX for all the non-fixed holidays that like Easter or those what depend on Easter, as easter with it "Sunday after the first full moon in spring" date is a little special.

Specs/DateTime/German/MergedExtractor.json
n01d: I would not use the / form as this will case ambiguity. Is 06/06 June 6th or June 2006? In English it could be both, in German it would be rather the Month/Year. What do you think? If we want to test for 06/06 in German as well we also need to add specs to the MergedParser that returns three results day/month, month/day and month/year.

Specs/DateTime/German/TimePeriodParser.json
n01d: No one would call 4 (16) h "Abends" (Evening) but rather "Nachmittags" (Afternoon).

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -1,4 +1,6 @@
-﻿//------------------------------------------------------------------------------
+﻿
```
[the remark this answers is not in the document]
n01d: As this is a tool generated file I have not idea how this got in there in the first place. Regenerated the file now.

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -211,20 +215,47 @@ public static class DateTimeDefinitions
...
 		public static readonly string YearSuffix = $@"(,?\s*({DateYearRegex}|{FullTextYearRegex}))";
-		public const string YearAfterRegex = @"^[.]";
-		public const string YearPeriodRegex = @"(für|über)\s*(\d\s*)(Jahr(en|e)?)";
-		public const string FutureSuffixRegex = @"^[.]";
-		public const string ComplexDatePeriodRegex = @"^[.]";
+		public const string YearAfterRegex = @"\b(o\s+(nach|in der Vergangenheit))\b";
+		public static readonly string YearPeriodRegex = $@"((((von|während|zwischen)\s+)?{YearRegex}\s*({TillRegex})\s*{YearRegex})|(((zwischen)\s+){YearRegex}\s*({RangeConnectorRegex})\s*{YearRegex}))";
+		public const string FutureSuffixRegex = @"\b(in\s+der\s+)?(zukunft|zukünftig)\b";
+		public static readonly string ComplexDatePeriodRegex = $@"(((von|während|in)\s+)?(?<start>.+)\s*({TillRegex})\s*(?<end>.+)|((zwischen)\s+)(?<start>.+)\s*({RangeConnectorRegex})\s*(?<end>.+))";
+		public static readonly Dictionary<string, string> HolidayNoFixedTimex = new Dictionary<string, string>
```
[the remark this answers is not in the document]
n01d: We decided to use the same implementation as in Chinese and Japanese for the HolidayParser just with empty dates. I changed the implementation now to simplify the implementation.

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -211,20 +215,47 @@ public static class DateTimeDefinitions
...
+			{ "easterday", "" },
+			{ "eastersunday", "" },
+			{ "eastermonday", "" },
+			{ "weiberfastnacht", "" },
+			{ "carnival", "" },
+			{ "ashwednesday", "" },
+			{ "palmsunday", "" },
+			{ "goodfriday", "" },
+			{ "ascensionofchrist", "" },
+			{ "whitsunday", "" },
```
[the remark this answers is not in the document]
n01d: https://en.wikipedia.org/wiki/Whitsunday

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -608,27 +656,48 @@ public static class DateTimeDefinitions
...
-			{ "newyearday", new string[] { "neujahrstag" } },
-			{ "newyearsday", new string[] { "neujahrstag" } },
+			{ "newyearday", new string[] { "neujahrstag", "neujahr" } },
+			{ "newyearsday", new string[] { "neujahrstag", "neujahr" } },
 			{ "inaugurationday", new string[] { "inaugurationday" } },
-			{ "groundhougday", new string[] { "Murmeltiertag" } },
+			{ "groundhougday", new string[] { "murmeltiertag" } },
 			{ "valentinesday", new string[] { "valentinstag" } },
 			{ "stpatrickday", new string[] { "stpatrickday" } },
-			{ "aprilfools", new string[] { "erster april", "aprilscherz" } },
```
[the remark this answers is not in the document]
n01d: Yes, "Erster April" is correct.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestHelpers.cs
```diff
@@ -705,7 +705,7 @@ public static IDateTimeParser GetGermanParser(DateTimeParsers parserName)
                 case DateTimeParsers.Duration:
                     return new BaseDurationParser(new GermanDurationParserConfiguration(commonConfiguration));
                 case DateTimeParsers.Holiday:
-                    return new BaseHolidayParser(new GermanHolidayParserConfiguration(commonConfiguration));
+                    return new HolidayParserGer(new GermanHolidayParserConfiguration(commonConfiguration));
```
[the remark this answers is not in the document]
n01d: I did not find a quick and easy solution to do it through config and adopted the Chinese implementation for now. We can refactor that later, but then this should be refactored for Chinese and Japanese as well.

Patterns/German/German-DateTime.yaml
```diff
@@ -382,22 +382,21 @@ AnUnitRegex: !nestedRegex
...
   references: [ YearRegex, RelativeRegex ]
 HolidayRegex3: !nestedRegex
-  def: ((dieses jahr)\s*)?(?<holiday>(canberra|ostern|columbus|thanks\s*giving|weihnachten|weihnachtstag|tag der arbeit|muttertag|vatertag|female|single|teacher's|youth|children|arbor|girls|chsmilbuild|lover|labor|inauguration|groundhog|valentine's|baptiste|bastille|halloween|veterans|memorial|mid(-| )autumn|moon|spring|lantern|qingming|dragon boat|new years'|new year's|new year 's|new years|new year)\s+(day))(\s+((diesen)\s+)?({YearRegex}|{RelativeRegex}\s+jahres))?
+  def: ((dieses Jahr)\s*)?(?<holiday>(canberra|columbus|thanks\s*giving|female|single|teacher's|youth|children|arbor|girls|chsmilbuild|lover|labor|inauguration|groundhog|valentine's|baptiste|bastille|halloween|veterans|memorial|mid(-| )autumn|moon|spring|lantern|qingming|dragon boat|new years'|new year's|new year 's|new years|new year)\s+(day))(\s+((diesen)\s+)?({YearRegex}|{RelativeRegex}\s+jahres))?
```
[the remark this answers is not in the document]
n01d: I removed those that have German forms and left those where people might use the English term to reference to them as they are e.g. US only

n01d:
I added a few more fixes and updated my comment above. Please let me know if we should change anything else.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDurationParser.cs
```diff
@@ -228,7 +228,7 @@ private DateTimeResolutionResult ParseAnUnit(string text)
                     var unitStr = this.config.UnitMap[srcUnit];
 
                     ret.Timex = TimexUtility.GenerateDurationTimex(numVal, unitStr, IsLessThanDay(unitStr));
-                    ret.FutureValue = ret.PastValue = double.Parse(numStr) * this.config.UnitValueMap[srcUnit];
+                    ret.FutureValue = ret.PastValue = double.Parse(numStr, CultureInfo.InvariantCulture) * this.config.UnitValueMap[srcUnit];
```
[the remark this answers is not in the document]
n01d: The problem with this was that it return wrong values for e.g. "half year". In this class the numStr is created with 

 It would create parse this to 0,5 year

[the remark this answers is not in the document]
tellarin: You're actually correct. This can change should stay. My comment applies only to Unit.

Specs/DateTime/German/HolidayExtractor.json
```diff
@@ -4,22 +4,22 @@
...
     },
-    "NotSupported": "javascript, python, dotNet",
+    "NotSupported": "javascript, python",
     "Results": [
       {
-        "Text": "Ostern",
+        "Text": "Ostern 2020",
         "Type": "date",
         "Value": {
-          "Timex": "XXXX",
```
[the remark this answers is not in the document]
n01d: @miserale will remove that from the extractor and will post an update.

n01d:
Hi @tellarin, @miserale and I made the changes we discussed today. I hope we are good to merge the PR now.

PR #994, opened 2018-11-20
[Java][Datetime] Port BaseDateParser from C# to Java

JuanAr:
- Skipped 3 unsupported tests failing after master rebase.
- Re added Pattern.CASE_INSENSITIVE flag due to failing tests. Will remove them in another thread and PR.
- Include nit code format


Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseDurationParser.java
```diff
@@ -55,11 +55,11 @@ public DateTimeParseResult parse(ExtractResult er, LocalDateTime reference) {
 
             if (innerResult.getSuccess()) {
                 innerResult.setFutureResolution(ImmutableMap.<String, String>builder()
-                        .put(TimeTypeConstants.DURATION, StringUtility.format((Double)innerResult.getFutureValue()))
+                        .put(TimeTypeConstants.DURATION, StringUtility.format((Double) innerResult.getFutureValue()))
```
tellarin: Nit: Can we tweak the current linting rules? We typically don't add spaces after casts.

PR #995, opened 2018-11-21
Add cases for TimePeriodExtractor & TimePeriodParser in Spanish

pedrovon:
Hello, I find the division of a day quite confusing. I've established a different time division than that of the cases already there. Mine is:
por la mañana 8:00-12:00
por la tarde 13:00-20:00
por la noche 20:00-23:59
I don't understand either why the original cases define mediodía as 12:00-16:00. I've checked the Real Academia Española and other resources, maybe it's more like hours near 12:00.
Just a little comment. Thank you!

### DIFF
--- a/Specs/DateTime/Spanish/TimePeriodExtractor.json
+++ b/Specs/DateTime/Spanish/TimePeriodExtractor.json
@@ -328,5 +328,830 @@
         "Length": 9
       }
     ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


tellarin:
You are correct. "mediodía" should not be a time range. It should be the same as "noon".
The other definitions follow default ranges that can be configured differently.

PR #996, opened 2018-11-21
[Java][Datetime] Port BaseTimeZoneParser from C# to Java

JuanAr:
- Skipped 2 unsupported tests new in master.
- Code formatting

PR #997, opened 2018-11-21
[Java][Datetime] Port BaseTimeParser from C# to Java

JuanAr:
- Ported BaseTimeParser  and its unit tests
- Based on #996. We'll rebase after it is merged to master.

### DIFF
--- a/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseTimeParser.java
+++ b/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseTimeParser.java
@@ -1,0 +1,352 @@
+package com.microsoft.recognizers.text.datetime.parsers;
+
+import com.google.common.collect.ImmutableMap;
+import com.microsoft.recognizers.text.ExtractResult;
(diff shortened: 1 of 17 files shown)


Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseTimeParser.java
```diff
@@ -0,0 +1,352 @@
...
+        result.setPastValue(resultTime);
+
+        result.setSuccess(true);
+
+        return result;
+    }
+
+    private boolean isAmDesc(String descStr, Match match) {
+        return checkRegex(config.getUtilityConfiguration().getAmDescRegex(), descStr)
+        || checkRegex(config.getUtilityConfiguration().getAmPmDescRegex(), descStr)
```
tellarin: Nit: Let's fix the linting rules to keep operators at the end of the previous line.

tellarin: @JuanAr this can come in one of the next PRs.

PR #998, opened 2018-11-21
[Java][Datetime] Port BaseDatePeriodParser from C# to Java

JuanAr:
- Based on #997. We'll rebase after it is merged to master.
- Skipped 5 tests not supported in java
- Includes a fix to
    - Add Capture class
    -  Replace in MatchGroup: String[] 'captures' for Capture[] 'captures'

PR #1000, opened 2018-11-21
[Java][Datetime] Port BaseTimePeriodParser from C# to Java

JuanAr:
- Based on #998. We'll rebase after it is merged to master.


JuanAr:
Rebased with master

PR #1005, opened 2018-11-22
[NL NumberWithUnits] Various Unit fixes in .NET

iMicknl:
- Add support for `kuub` and change testcase to right unit
- Fix a few currency testcases
- Add `K` for Kelvin

@tellarin I found out that the capital 'K' was added for Kelvin, however the definitions weren't regenerated. Can you check if this is correct?

@hugo-vrijswijk can you review my changes? I changed some definitions you added in #942


Specs/NumberWithUnit/Dutch/CurrencyModel.json
```diff
@@ -198,7 +198,15 @@
         "Text": "kwartje",
         "TypeName": "currency",
         "Resolution": {
-          "unit": "Euro",
+          "unit": "Netherlands guilder",
```
tellarin: Is this right? Doesn't it apply to Euro too?

iMicknl: The Netherlands guilder had a 25 cent coin, nicknamed ‘kwartje’. With Euro we don’t use those nicknames originally. 

https://en.wikipedia.org/wiki/Twenty-five_cent_coin_(Netherlands)

tellarin: Then it looks good to me. I'll merge and we can refine if there's extra feedback. Thanks!

PR #1020, opened 2018-11-29
Fix broken test cases for DatePeriodParser

JuanAr:
- Some tests were broken due to changes in test cases.
- Also, some tests were skipped due to error in other classes that were fixed afterwards.


JuanAr:
Do not merge this PR since it braked C# unit tests. We'll review and update it.

PR #1024, opened 2018-12-03
Make "week of" cases consistency

guom08:
All "week of" cases would follow the ISO week definition, that is:
"first week of {month}/{year}": the week that has the month/year's first Thursday in it
"last week of {month}/{year}": the week that has the month/year's last Thursday in it

For "last week of year", the week number would be the actual week number
For "last week of month", the week number would be fixed at "5" because there might be more than one week number for Past and Future resolution, while we only have one Timex expression for now. So we choose the max possible number "5" to indicate "last week of month" cases.

The modifications in this PR only applied to .NET. The support for Javascript/Java/Python would come in future PR.

In this PR, I also add the support for cases like "first week of March 2015" (week of month year). But support for these cases are limited for English language. Support for other languages would come in future PR.


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1427,40 +1420,22 @@ private DateTimeResolutionResult ParseWeekOfYear(string text, DateObject referen
...
-                }
-
                 var cardinal = this.config.CardinalMap[cardinalStr];
-                targetWeekMonday = firstDayWeekThursday.This(DayOfWeek.Monday)
+                targetWeekMonday = GetFirstThursday(year).This(DayOfWeek.Monday)
                     .AddDays(Constants.WeekDayCount * (cardinal - 1));
 
-                ret.Timex = TimexUtility.GenerateWeekTimex(targetWeekMonday);
+                var weekTimex = TimexUtility.GenerateWeekTimex(cardinal);
+                var yearTimex = DateTimeFormatUtil.LuisDate(year);
```
tellarin: This should be a single timex generation method that receives the necessary input and returns the final timex.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1744,46 +1719,39 @@ private DateTimeResolutionResult ParseWhichWeek(string text, DateObject referenc
...
             {
-                pastDate = ComputeDate(cardinal, 1, month, year - 1);
-                if (pastDate.Month != month)
-                {
-                    pastDate = pastDate.AddDays(-Constants.WeekDayCount);
-                }
+                pastDate = GetMondayOfTargetWeek(cardinalStr, month, year - 1);
             }
 
+            var monthTimex = DateTimeFormatUtil.LuisDate(year, month);
```
tellarin: Let's make sure all timex generation code comes from the generator. Even in a case like this, it could be a wrapper method. Centralizing timex generation will make maintenance easier.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1744,46 +1719,39 @@ private DateTimeResolutionResult ParseWhichWeek(string text, DateObject referenc
...
             }
 
-            ret.Timex += "-W" + cardinal.ToString("D2");
+            var cardinal = GetWeekNumberForMonth(cardinalStr);
+
+            // Note that if the cardinalStr equals to "last", the cardinalNumber would be fixed at "5"
+            // This may lead to some inconsistancy between Timex and Resolution
+            // the StartDate and EndDate of the resolution would always be correct (following ISO week definition)
+            // But week number for "last week" might be inconsistancy with the resolution as we only have one Timex, but we may have past and future resolution which may have different week number
+            var weekTimex = TimexUtility.GenerateWeekTimex(cardinal);
```
tellarin: Same

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1744,46 +1719,39 @@ private DateTimeResolutionResult ParseWhichWeek(string text, DateObject referenc
...
+            var monthTimex = DateTimeFormatUtil.LuisDate(year, month);
+
             if (noYear)
             {
-                ret.Timex = "XXXX" + "-" + month.ToString("D2");
-            }
-            else
-            {
-                ret.Timex = year.ToString("D4") + "-" + month.ToString("D2");
+                monthTimex = DateTimeFormatUtil.LuisDate(Constants.InvalidYear, month);
```
tellarin: Same

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1798,24 +1766,100 @@ private static DateTimeResolutionResult GetWeekOfMonth(int cardinal, int month,
...
+        private int GetWeekNumberForMonth(string cardinalStr)
+        {
+            int cardinal;
+
+            if (config.IsLastCardinal(cardinalStr))
+            {
+                // "last week of month" might not be "5th week of month"
+                // Sometimes it can also be "4th week of month" depends on specific year and month
+                // But as we only have one Timex, so we use "5" to indicate last week of month
+                cardinal = 5;
```
tellarin: Constant

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
guom08: Fixed

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
guom08: Fixed

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
guom08: Fixed

PR #1034, opened 2018-12-05
[CI][Java] Fix resources regeneration

enzocanoo:
We found out that the resources are not being re-generated when a new build is triggered.

This pull-request:
- Update `.java` resources for all packages.
- Update the date-time `resource-definitions.json` spec to include all needed resources.
- Add a new `build.ci.cmd` script to handle the generation of resources.
- Update the `appveyor.yml` script to use `build.ci.cmd`.
- Fix some issues with new resources.


Java/build.ci.cmd
```diff
@@ -0,0 +1,17 @@
+@ECHO off
+
+ECHO # Java environment info
+CALL mvn -v
+
+ECHO # Clean resources
+DEL /S /Q libraries\recognizers-text-number\src\main\java\com\microsoft\recognizers\text\number\resources\*.java
```
tellarin: Can't we just call here the same resource generation script that already exists?

tellarin: This would avoid two build scripts.

PR #1043, opened 2018-12-06
Migrate SimpleConsole to .net core

hansmbakker:
As the title says, modernization of the project.


tellarin:
Strangely it seems to be failing with "2018-12-07T03:09:00.3349077Z     One or more projects are incompatible with .NETCoreApp,Version=v2.1.", but it works in AppVeyor.

PR #1044, opened 2018-12-07
Fix number ranges wrongly extracted as datetime ranges

WujiaShi:
Fixed the Century regex, Written Number Regex, Full Text Year Regex


JavaScript/package-lock.json
```diff
@@ -12,34 +12,34 @@
 			"resolved": "https://registry.npmjs.org/@ava/babel-preset-stage-4/-/babel-preset-stage-4-1.1.0.tgz",
 			"integrity": "sha1-rmC+iBoLq/fTX1Krp3DR9hlPdr0=",
 			"requires": {
-				"babel-plugin-check-es2015-constants": "^6.8.0",
```
tellarin: Can you revert these changes? We should only bump dependencies if really necessary for some new functionality or security fix.

Patterns/English/English-DateTime.yaml
```diff
@@ -1,1085 +1,1085 @@
----
-TillRegex: !nestedRegex
-  def: (?<till>\b(to|till|til|until|thru|through)\b|{BaseDateTime.RangeConnectorSymbolRegex})
-  references: [ BaseDateTime.RangeConnectorSymbolRegex ]
-RangeConnectorRegex : !nestedRegex
```
tellarin: Please be careful no not commit changes that are only line breaks. This makes it much harder to review a PR. Can you please revert them in this file?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -1,9529 +1,9545 @@
-[
```
tellarin: Please be careful no not commit changes that are only line breaks. This makes it much harder to review a PR. Can you please revert them in this file?

Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -8354,5 +8354,21 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I have twenty-one coins in my pocket",
```
tellarin: There are cases missing from the spec in the previous PR, right? Can you add them back here and in DateTimeModel.json?

PR #1051, opened 2018-12-08
Improve German DateTime recognition, mainly for Sets and DateTimeRanges

n01d:
Improve German DateTime recognition by
- fixing incorrect "from" token
- introducing more terms for time of day
- improving the specs
- adding further holidays
- adding the ambiguity filter for German


.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -1,4 +1,6 @@
-﻿//------------------------------------------------------------------------------
+﻿
```
tellarin: Nit: Be careful not to push this change. Not sure why it happens sometimes when VS regenerates the .cs file

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -33,8 +33,8 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "Comment": "Currently not supported. 04 will be parsed as the year",
-    "NotSupported": "dotnet, javascript, python",
+    "Comment": "Should be removed, as 22.04 could be an (incorrectly) formatted date (22.04.) or time (22:04).",
+    "NotSupported": "dotNet, javascript, python",
```
tellarin: No need for this capitalization change. It just makes the review harder. Let's always keep it lowercase.

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -33,8 +33,8 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "Comment": "Currently not supported. 04 will be parsed as the year",
-    "NotSupported": "dotnet, javascript, python",
+    "Comment": "Should be removed, as 22.04 could be an (incorrectly) formatted date (22.04.) or time (22:04).",
```
tellarin: If you saw this sentence written this way? Why interpretation would you tend to choose?

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -548,7 +542,7 @@
       "ReferenceDateTime": "2016-11-07T16:12:00"
     },
     "Comment": "Resolution is null",
-    "NotSupported": "dotnet, javascript, python",
+    "NotSupported": "dotNet, javascript, python",
```
tellarin: Any insights on why the resolution is still null?
Nit: Revert to lowercase.

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -972,8 +991,8 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T16:12:00"
     },
-    "Comment" : "Output contains a trailing whitespace",
-    "NotSupported": "dotnet, javascript, python",
+    "Comment" : "Output contains a trailing whitespace, morgens in this context means morning but is getting recognized as morgen = tomorrow",
```
tellarin: How would you write in German "I'll be away from 5 to 7 in the morning.", "I'll be away from 5 to 7 tomorrow.", and "I'll be away  tomorrow from 5 to 7."?

Specs/DateTime/German/DateTimePeriodParser.json
```diff
@@ -57,7 +57,7 @@
       "ReferenceDateTime": "2016-11-07T16:12:00"
     },
     "Comment": "Space at the end",
-    "NotSupported": "dotnet, javascript, python",
+    "NotSupported": "dotNet, javascript, python",
```
tellarin: Nit: revert.

Specs/DateTime/German/MergedExtractor.json
```diff
@@ -61,8 +61,8 @@
   },
   {
     "Input": "Das meeting geht Morgen von 12 bis 14 Uhr.",
-    "Comment": "Returning only time",
-    "NotSupported": "dotnet, javascript, python",
+    "Comment": "Returning timerange, not datetimerange",
+    "NotSupported": "dotNet, javascript, python",
```
tellarin: Same

Specs/DateTime/German/MergedParser.json
```diff
@@ -76,29 +76,74 @@
...
+      }
+    ]
+  },
+  {
+    "Input": "Letzten Freitag Nachmittag habe ich Feierabend gemacht",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "Comment": "Returning future resolution instead of past, seems to work for different reference date, see below.",
+    "NotSupported": "dotNet, javascript, python",
```
tellarin: What do you mean "seems to work for different reference date"? If this works correctly, why marked as NotSupported in dotnet?

Patterns/German/German-DateTime.yaml
```diff
@@ -1018,12 +1021,20 @@ SpecialDecadeCases: !dictionary
   entries:
     'zweitausender': 2000
 DefaultLanguageFallback: DMY
-DurationDateRestrictions: []
+DurationDateRestrictions: [ heute , jetzt , momentan , aktuell , gerade ]
+AmbiguityFiltersDict: !dictionary
```
tellarin: It seems the definitions file was not regenerated after adding this in the YAML file? Please re-gen and update the PR.

Specs/DateTime/German/MergedParser.json
n01d: The test case with "ReferenceDateTime": "2016-11-07T00:00:00" returns a future date, the next with "ReferenceDateTime": "2018-12-09T00:00:00" returns a past reference date. In the context of the sentence the past date would be correct, but I don't understand the behavior. Shouldn't this always return behave the same or return future and past? "Letzten Freitag" is "Last Friday" in the sentence.

Specs/DateTime/German/DateTimeModel.json
n01d: Currently all cases with "weekday"+"timeofday" in one sentence return null. Do you seen an easy way to map "weekday"+"timeofday" to "weekday"+ " "+"timeofday"? Do you think adding an optional weekdaygroup to the timeofday regex would work? I will try this some time this week.

Specs/DateTime/German/DateTimeModel.json
n01d: I would understands that it is day.year but only because of the "am" and as it is "22". "Ich komme 22.04 zurück" would be ambigous. This could be hh.mm or dd.yy. "Ich komme 12.04 zurück" would be almost impossible as it could be hh.mm or dd.mm or mm.yy. I would decide not to support this case with the missing training "." , at least for now.

Specs/DateTime/German/DateTimeModel.json
n01d: "I'll be away from 5 to 7 in the morning." -> "Ich werde von 5 bis 7 am morgen nicht da sein." or "Ich werde von 5 bis 7 morgens nicht da sein"
 "I'll be away from 5 to 7 tomorrow." -> "Ich werde morgen von 5 bis 7 nicht da sein" or "Ich werde von 5 bis 7 morgen nicht da sein" or "Morgen werde ich von 5 bis 7 nicht da sein"
"I'll be away tomorrow from 5 to 7." -> see above. This position of "tomorrow" wouldn't work in German.

PR #1055, opened 2018-12-11
Fix override warnings build under .Net in PT/ES/FR/GE/IT

WujiaShi:
Remove repeating TimeZoneParser in Portuguese, French, German, Spanish, and Italian. A DummyTimeZone file is added to avoid override warnings when build under .Net

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/DummyTimeZoneParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/DummyTimeZoneParser.cs
@@ -1,0 +1,29 @@
+using System.Collections.Generic;
+
+using DateObject = System.DateTime;
+
+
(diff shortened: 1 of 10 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/DummyTimeZoneParser.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        public List<DateTimeParseResult> FilterResults(string query, List<DateTimeParseResult> candidateResults)
+        {
+            return candidateResults;
+        }
+
+        public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
+        {
+            return null;
```
tellarin: If there is no output from Parse, should it really be null or an empty DateTimeParseResult?

PR #1064, opened 2018-12-17
Replace 'RegExp.test()' with RegExpUtility.isMatch()'

dfavretto:
Depends on #1065 
# Proposed Changes
  - Replace `regExp.test()` with `RegExpUtility.isMatch()` in `recognizers-number` for **Chinese**. When executing the same Regex method in a single scope, the starting index in the Regex is modified and results may be affected ([more info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#Description))

# Testing
  - Run all `JavaScript` tests at once.

### DIFF
--- a/JavaScript/packages/recognizers-number/src/number/cjkParsers.ts
+++ b/JavaScript/packages/recognizers-number/src/number/cjkParsers.ts
@@ -61,7 +61,7 @@ export class BaseCJKNumberParser extends BaseNumberParser {
         } else if (extra.includes("Num")) {
             getExtResult.text = this.replaceFullWithHalf(getExtResult.text);
             result = this.digitNumberParse(getExtResult);
-            if(this.config.negativeNumberSignRegex.test(getExtResult.text) && result.value > 0){
+            if(RegExpUtility.isMatch(this.config.negativeNumberSignRegex, getExtResult.text) && result.value > 0){
                 result.value = - result.value;
             }
             result.resolutionStr = this.toString(result.value);


.NET/Microsoft.Recognizers.Definitions/French/DateTimeDefinitions.cs
```diff
@@ -196,7 +196,7 @@ public static class DateTimeDefinitions
 		public const string InexactNumberRegex = @"\b(quelque|quel qu[ée]s|quelqu[ée]s|plusieur|plusieurs|divers)\b";
 		public static readonly string InexactNumberUnitRegex = $@"({InexactNumberRegex})\s+({DurationUnitRegex})";
 		public static readonly string RelativeTimeUnitRegex = $@"(((({ThisPrefixRegex})?)\s+({TimeUnitRegex}(\s*{NextSuffixRegex}|{PastSuffixRegex})?))|((le))\s+({RestrictedTimeUnitRegex}))";
-		public static readonly string RelativeDurationUnitRegex = $@"(((?<=({ThisPrefixRegex})\s+)?\b({DurationUnitRegex})(\s+{NextSuffixRegex}|{PastSuffixRegex})?)|((le|my))\s+({RestrictedTimeUnitRegex}))";
+		public static readonly string RelativeDurationUnitRegex = $@"((\b({DurationUnitRegex})(\s+{NextSuffixRegex}|{PastSuffixRegex})?)|((le|my))\s+({RestrictedTimeUnitRegex}))";
```
guom08: I notice that this Regex is differrent with the English implementation.
Is there any special reason to remove the "ThisPrefixRegex"?
How can we support cases like "current hour"?

.NET/Microsoft.Recognizers.Definitions/Italian/DateTimeDefinitions.cs
```diff
@@ -196,7 +196,7 @@ public static class DateTimeDefinitions
 		public const string InexactNumberRegex = @"\b(un pò|alcuni|parecchi)\b";
 		public static readonly string InexactNumberUnitRegex = $@"({InexactNumberRegex})\s+({DurationUnitRegex})";
 		public static readonly string RelativeTimeUnitRegex = $@"(((({ThisPrefixRegex})?)\s+({TimeUnitRegex}(\s*{NextSuffixRegex}|{PastSuffixRegex})?))|((le))\s+({RestrictedTimeUnitRegex}))";
-		public static readonly string RelativeDurationUnitRegex = $@"(((?<=({ThisPrefixRegex})\s+)?\b({DurationUnitRegex})(\s+{NextSuffixRegex}|{PastSuffixRegex})?)|((le|my))\s+({RestrictedTimeUnitRegex}))";
+		public static readonly string RelativeDurationUnitRegex = $@"((\b({DurationUnitRegex})(\s+{NextSuffixRegex}|{PastSuffixRegex})?)|((le|my))\s+({RestrictedTimeUnitRegex}))";
```
guom08: The same

PR #1065, opened 2018-12-17
Fix regular expressions in .yaml files

GasparAcevedoZainSouthworks:
Update `.yaml` files to solve some issues related to regular expressions _lookbehind assertions_, due to the usage of quantifiers can't be resolved by NodeJS regular expression's engine because it needs to be able to figure out _how many characters to step back_ before checking the lookbehind (more info [here](https://www.regular-expressions.info/lookaround.html#limitbehind)).

In `French-DateTime.yaml` file, the `RelativeDurationRegex` has a _lookbehind assertion_ quantified by the optional character `?`. The NodeJS's regular expression engine can't handle it. Also, the usage of both expression is mistaken because it says that the lookbehind could be there or not, so it's the same use both expressions or none of them. 
image

This also happens in the next files:
*  `Italian-DateTime.yml` (`RelativeDurationUnitRegex`)
*  `Spanish-DateTime` (`OneWordPeriodRegex`)

In `French-Numbers.yaml`, `HalfADozenRegex`, a _lookbehind assertion_ is quantified by `+`.  Its usage is mistaken not only because the NodeJS's engine can't handle it, but also for the other engines (like the .NET one) looks for a `word boundary` expression (`\b`) to appear one or more times, but it only could appear once (if it appears). Therefore, we removed the `+` quantifier.

image

These are the tests execution's results in JavaScript project:
image


.NET/Microsoft.Recognizers.Definitions/French/NumbersDefinitions.cs
```diff
@@ -78,7 +78,7 @@ public static class NumbersDefinitions
 		public static readonly string[] WrittenGroupSeparatorTexts = { "point", "points" };
 		public static readonly string[] WrittenIntegerSeparatorTexts = { "et", "-" };
 		public static readonly string[] WrittenFractionSeparatorTexts = { "et", "sur" };
-		public const string HalfADozenRegex = @"(?<=\b)+demi\s+douzaine";
+		public const string HalfADozenRegex = @"(?<=\b)demi\s+douzaine";
```
guom08: This change should be fine.

.NET/Microsoft.Recognizers.Definitions/French/DateTimeDefinitions.cs
```diff
@@ -196,7 +196,7 @@ public static class DateTimeDefinitions
 		public const string InexactNumberRegex = @"\b(quelque|quel qu[ée]s|quelqu[ée]s|plusieur|plusieurs|divers)\b";
 		public static readonly string InexactNumberUnitRegex = $@"({InexactNumberRegex})\s+({DurationUnitRegex})";
 		public static readonly string RelativeTimeUnitRegex = $@"(((({ThisPrefixRegex})?)\s+({TimeUnitRegex}(\s*{NextSuffixRegex}|{PastSuffixRegex})?))|((le))\s+({RestrictedTimeUnitRegex}))";
-		public static readonly string RelativeDurationUnitRegex = $@"(((?<=({ThisPrefixRegex})\s+)?\b({DurationUnitRegex})(\s+{NextSuffixRegex}|{PastSuffixRegex})?)|((le|my))\s+({RestrictedTimeUnitRegex}))";
+		public static readonly string RelativeDurationUnitRegex = $@"((\b({DurationUnitRegex})(\s+{NextSuffixRegex}|{PastSuffixRegex})?)|((le|my))\s+({RestrictedTimeUnitRegex}))";
```
guom08: I understand the reason for these changes.
But I also notice that this Regex is not the same implementation as the English version. I'm not quite familiar with French and Italian, is this because of some language specific reasons?
This change would make this Regex can't support cases like "current hour" in French and Italian.
I agree that this is not quite common cases, but can we find a solution to also include "ThisPrefixRegex" in this method? Maybe reference the changes in #972?

Thanks.

GasparAcevedoZainSouthworks: Hi @guom08. First of all, sorry for the delay in answering. I'm not quite familiar with French and Italian, but I assume that the differences between the English and French/Italian version of this Regex are related to the language grammar rules. Regarding the changes in the French and Italian implementation of this Regex, there will be no problems on supporting cases like "current hour/minute/day/week/month/etc", mainly because checking the tests related to this regex (DateTimeExtractor, DateTimeModel, and DateTimeParser) there are no test which cover this definition, but also because there are more than one regex that covers these cases. For example, here are some screenshots of the successful execution of a possible scenario of this regex (`Je serai dehors dans cette semaine`---> `I'll be out this week`; defined in [DatePeriodExtractor](https://github.com/Microsoft/Recognizers-Text/blob/master/Specs/DateTime/French/DatePeriodExtractor.json#L135)), both in JS and C# versions:

image

image

guom08: I see, thanks.
But for this PR, can we wait for the discussion of PR #1069? As NodeJS version 10.30 can support look behind native.

If this PR is urgent to merge, please let me know.

tellarin:
@GasparAcevedoZainSouthworks is this PR still relevant after the change to depend on Node 10.15.0?

GasparAcevedoZainSouthworks:
> @GasparAcevedoZainSouthworks is this PR still relevant after the change to depend on Node 10.15.0?

@tellarin It's necessary due to it fixes some regular expressions that will we used with the changes committed in the PR #1069, regarding the Lookbehind Assertions and Named Capture Groups

JuanAr:
@guom08 @tellarin let me know if there's any additional feedback to apply to merge these changes.

PR #1069, opened 2018-12-18
[nodejs] Update NodeJs version - add native support for lookbehind

dfavretto:
Depends on #1064 and #1065
# Proposed Changes

#### Update NodeJS version reference:
In version 10.3.0, the NodeJS RegExp engine [starts supporting](https://node.green/#ES2018) **Lookbehind Assertions** and **Named Capture Groups**, enabling us to remove the code for pre & post processing the RegExp which contain Lookbehinds (unfortunately, the Named Capture Groups logic cannot be removed as repetead names still won't work properly).

Also, the update improves the performance in the tests, as there's less logic to execute for every test, making the execution time to drop.

image


iMicknl:
Is Node ```10.30.0``` required after this PR or is it optional? If yes, this would be a breaking change and can also potential break a lot of `BotBuilder-js` bots, since they use Recognizers-Text in the `botbuilder-dialogs` package.

I would also suggest to add a `engines` property to the package.json, where you specifiy the minimum Nodejs version.

GasparAcevedoZainSouthworks:
> Is Node `10.30.0` required after this PR or is it optional? If yes, this would be a breaking change and can also potential break a lot of BotBuilder-js bots, since they use Recognizers-Text in the `botbuilder-dialogs` package.
> 
> I would also suggest to add a `engines` property to the package.json, where you specifiy the minimum Nodejs version.

Hi @iMicknl. The NodeJS v10.3.0 is required due to it being the first version which supports lookbehinds assertions and named groups natively.

Regarding the possible breaking changes, we have tested the `botbuilder-dialogs` package with a reference to a local package of `recognizers-text` with the actual PR's changes by referencing them through the package.json file in `botbuilder-dialogs`:

image

Then, we verified that these changes didn't break the project by running the tests cases as these screenshots show:
After running `npm run build` command:
image

Test results after running `npm run test` command:
image
Also, we created a PR (#709) which includes the NodeJS version update to `v10.3.0` in BotBuilder-JS.

iMicknl:
@GasparAcevedoZainSouthworks thanks! Most of the samples for `botbuilder-js` target Node.js version 8.5 or higher. In my opinion targeting 10.3.0, which is higher than the current Node.js LTS version (10.15.0), is too high for compatibility with `botbuilder-dialogs`. Or it should be a major version update due to breaking changes (semver).

@stevengum any thoughts? Which Node.js version is `botbuilder-js` actively targeting + what are people running nowadays in production.

GasparAcevedoZainSouthworks:
> @GasparAcevedoZainSouthworks thanks! Most of the samples for `botbuilder-js` target Node.js version 8.5 or higher. In my opinion targeting 10.3.0, which is higher than the current Node.js LTS version (10.15.0), is too high for compatibility with `botbuilder-dialogs`. Or it should be a major version update due to breaking changes (semver).
> 
> @stevengum any thoughts? Which Node.js version is `botbuilder-js` actively targeting + what are people running nowadays in production.

Hi @iMicknl. It seems that there is a misunderstanding regarding the targeting NodeJS version in this PR, because the actual one is not the `v10.30.0` version, but it's the NodeJS `v10.3.0`, a lower version than the `v10.15.0` (current LTS). However, it’s a good idea to use the current LTS version instead of the PR's one, so we are applying the necessary changes to accomplish this.

GasparAcevedoZainSouthworks:
@iMicknl, we have tested the NodeJS version `v10.15.0` (and the latest typescript version, [`v10.12.18`](https://www.npmjs.com/package/@types/node)) and this is the result of the tests execution in the `Recognizers-text` project:

image

These are the changes we have applied (but we haven't committed them yet):
The `engines` property section in [`package.json` ](https://github.com/Microsoft/Recognizers-Text/pull/1069/commits/c1a0aac4963e4ce21450799beb99b43019f6b0ac#diff-d1b9d925381f04c7fb098ca2077740bbR91)file:

image

The `@types/node` version, also in `package.json` file: 

image

If you agree with this, we will commit these changes.

**Note:** It's important to mention that in case the `v.10.15.0` is applied, this PR will depend on the [PR#1118](https://github.com/Microsoft/Recognizers-Text/pull/1118), because the exception explained there (`Allocation Fail Error`) will be thrown in this project for the V8 version that the NodeJS `v.10.15.0` implements.

iMicknl:
@GasparAcevedoZainSouthworks my apologies, you are right. I misread 10.3.0 for 10.30.0, however my point stays the same. I understand that the tests are running fine on the newer Node.js version and it is great that performance is improving, however we should keep current users in mind running a lower version of Node.js.

If we bump the minimum required Node.js version from 8.x to 10.x, it is a breaking change for current users of the package. According to [semver](https://semver.org/) the package version should also receive a major version bump (1.1.4 to 2.x at least).

[`botbuilder-dialogs`](https://www.npmjs.com/package/botbuilder-dialogs) has the package version pinned on `1.1.2`, after the issues with 1.1.1. It won't affect the `botbuilder-js` package at the moment, however this can affect other users of the package.

tellarin:
@GasparAcevedoZainSouthworks, I think this is Ok to merge with the ```package.json``` you mentioned.
Can you also confirm that the SampleConsole works fine in multiple cultures in #1118? If it does, I'll merge that one first, then this PR with the updated package files.

Next release of the npm packages will follow a major version bump, as suggested by @iMicknl. 

If there's no other concern, we can merge soon. Thanks.

GasparAcevedoZainSouthworks:
@tellarin The conflicts have been solved. Also, I updated the `package.json` file with the nodeJS v10.15.0.

stevengum:
@iMicknl, we're going to keep targeting 8.x in botbuilder-js out of embedded scenario support and some cloud providers not yet supporting the active LTS. At the bare minimum I expect we'll drop 8.x support after it's maintenance period ends, which is [Jan 2020](https://nodejs.org/en/about/releases/).

@sgellock or @stevenic if i'm wrong on any of the above please correct me.

stevengum:
@iMicknl, @tellarin discussions are ongoing, so disregard my earlier statements. More to follow soon

stevengum:
@iMicknl, @tellarin, we're moving to [10.3.0](https://github.com/Microsoft/botbuilder-js/pull/709) with 4.3, which will drop in 1 month's time.

tellarin:
@GasparAcevedoZainSouthworks I see that in one of the latest commit the node version is bumped to 10.12, but @stevengum mentioned 10.3 as the BotFramework target. How can we reach common ground here?

GasparAcevedoZainSouthworks:
@tellarin the reason why we updated to **v10.12.18** was that, at the moment when we created this PR, this was the latest version of **@types/node** which has native support for lookbehind assertions and named groups.

image

Also, if you check the [Version History](https://www.npmjs.com/package/@types/node?activeTab=versions) you'll see that when we created this PR, there were only two versions of @types/node **v10.x.x**: _10.12.17_ & _10.12.18_.

tellarin:
@GasparAcevedoZainSouthworks I get the version changes in the PR hsitory, my question was about syncing the final dependency to the node version that BotFramework 4.3 will target (referenced above in this PR's discussion). 
Does everything work fine if we update this PR to depend on 10.3.0?

GasparAcevedoZainSouthworks:
@tellarin The PR now depends on 10.3.0.

Issue #1030, opened 2018-12-04
[EN DateTimeV2] "from 6 to 7" not recognized as DateTime under CalendarMode


Theodosia1005:
**Describe the bug**
We expect this to return with a time range, but got no result.
DateTimeRecognizer.RecognizeDateTime("from 6 to 7", Culture.English, DateTimeOptions.CalendarMode)

**To Reproduce**
1. var results = DateTimeRecognizer.RecognizeDateTime("from 6 to 7", Culture.English, DateTimeOptions.CalendarMode)
2. results.results == 0

**Expected behavior**
got the time range in result

**Sample input/output**

**Platform (please complete the following information):**
 - Platform: .Net
 - Environment: nuget package
 - Version of package v1.1.4

**Additional context**

guom08:
We already support cases like "from 6 to 7pm" and "from 6 to 7 tomorrow" before.
But as we may also have cases like "from 6 to 7 people", so "from 6 to 7" without any context would cause some problem.

We have merged code for supporting "from 6 to 7" appear at the end of the sentence for CalendarMode, so if this time range appear at the end of the query, it's supported in CalendarMode now.

I will close this issue for now.

PR #1074, opened 2018-12-20
Fix "from 6 to 7" cannot be recognized under Calendar Mode

WujiaShi:
#1030 
Fix "from 6 to 7" cannot be recognized as a date time range under Calendar Mode when build under .Net. 
Remove outdated specs.


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
```diff
@@ -291,5 +297,26 @@ private List<Token> MatchTimeOfDay(string text)
 
             return ret;
         }
+
+        // Support cases like "from 6 to 7" which are pure number ranges
+        // Only when the number range is at the end of a sentence, it will be considered as a time range
+        private List<Token> MatchPureNumberCases(string text)
+        {
+            var ret = new List<Token>();
+            foreach (var regex in this.config.SimpleCasesRegex)
```
guom08: Why we use SimpleCasesRegex in this case? Should we have one specific Regex for the pure number case?

Specs/DateTime/English/DateTimeModelCalendarMode.json
```diff
@@ -998,5 +998,36 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I will arrange a meeting from 6 to 7",
```
guom08: Also please add cases for "I will arrange a meeting from 6 to 7."

PR #1078, opened 2018-12-26
Fix the wrong parse result of "29th" in non-leap year

Irrelevances:


JavaScript/packages/recognizers-date-time/src/dateTime/constants.ts
```diff
@@ -27,6 +27,8 @@ export class Constants {
     static readonly SemesterMonthCount: number = 6;
     static readonly TrimesterMonthCount: number = 3;
     static readonly FourDigitsYearLength: number = 4;
+    static readonly MaxMonth: number = 11;
+    static readonly MinMonth:number = 0;
```
guom08: NIT: format

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DateParser.cs
```diff
@@ -184,37 +184,93 @@ protected DateTimeResolutionResult ParseImplicitDate(string text, DateObject ref
...
                         }
 
                         if (pastDate >= referenceDate)
                         {
-                            pastDate = pastDate.AddMonths(-1);
+                            if (IsValidDate(year, month - 1, day))
+                            {
+                                pastDate = pastDate.AddMonths(-1);
+                            }
+
```
guom08: NIT: remove this blank line

PR #1081, opened 2018-12-26
Add script to process T4 templates

JuanAr:
- Created script to process all T4 templates
- Removed T4 generator from csproj to disable auto-processing
- Include BuildResources as part of Build process

### DIFF
--- a/.NET/Build.CI.cmd
+++ b/.NET/Build.CI.cmd
@@ -8,6 +8,16 @@ ECHO.
 ECHO # Restoring NuGet dependencies
 CALL nuget restore
 
+ECHO.
+ECHO # Running BuildResources.cmd
+PUSHD Microsoft.Recognizers.Definitions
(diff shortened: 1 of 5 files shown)


iMicknl:
@JuanAr does the BuildResources script work on your dev environment? I can't get it working, the error I receive is `"No input file specified."` which is thrown in [T5/TextTransform.cs](https://github.com/atifaziz/t5/blob/6f93925ad1ecbd676013e1c3abc3fbfc75f9ba39/src/TextTransform/TextTransform.cs#L78).

JuanAr:
@iMicknl I've tweaked the T4 processing approach to make it more robust in #1093. I've included your improvements as well in c39921281f98b621c025adf6eb3680c97c3f058b since I've moved BuildResources.cmd so there won't be conflicts or missing fixes.
Thanks for your feedback!

PR #1080, opened 2018-12-26
Remove duplicated tests

dfavretto:
~Depends on #1079 for better understanding the changes in the diff; this can be merged after that PR is merged.~

# Proposed Changes
  - Duplicated tests were removed from the Specs

PR #1082, opened 2018-12-27
[DO NOT MERGE] Update 'ava' dependency to v1.0.0

dfavretto:
Depends on #1080 being merged before, or tests won't pass. This is due to ava v1.0.0 failing when encountering duplicated test titles.

# Proposed changes
  - Update '[ava](https://github.com/avajs/ava)' dependency to v1.0.0 for [recent major release](https://blog.sindresorhus.com/ava-1-0-861d780b2d81).
  - **Modify title structure for DateTime tests**; right now, title is formed with test `Input`, but different tests with different `ReferenceDateTime` have identical `Input`, so [adding the `ReferenceDateTime`](https://github.com/Microsoft/Recognizers-Text/compare/master...southworkscom:feature/southworks/update-ava-dependency#diff-f4d950b7c5709d0830a08de72c79e3d6R12) to the title for making them unique.

### DIFF
--- a/JavaScript/.babelrc
+++ b/JavaScript/.babelrc
@@ -1,7 +1,7 @@
 {
   "presets": [
     [
-      "es2015",
+      "@ava/stage-4",
       {
         "modules": false
       }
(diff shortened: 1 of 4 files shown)


tellarin:
@dfavretto this should be OK to merge now, right?

Batta32:
> @dfavretto this should be OK to merge now, right?

Hi @tellarin,
Yes, we have tested these changes and they work as intended.

PR #1084, opened 2018-12-27
Handle tomorrow/morning ambiguity for German, add to Holidays

n01d:
- Handle morgen (tomorrow) / morgen (morning) ambiguity for German
- Add more holidays for German DateTime
- Update and add to test specs for Datetime German


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
```diff
@@ -35,6 +35,9 @@ public List<ExtractResult> Extract(string text, DateObject reference)
                 timePeriodErs = TimeZoneUtility.MergeTimeZones(timePeriodErs, config.TimeZoneExtractor.Extract(text, reference), text);
             }
 
+            //TODO: Quick fix to solve german morgen (morning) / morgen (tomorrow) ambiguity, move to config
+            if (text.Equals("morgen")) timePeriodErs.Clear();
```
guom08: I also notice this ambiguity thing in some of the test cases, thanks for fixing.
But this fix seems a little weird to me. How about the text is "morgen." (with dot ending) or "morgen" is part of the query?
Also, we suggest use bracket after if statement in C#.

Specs/DateTime/German/DateTimePeriodParser.json
```diff
@@ -21,7 +21,7 @@
           }
         },
         "Start": 8,
-        "Length": 26
+        "Length": 25
```
guom08: What will happen if we do have a whitespace at the end of the query?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
n01d: Fixing it with a hard match in the base class isn't a nice way to do it, but the only way we found without changing the configs for all languages. This adresses cases where the first word ist "morgen" like "morgen abend" or "morgen früh", so there will never be a "." or other non word char between those words, as this would change the meaning. I will add the brackets.

Specs/DateTime/German/DateTimePeriodParser.json
n01d: It will return the same result without the whitespace. I believe this is the expected behavior so I changed this in the spec.

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -66,12 +66,12 @@ public static class DateTimeDefinitions
 		public static readonly string OnRegex = $@"(?<=\bam\s+)({DayRegex}s?)\b";
 		public const string RelaxedOnRegex = @"(?<=\b(am|an dem)\s+)((?<day>10|11|12|13|14|15|16|17|18|19|1|20|21|22|23|24|25|26|27|28|29|2|30|31|3|4|5|6|7|8|9)([\.]))";
 		public static readonly string ThisRegex = $@"(((diese((n|m)|(\s*woche))(\s*am)?\s+){WeekDayRegex})|({WeekDayRegex})|diese(n|r)?\s*(sommer|winter|frühling|herbst))";
-		public static readonly string LastDateRegex = $@"({{LastPrefixRegex}}(\s*(woche|monat|jahr)?(\s*(am|im))?)?\s+({WeekDayRegex}|sommer|winter|frühling|herbst))|((am\s+)?{WeekDayRegex}(\s+{{LastPrefixRegex}}\s*woche))";
+		public static readonly string LastDateRegex = $@"({PastPrefixRegex}(\s*(woche|monat|jahr)?(\s*(am|im))?)?\s+({WeekDayRegex}|sommer|winter|frühling|herbst))|((am\s+)?{WeekDayRegex}(\s+{PastPrefixRegex}\s*woche))";
```
tellarin: If it's LastDateRegex, why does it reference PastPrefixRegex and not LastPrefixRegex?

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -131,29 +131,29 @@ public static class DateTimeDefinitions
...
-		public static readonly string SpecificTimeBetweenAnd = $@"(zwischen\s+)(?<time1>({TimeRegex2}|({HourRegex}|{PeriodHourNumRegex})(\s*(?<leftDesc>{DescRegex}))?))\s*{RangeConnectorRegex}\s*(?<time2>({TimeRegex2}|({HourRegex}|{PeriodHourNumRegex})(\s*(?<rightDesc>{DescRegex}))?))";
-		public const string PrepositionRegex = @"(?<prep>^(um|am|vom|von|in)(\s+(der|dem|den))?$)";
-		public const string TimeOfDayRegex = @"\b(?<timeOfDay>(((((?<early>(früh am|am frühen|früher)(\s+|-))|(?<late>(spät am|am späten|später)(\s+|-)))?(morgen(s)?(?! (früh|vor|nach|abend|nacht|morgen))|(vor|nach)mittag(s)?|abend(s)?|früh|nacht(s)?)))))\b";
-		public static readonly string SpecificTimeOfDayRegex = $@"\b((({StrictRelativeRegex}|heute)\s+{TimeOfDayRegex})\b|\bheute)s?\b";
+		public static readonly string PureNumFromTo = $@"(?<preDesc>({PmRegex}|{AmRegex})\s+)?((von)\s+)?({HourRegex}|{PeriodHourNumRegex})(\s*(?<leftDesc>({PmRegex}|{AmRegex}|{DescRegex})))?\s*{TillRegex}\s*({HourRegex}|{PeriodHourNumRegex})(?<rightDesc>\s*({PmRegex}|{AmRegex}|{DescRegex})){{0,2}}";
```
tellarin: Are these four changes for consistency with English?

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
```diff
@@ -703,7 +707,7 @@ public static class DateTimeDefinitions
 			{ @"zweitausender", 2000 }
 		};
 		public const string DefaultLanguageFallback = @"DMY";
-		public static readonly string[] DurationDateRestrictions = { "heute", "jetzt", "momentan", "aktuell", "gerade" };
+		public static readonly string[] DurationDateRestrictions = { "today", "now" };
```
tellarin: Why back in English?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
```diff
@@ -35,6 +35,12 @@ public List<ExtractResult> Extract(string text, DateObject reference)
                 timePeriodErs = TimeZoneUtility.MergeTimeZones(timePeriodErs, config.TimeZoneExtractor.Extract(text, reference), text);
             }
 
+            //TODO: Quick fix to solve german morgen (morning) / morgen (tomorrow) ambiguity, move to config
+            if (text.Equals("morgen"))
```
tellarin: Can you explain how this is solving the problem? I can't follow it from the code only.

Specs/DateTime/German/MergedParser.json
```diff
@@ -118,20 +117,21 @@
...
-        "Start": 8,
-        "Length": 18
+        "Start": 0,
+        "Length": 26
       }
     ]
   },
   {
     "Input": "Letzten Freitag Nachmittag habe ich Feierabend gemacht",
+    "Comment": "In German it should be the 07.12. not the 30.11. It is false by design",
```
tellarin: Not sure what 'false by design' means here.

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
n01d: I translated them before, but they were not used. So I put back the English terms to avoid confusion.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
n01d: The problem was that cases like "Morgen Abend" were as "Morning Evening" as "Morgen" can mean both "tomorrow" and "morning". When the extractor extracts "Abend" in the is example it will take the string before that to look for a relative shift to another day like "yesterday", "tomorrow" etc. When if tries to do this on the string "morgen" it will be extracted as a time period ("morning"), not as as "tomorrow". By filtering out "morgen" from the DateTimePeriodExtractor will fix the problem as only in the case where "morgen" is NOT a time period the string "morgen" will be passed to this extractor. I know this isn't a nice fix, but the only one we found in the given time. We already spend days trying to solve this through regex. It should also be solvable through the config but we did not want to introduce changes to the interface and configs for all other languages and did not have more time to work on this.

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
n01d: I just checked and it is the same for English. I have not idea why there is no "LastPrefixRegex". This must have been a mistake in the regex we only came across now.

.NET/Microsoft.Recognizers.Definitions/German/DateTimeDefinitions.cs
n01d: I am not sure which four changes you mean exactly. For "PureNumFromTo" we had to add the Am/Pm regex in the beginning and after the first time for German to cover cases like "Am Vormittag von 10 - 11" or "von 10 vormittags bis 11".

n01d:
Just let me know if you need more information and we can jump on a quick call as needed.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
tellarin: I suggest you create a method called GermanMorgenWorkaround that receives the extractedResults and returns the cleared version if this condition happen. In that method add this description above as comment. Leave the TODO in the current place and add "To be removed after German DateTime v1 is out."
What do you think?

Specs/DateTime/German/MergedParser.json
tellarin: @guom08 would we have a nice way of handling such cases?

tellarin:
> Just let me know if you need more information and we can jump on a quick call as needed.

I left a minor comment on the hardcoded string issue, let's do that for now and solve the conflicts, then we can merge this PR.

tellarin:
@n01d can you resolve the conflicts to trigger CI?

n01d:
Resolved the conflicts and updated the PR

PR #1093, opened 2018-12-28
[BuildProcess] Integrate T4 Transformation into BuildProcess and Visual Studio

JuanAr:
- We moved all T4 Transformation processing to a different project
- Leveraged post build event to process all T4
- Added explicit MSBUILD call to make sure resources are generated before actual build


guom08:
@JuanAr Thanks for your work on adding this resource generating part as step of build.
I have approved your changes and will merge this PR after you resolve the simple conflicts.
Thanks!

PR #1094, opened 2018-12-28
Add new sample: RecognizerFunction for detecting entities

iMicknl:
I added my RecognizerFunction (serverless API, Azure Function) to the samples directory. This can be useful when testing new releases or to make a (pre)release available for an application when LUIS.ai didn't receive the update yet.

- Uses the project references to the development Recognizers-Text packages , but you can switch it easily to a NuGet version
- Replace `$(SolutionDir)` with `$(ProjectDir)..\` due to a [known 'bug'](https://developercommunity.visualstudio.com/content/problem/172053/solutiondir-micro-becomes-undefined-in-database-pr.html). Building the solution works without this change, publishing doesn't... Is this Robycopy step still necessary? Running build.cmd still works after this change.

example URL:
```https://[app-name].azurewebsites.net/api/entities?culture=nl&text=vijfduizend```

output:
```javascript
[{"Text":"vijfduizend","Start":0,"End":10,"TypeName":"number","Resolution":{"value":"5000"}}]
```


guom08:
@iMicknl I appreciate this work for publishing the Recognizer's functionality as serverless API.
Please also note that we have a simple website for testing each type of the Recognizers:
http://recognizers.cloudapp.net/swagger/ui/index#!/DateTime/DateTime_DateTime

You may also use it if you want.

PR #1100, opened 2019-01-01
[NL DateTime] Add Start + Length properties to Dutch DateTime tests

iMicknl:
To not clutter my upcoming PR for .NET DateTime implementation for Dutch, adding all required properties to the testcases.

- Add missing Start + Length properties to add Dutch DateTime tests (autogenerated)

### DIFF
--- a/Specs/DateTime/Dutch/SetExtractor.json
+++ b/Specs/DateTime/Dutch/SetExtractor.json
@@ -4,7 +4,9 @@
     "Results": [
       {
         "Text": "wekelijks",
-        "Type": "set"
+        "Type": "set",
+        "Start": 7,
+        "Length": 9
       }
     ],
     "NotSupported": "dotNet",
@@ -15,7 +17,9 @@
     "Results": [
       {
         "Text": "dagelijks",
-        "Type": "set"
+        "Type": "set",
+        "Start": 7,
+        "Length": 9
       }
     ],
     "NotSupported": "dotNet",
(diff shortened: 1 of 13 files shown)


Specs/DateTime/Dutch/SetExtractor.json
```diff
@@ -92,7 +108,9 @@
     "Results": [
       {
         "Text": "15:00 elke dag",
-        "Type": "set"
+        "Type": "set",
+        "Start": -1,
```
guom08: I will merge this PR.
But seems we don't support these kinds of cases for now, we only support extract text which is part of the original query.

Specs/DateTime/Dutch/SetExtractor.json
```diff
@@ -103,7 +121,9 @@
     "Results": [
       {
         "Text": "15:00 iedere dag",
-        "Type": "set"
+        "Type": "set",
+        "Start": -1,
```
guom08: Same

Specs/DateTime/Dutch/SetExtractor.json
iMicknl: Resolved in #1115

Specs/DateTime/Dutch/SetExtractor.json
iMicknl: Resolved in #1115

PR #1110, opened 2019-01-02
[StyleCop] Fix all StyleCop warnings in FrenchDateTimePeriodExtractorConfiguration

JuanAr:
- Reordered fields, properties and methods. Public before private and static before non-static.
- Removed regions
- Added spaces when needed

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/French/Extractors/FrenchDateTimePeriodExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/French/Extractors/FrenchDateTimePeriodExtractorConfiguration.cs
@@ -7,47 +7,8 @@ namespace Microsoft.Recognizers.Text.DateTime.French
 {
     public class FrenchDateTimePeriodExtractorConfiguration : BaseOptionsConfiguration, IDateTimePeriodExtractorConfiguration
     {
-        public string TokenBeforeDate { get; }
-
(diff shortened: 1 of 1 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/French/Extractors/FrenchDateTimePeriodExtractorConfiguration.cs
```diff
@@ -7,47 +7,8 @@ namespace Microsoft.Recognizers.Text.DateTime.French
...
-
-        private static readonly Regex TimeTimeUnitRegex =
-            new Regex(DateTimeDefinitions.TimeUnitRegex, RegexOptions.Singleline);
-
-        private static readonly Regex TimeFollowedUnit =
-            new Regex(DateTimeDefinitions.TimeFollowedUnit, RegexOptions.Singleline);
-
         public static readonly Regex TimeNumberCombinedWithUnit =
-            new Regex(DateTimeDefinitions.TimeNumberCombinedWithUnit, RegexOptions.Singleline);
+    new Regex(DateTimeDefinitions.TimeNumberCombinedWithUnit, RegexOptions.Singleline);
```
guom08: NIT: Indentation. Please revert this change in the following PR if you can, thanks.

JuanAr: Done in https://github.com/Microsoft/Recognizers-Text/pull/1116.
Thanks for the feedback!

PR #1127, opened 2019-01-03
Fix all StyleCop warnings in EnglishTimeParserConfiguration

JuanAr:
- reordered properties and constructors
- fixed variable names
- removed extra lines

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishTimeParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishTimeParserConfiguration.cs
@@ -1,7 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Text.RegularExpressions;
-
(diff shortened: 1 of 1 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishTimeParserConfiguration.cs
```diff
@@ -114,18 +114,17 @@ public void AdjustBySuffix(string suffix, ref int hour, ref int min, ref bool ha
                         }
                     }
 
-                    var pmStr = match.Groups[Constants.PmGroupName].Value;
-                    if (!string.IsNullOrEmpty(pmStr))
+                    var time_pmStr = match.Groups[Constants.PmGroupName].Value;
```
tellarin: Isn't this a weird naming scheme?

JuanAr: Yes, we are going to change it to `matchAmStr` to better describe the content of the variable.

PR #1131, opened 2019-01-03
[StyleCop] Fix all StyleCop warnings in GermanTimeExtractorConfiguration

JuanAr:
- reordered constructor
- parameters in new line

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/German/Extractors/GermanTimeExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/German/Extractors/GermanTimeExtractorConfiguration.cs
@@ -9,11 +9,13 @@ namespace Microsoft.Recognizers.Text.DateTime.German
     {
         // part 1: smallest component
         // --------------------------------------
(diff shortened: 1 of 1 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/German/Extractors/GermanTimeExtractorConfiguration.cs
```diff
@@ -9,11 +9,13 @@ public class GermanTimeExtractorConfiguration : BaseOptionsConfiguration, ITimeE
     {
         // part 1: smallest component
         // --------------------------------------
-        public static readonly Regex DescRegex = new Regex(DateTimeDefinitions.DescRegex,
+        public static readonly Regex DescRegex = new Regex(
```
tellarin: Nit: specifically for regexes, this class is not really following the same convention as the others, right?

JuanAr: We updated to single line parameters instead.

PR #1199, opened 2019-01-08
Fix issue#1174

HaoYangMS:
Fix issue#1174


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberRangeParser.cs
```diff
@@ -69,8 +69,8 @@ private ParseResult ParseNumberRangeWhichHasTwoNum(ExtractResult extResult)
                 endValue = nums[0];
             }
 
-            var startValueStr = Config.CultureInfo != null ? startValue.ToString(Config.CultureInfo) : startValue.ToString();
-            var endValueStr = Config.CultureInfo != null ? endValue.ToString(Config.CultureInfo) : endValue.ToString();
+            var startValueStr = startValue.ToString();
```
tellarin: Isn't it better to specify InvariantCulture? This will avoid issues with the local settings on the machine where the code runs.

HaoYangMS: Thank you for your suggestion. It is very helpful. I will fix this in next iteration.

PR #1201, opened 2019-01-08
Fix Spanish DatePeriod failed cases caused by preposition and grammar

acblacktea:
Start fix Spanish DateTime Test Case for .Net, containing DatePeriodExtractor

Current status:
DatePeriodExtractor: ❌(152 of 365 still unsupported)

this pr fixed cases caused by preposition (en, desde, de, entre, desde el, desde la, desde las, del, de la, de las), including: 
1."Estare afuera en Septiembre", remove extracting "en"
2. Support relative date combination with month and day
3. Add preposition "de" in some Regexes, because some of test cases depends on this
4. Add month word such like "Sep"
5. Support the format "$day$month$preposition$day$month$preposition$year" 
6. Enable cases have already passed.


Patterns/Spanish/Spanish-DateTime.yaml
```diff
@@ -57,7 +57,9 @@ DayBetweenRegex: !nestedRegex
   def: \b((entre|entre\s+el)\s+)({DayRegex})\s*{AndRegex}\s*({DayRegex})\s+{MonthSuffixRegex}((\s+|\s*,\s*){YearRegex})?\b
   references: [ DayRegex, AndRegex, MonthSuffixRegex, YearRegex ]
 OneWordPeriodRegex: !simpleRegex
-  def: \b(((pr[oó]xim[oa]?|est[ea]|[uú]ltim[oa]?|en)\s+)?(?<month>abril|abr|agosto|ago|diciembre|dic|febrero|feb|enero|ene|julio|jul|junio|jun|marzo|mar|mayo|may|noviembre|nov|octubre|oct|septiembre|setiembre|sept|set)|(?<=\b(del|de la|el|la)\s+)?(pr[oó]xim[oa](s)?|[uú]ltim[oa]?|est(e|a))\s+(fin de semana|semana|mes|año)|fin de semana|(mes|años)? a la fecha)\b
+  def: \b(((pr[oó]xim[oa]?|est[ea]|[uú]ltim[oa]?)\s+)?(?<month>abril|abr|agosto|ago|diciembre|dic|febrero|feb|enero|ene|julio|jul|junio|jun|marzo|mar|mayo|may|noviembre|nov|octubre|oct|septiembre|setiembre|sept|set|sep)|(?<=\b(del|de la|el|la)\s+)?(pr[oó]xim[oa](s)?|[uú]ltim[oa]?|est(e|a))\s+(fin de semana|semana|mes|año)|fin de semana|(mes|años)? a la fecha)\b
```
guom08: Seems the second regex is not needed? We would like not to match "en" anyway, will just removing "en" from the first Regex work?
If the second regex is necessary, please consider to define a MonthRegex to contain all months and reference it in both OneWordPeriodRegex and OneWordInMonthPeriodRegex.

acblacktea: yes I see will modify, it's unnecessary to add another regex OneWordInMonthPeriodRegex,. In the future if I meet some bugs that need add regex focus on months,  I will define a MonthRegex .

PR #1257, opened 2019-01-10
[StyleCop] Fix all warnings in DateTime project's files (Reordered methods and using)

JuanAr:
- Reordered methods and using statements.
- Fixed spacing and trailing commas when needed.
- Renamed variables.
- Removed regions.


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/DateParser.cs
```diff
@@ -52,13 +52,18 @@ public virtual DateTimeParseResult Parse(ExtractResult er, DateObject referenceD
                 Type = er.Type,
                 Data = er.Data,
                 Value = value,
-                TimexStr = value == null ? "" : ((DateTimeResolutionResult)value).Timex,
-                ResolutionStr = ""
+                TimexStr = value == null ? V : ((DateTimeResolutionResult)value).Timex,
```
tellarin: If the idea is to use a temp variable, let's give it a better name.

tellarin: But why is it needed? Especially only in this occurrence in this class.

JuanAr: Not needed. Feedback applied.

PR #1266, opened 2019-01-10
[StyleCop] Fix all warnings in DateTime files (Reordered usings and methods)

JuanAr:
- Reordered using statements and methods.
- Fixed spacing.
- Added trailing commas and parenthesis when needed.


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -352,21 +382,22 @@ private DateTimeResolutionResult ParseImplicitDate(string text, DateObject refer
             if (match.Success)
             {
                 int month = referenceDate.Month, year = referenceDate.Year;
+
                 // create a extract result which content ordinal string of text
-                ExtractResult erTmp = new ExtractResult
+                ExtractResult ertmp = new ExtractResult
```
tellarin: Nit: ??

JuanAr: We'll better rename this one.

PR #1268, opened 2019-01-11
 Support relative references like "last", "next one", "previous one"

Irrelevances:
Fix issue 229 and issue 1027.


.NET/Microsoft.Recognizers.Text.Number/Parsers/INumberParserConfiguration.cs
```diff
@@ -13,6 +13,15 @@ public interface INumberParserConfiguration
...
 
+<<<<<<< HEAD
+        ImmutableDictionary<string, string> RelativeReferenceMap { get; }
+
+        #endregion
+
+        #region language settings
+
+=======
+>>>>>>> upstream/master
```
guom08: Be careful when you merge the code.

PR #1296, opened 2019-01-11
[StyleCop] Fix all warnings in Japanese\TimePeriodParser file

JuanAr:
- Classes were separated into different files.
- Reordered methods.
- Added spaces and trailing commas when needed.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/TimePeriodFunctions.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/TimePeriodFunctions.cs
@@ -1,0 +1,166 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using DateObject = System.DateTime;
+
+namespace Microsoft.Recognizers.Text.DateTime.Japanese
+{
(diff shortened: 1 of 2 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/TimePeriodFunctions.cs
```diff
@@ -0,0 +1,166 @@
+﻿using System;
```
guom08: The same, I'm not quite sure whether we should put this file under "Parsers" folder or not. @tellarin

tellarin: I agree with Mu. They seem generic and not Japanese-specific. Maybe in /Utilities?

JuanAr:
@tellarin @guom08 We'll close this PR and do the class refactor: unify Chinese and Japanese as they both use identical public classes and move them to the Utilities folder. We'll fix the StyleCop warnings as part of this new PR.

tellarin:
Thanks!

PR #1303, opened 2019-01-14
[CI][DotNet] Fix Build.CI.cmd to fail on build errors

JuanAr:
# Description
- Fix Build.CI.cmd to abort and throw an error if the build fails. Currently, if the compile step fails but the tests pass, the build is marked as successful.
- We found this issue after #1280 was merged which should have broken the build but it still passes due to this issue.


JuanAr:
This PR should fail (this is expected) until #1301 is merged into `master` (which fixes the error introduced in #1280)

tellarin:
Close to reopen and re-trigger build.

JuanAr:
@tellarin We'll unify the build scripts into a single `build.ci.cmd` using the more robust tool finding approach used in `build.cmd`.

PR #1310, opened 2019-01-14
[StyleCop] Fix all warnings in Japanese\TimeParser file

JuanAr:
- Classes were separated into different files.
- Reordered using statements and methods.
- Added spaces and trailing commas when needed.


.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/TimeResult.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions.Japanese;
+using DateObject = System.DateTime;
+
+namespace Microsoft.Recognizers.Text.DateTime.Japanese
```
tellarin: Similar comments to previous PRs.

PR #1325, opened 2019-01-15
[StyleCop] Fix all the warnings in NumberWithUnitExtractor

JuanAr:
- Add trailing coma in multi-line initializers
- Remove trailing whitespace
- Remove extra blank line
- Extract DinoComparer class
- Extract PrefixUnitResult class

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/DinoComparer.cs
+++ b/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/DinoComparer.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Text.Matcher;
+
+namespace Microsoft.Recognizers.Text.NumberWithUnit
+{
+    public class DinoComparer : IComparer<string>
+    {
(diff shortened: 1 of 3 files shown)


.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/DinoComparer.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Text.Matcher;
+
+namespace Microsoft.Recognizers.Text.NumberWithUnit
+{
+    public class DinoComparer : IComparer<string>
```
tellarin: Nit: @enzocanoo, do you remember what this comparer does differently? We can perhaps give it a better name. It seems it was added by you back in 7109d76e.
Not a major issue. I'll merge as is, but let's see if we can give it a clearer name.

tellarin: Perhaps also move it out of /Extractors to /Utilities as it's potentially useful elsewhere.

tellarin:
@JuanAr @Bill7zz @enzocanoo approving with minor comments.

PR #1351, opened 2019-01-18
Fixing Java build break

tellarin:
- Re-disabling "Hoy estaré fuera de cinco a siete" in DateTimePeriod parser for Java
- Disabling test in dotnet as well as behaviour seems incorrect

### DIFF
--- a/Specs/DateTime/Spanish/DateTimePeriodExtractor.json
+++ b/Specs/DateTime/Spanish/DateTimePeriodExtractor.json
@@ -12,12 +12,13 @@
   },
   {
     "Input": "Hoy estaré fuera de cinco a siete",
-    "NotSupported": "javascript, python",
+    "NotSupported": "javascript, python, java, dotnet",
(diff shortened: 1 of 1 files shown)


tellarin:
@JuanAr @guom08 It seems this case was brought back in one of the recent PRs. My bad that I didn't notice it. 

I've added some comments in the spec. There is no datetimeperiod in that sentence, only two separate entities: a Date and a TimePeriod.

This test in DateTimePeriodExtractor.json shouldn't extract anything ("[]"), but on DateTimeModel.json it should extract both separate entities.

tellarin:
@JuanAr Interestingly too, when the unit test fails in AppVeyor, the build log just aborts.

The last line there is 
"Running com.microsoft.recognizers.text.tests.numberwithunit.NumberWithUnitTest",
so it's not straightforward to see that one test case failed for a specific spec, unless running the build locally. 

Could we somehow improve it?

JuanAr:
@tellarin strange indeed. We'll take care of this and fix the build script.

PR #1352, opened 2019-01-18
Improve the performance of TimeZone: replacing regexes with StringMat…

Sothan:
Improve the performance of TimeZone by replacing regex with StringMatcher
- The optimized processing speed of TimeZone is about 3.7 times as the previous version.


Patterns/English/English-TimeZone.yaml
```diff
@@ -4,8 +4,301 @@ DirectUtcRegex: !simpleRegex
   def: \b(utc|gmt)(\s*[+\-\u00B1]?\s*[\d]{1,2}(\s*:\s*[\d]{1,2})?)?\b
 AbbreviationsRegex: !simpleRegex
   def: (\b)(AFT|AKST|AKDT|ABST|AST|ARBST|ARST|ART|ADT|ACST|AEST|AEDT|AZT|AZST|AZOT|AZOST|BST|CCST|CT|CVT|AMT|AMST|ACDT|CAST|CADT|BTT|CBT|CBST|CET|CEST|SBT|CST|CDT|CSTM|CDTM|BIT|ET|EAT|EET|EEST|ESAT|ESAST|EST|EDT|ESTM|EDTM|EGST|YEKT|FJT|FJST|GET|GMT|GNST|GNDT|GST|GTBST|HAST|HST|HADT|HDT|IST|IRST|IRDT|ISST|ISDT|JST|JDT|PETT|KST|LINT|MAGT|MAGST|MUT|MAT|MEST|MVST|MOST|WEST|MST|MDT|MSTM|MDTM|MYST|NCAST|NMST|NMDT|NPT|NZT|NZST|NZDT|NST|NDT|IRKT|KRAT|PSAT|PSAST|PST|PDT|PSTM|PDTM|PKT|PYT|PYST|RST|MSK|SAEST|SAPST|SAWST|SST|SMST|THA|SNST|SAST|SLT|SGT|TIST|TAST|TADT|TST|TOT|TOST|TRT|ULAT|UTC|VET|VLAT|AWST|WAT|ECT|WAST|WET|WPST|YAKT|PDST|PT|AWDT|COT|HKT)(\b)
+AbbreviationsList: !list
```
guom08: Is FullNameRegex and AbbreviationsRegex still used in some piece of the code?
If not, I suggest to remove these two Regexes from this yaml file.
Thanks.

guom08: Build failed

Sothan: C# does not currently use these two regexes, but Java and C# are not synchronized, so they are not deleted for now. These two regexes should be deleted after Java implements StringMatcher.

.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
```diff
@@ -20,6 +20,299 @@ public static class TimeZoneDefinitions
...
+		public static readonly IList<string> AbbreviationsList = new List<string>
+		{
+			@"AFT",
```
tellarin: No need to duplicate regex and list. You can pre-process the regexes to build these lists.

Patterns/English/English-TimeZone.yaml
tellarin: @Sothan I left a comment in the review. Can you also implement the Java matcher?

Sothan: Sorry, the StringMatcher is already implemented in Java, I will refactor this PR ASAP.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseTimeZoneExtractor.java
```diff
@@ -90,12 +90,20 @@ public BaseTimeZoneExtractor(ITimeZoneExtractorConfiguration config) {
 
     private List<Token> matchTimeZones(String text) {
         List<Token> ret = new ArrayList<>();
-        for (Pattern regex : config.getTimeZoneRegexes()) {
-            Match[] matches = RegExpUtility.getMatches(regex, text);
-            for (Match match : matches) {
+
+        // Direct TUC matches
```
tellarin: Typo: UTC

PR #1377, opened 2019-01-22
[StyleCop] Fix StyleCop warnings in IDateParser

JuanAr:
-Add spaces to comments
-Change "" to string.Empty
-Add the file TimeTypeConstants with the class TimeTypeConstants
-Add rule to ignore the underscore in the TimeTypeConstants variables names


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/TimeTypeConstants.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using DateObject = System.DateTime;
+
+namespace Microsoft.Recognizers.Text.DateTime
+{
+    [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1310: CSharp.Naming : Field names must not contain underscores.", Justification = "Constant names are written in upper case so they can be readily distinguished from camel case variable names.")]
+    public static class TimeTypeConstants
```
tellarin: This file should probably go out of the /Parsers directory.

PR #1387, opened 2019-01-23
Improve the performance of LocationTimeZoneMatcher by pre-check

Sothan:
- Remove the duplicate items appeared in FullNameList and MajorLocations
- Improve the performance of LocationTimeZoneMatcher by pre-check


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeZoneExtractor.cs
```diff
@@ -51,7 +51,35 @@ private IEnumerable<Token> MatchLocationTimes(string text)
 
             var timeMatch = config.LocationTimeSuffixRegex.Matches(text);
```
tellarin: This cost is being paid even if other entities ending in "time" were already extracted. Can't this be improved?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeZoneExtractor.cs
```diff
@@ -51,7 +51,35 @@ private IEnumerable<Token> MatchLocationTimes(string text)
...
 
-            if (timeMatch.Count != 0)
+            // Before calling a Find() in location matcher, check if all the matched suffixes by
+            // LocationTimeSuffixRegex are already inside tokens extracted by TimeZone matcher.
+            // If so, don't call the Find() as they have been extracted by TimeZone matcher, otherwise, call it.
```
tellarin: Is it worth it saving only the Find time? It seems there's more benefit to try and improve the earlier step.

tellarin: After sync offline, there's not much difference between the two alternatives.

PR #1397, opened 2019-01-24
Fix build break after timezone optimization changes made Dutch out of sync

tellarin:
- Fix build break after timezone optimization changes made Dutch out of sync;
- Minor refactoring to reuse new timezone match loader;
- Removed most non-localized string in Dutch TimeZone YAML. Re-use needs to be improved later;
- Small stylecop fixes in Dutch.


.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimeZoneUtility.cs
```diff
@@ -52,5 +55,22 @@ public static bool ShouldResolveTimeZone(ExtractResult er, DateTimeOptions optio
 
             return enablePreview && hasTimeZoneData;
         }
+
+        public static StringMatcher BuildMatcherFromLists(params List<string>[] collections)
+        {
+            StringMatcher matcher = new StringMatcher(MatchStrategy.TrieTree, new NumberWithUnitTokenizer());
+            List<string> matcherList = new List<string>();
+
+            foreach (List<string> collection in collections)
```
wgx998877: we can selectMany from the collections to make code simplier

iMicknl:
Thanks @tellarin!

PR #1400, opened 2019-01-24
Support case “next”

Irrelevances:


Python/libraries/recognizers-number/recognizers_number/resources/english_numeric.py
```diff
@@ -231,5 +231,6 @@ class EnglishNumeric:
                                  ("next to last", "CURR-1"),
                                  ("penultimate", "CURR-1"),
                                  ("the last but one", "CURR-1"),
-                                 ("antepenultimate", "CURR-2")])
+                                 ("antepenultimate", "CURR-2"),
```
tellarin: I see many mistakes. antepenultimate should be N-2, not CURR-2, for example.

Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/resources/EnglishNumeric.java
```diff
@@ -75,7 +75,7 @@ public static String NumbersWithPlaceHolder(String placeholder) {
 
     public static final String NumberOrdinalRegex = "(first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth|eleventh|twelfth|thirteenth|fourteenth|fifteenth|sixteenth|seventeenth|eighteenth|nineteenth|twentieth|thirtieth|fortieth|fiftieth|sixtieth|seventieth|eightieth|ninetieth)";
 
-    public static final String RelativeOrdinalRegex = "(last|next one|previous one|the second to last|the one before the last one|next to last|penultimate|the last but one|antepenultimate)";
+    public static final String RelativeOrdinalRegex = "(last|next|next one|previous one|the second to last|the one before the last one|next to last|penultimate|the last but one|antepenultimate)";
```
tellarin: This seems strange. Was this file auto-generated? "next one" should come before "next" for it to work correctly. Also, why is the order in the regex only different in Java?

Patterns/English/English-Numbers.yaml
```diff
@@ -49,7 +49,7 @@ RoundNumberOrdinalRegex: !simpleRegex
 NumberOrdinalRegex: !simpleRegex
   def: (first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth|eleventh|twelfth|thirteenth|fourteenth|fifteenth|sixteenth|seventeenth|eighteenth|nineteenth|twentieth|thirtieth|fortieth|fiftieth|sixtieth|seventieth|eightieth|ninetieth)
 RelativeOrdinalRegex: !simpleRegex
-  def: (last|next one|previous one|the second to last|the one before the last one|next to last|penultimate|the last but one|antepenultimate)
+  def: (last|next one|previous one|the second to last|the one before the last one|next to last|penultimate|the last but one|antepenultimate|next)
```
tellarin: This regex can be optimized quite a bit with groups and optional tokens, no?

PR #1404, opened 2019-01-24
[JavaScript] Rename DinoComparer to StringComparer and update implementation 

JuanAr:
# Description

- Change the method DinoComparer to StringComparer for more clarity within the code. This method will take two strings and compare them, taking into consideration that they may be null or empty. 
- The method was refactored for a more clear implementation. 
- The extractor.ts file was updated to work with the changed name.

All the tests passed succesfully

imagen

### DIFF
--- a/JavaScript/packages/recognizers-number-with-unit/src/numberWithUnit/extractors.ts
+++ b/JavaScript/packages/recognizers-number-with-unit/src/numberWithUnit/extractors.ts
@@ -330,16 +330,16 @@ export class NumberWithUnitExtractor implements IExtractor {
         return RegExpUtility.getSafeRegExp(pattern, options);
     }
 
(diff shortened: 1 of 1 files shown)


JavaScript/packages/recognizers-number-with-unit/src/numberWithUnit/extractors.ts
```diff
@@ -322,66 +322,27 @@ export class NumberWithUnitExtractor implements IExtractor {
...
-                    if (xl > yl) {
-                        return 1;
-                    }
-
-                    return 0;
-                }
-            }
+        else
+        {
+            if (!stringA) return 1;
```
tellarin: To keep with the code standard, can you add all the {}?

PR #1407, opened 2019-01-24
[Java][Datetime] Remove unused regexes from English and Spanish

GasparAcevedoZainSouthworks:
# Description
Remove unused regexes from English, and Spanish parsers and extractors. This is just to clean the code. Also fixed some porting errors that were found while applying these changes.

PR #1408, opened 2019-01-24
[DotNet][Datetime] Remove unused regexes from English, Spanish, and Chinese

JuanAr:
*Disclaimer*: _Because we are removing multiple unused `public static readonly` properties, this involves a breaking change if final users are relying on them. However, they are not being used anymore anywhere in the code.
Our suggestion, if we want to maintain the external interface intact is to review and explicitly expose them or comment on those properties to keep._

# Description
Remove unused regexes from English, Spanish, and Chinese parsers and extractors. This is just to clean the code.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/DateExtractorChs.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/DateExtractorChs.cs
@@ -14,15 +14,6 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
 
         public static readonly Regex DayRegex = new Regex(DateTimeDefinitions.DayRegex, RegexOptions.Singleline);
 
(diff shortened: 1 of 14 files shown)


ParadoxARG:
We are going to validate that the regexes we are removing are not being used in the main extractors(English and Chinese) before removing them. If for example they are being used in English but not in French we are going to keep the unused regex and add decorate it with the following comment: `// The functionality for this regex has not yet been implemented`.
BTW, we are not editing the yaml files, so the regexes are not being removed from the project. The only thing being removed is the unused instantiations.

ParadoxARG:
Changes applied.

tellarin:
I'll temporarily close the 4 PRs (#1407, #1408, #1421, #1426) related to removing regexes as there was no time to properly review and many folks here will be on a week-long holiday this week. I'll re-open when everyone is back.
Sorry for the delay.

PR #1419, opened 2019-01-25
[StyleCop] Fix warnings in Text.Number Project

JuanAr:
- XML comment analysis warning

SpanishNumberRangeParserConfiguration  …
- Move constructor before properties
- Move constructor initializers on their own line
- Add class access modifier

BasePercentageExtractor  …
- Reorder properties a fields
- Set field to a property to keep the variable name with upper-case
- Add returns comments in methods
- Add trailing coma in multi-line initializers
- Use String.Empty instead empty strings

NumberMode  …
- Update comment to fix CS1570 XML comment has badly formed

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.csproj
+++ b/.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.csproj
@@ -8,6 +8,18 @@
     <CodeAnalysisRuleSet>../Recognizers-Text.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <PropertyGroup>
+	  <!--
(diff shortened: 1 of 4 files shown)


.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.csproj
```diff
@@ -8,6 +8,18 @@
     <CodeAnalysisRuleSet>../Recognizers-Text.ruleset</CodeAnalysisRuleSet>
   </PropertyGroup>
 
+  <PropertyGroup>
+	  <!--
+		Make sure any documentation comments which are included in code get checked for syntax during the build, but do
```
tellarin: Nit: indentation level is inconsistent with the rest of the file.

PR #1425, opened 2019-01-28
[StyleCop] Make Build process fail on Stylecop issues

ceciliaavila:
Disclaimer: this PR will fail until we fix all StyleCop's errors.

- Changed ruleset to force an error instead of a warning.
- Fixed ruleset reference in Definitions.csproj
- Added rule CS8019 regarding unnecessary using directives.


tellarin:
Interestingly the AppVeyor build fails, but the VSO doesn't.

JuanAr:
@tellarin We'll setup a VSO build in our repo and let you know if there's any extra step to make this solution work in VSO.

tellarin:
@JuanAr I've just noticed that the build in AppVeyor is no longer keeping the complete build log. I've tried download the log from some of the latest builds and they are always truncated.
Not sure if this is actually recent or not, but last time I looked into this, the full log was there.
Did the build output become much longer than previously?

tellarin:
I've also noticed that the output of a build in Visual Studio seems out of order. At least in my local dev env it seems all projects are built first, them the resources are generated from the YAML files.

JuanAr:
@tellarin I've checked a few latest builds and could download the logs. SA warnings increased the log length when we enabled Style and might have created a transient problem with logs. This should no longer be an issue since SA warnings decreased significantly.

In this specific PR, we are turning SA warnings into errors and since there are still some SA warnings, the build process will stop after building .NET. We might want to store the issues, continue building other languages, and finally fail with the stored issues info. This will be easier with PowerShell.

Regarding the order of YAML processing. I've just checked and it seems to be working as expected for the root's build.cmd (builds all platforms.) However, I wonder if this is specific of .NET which uses a Post-Build script on `Microsoft.Recognizers.Definitions.Common.csproj`. I'll take a look at that unless you experienced it in another platform.

tellarin:
@Aliandi, can you resolve this last conflict? I can merge after that.
Thanks!

Aliandi:
@tellarin Conflicts solved!

PR #1438, opened 2019-01-31
Fix Spanish DatePeriod cases

acblacktea:
Fix Spanish DatePeriod cases below type.

1.Ya no estaré en la 3ª semana de 2027
2.semana de 15 septiembre
3.Me iré en el fin de semana


Patterns/Spanish/Spanish-DateTime.yaml
```diff
@@ -100,10 +100,9 @@ SeasonRegex: !nestedRegex
 WhichWeekRegex: !simpleRegex
   def: (semana)(\s*)(?<number>\d\d|\d|0\d)
 WeekOfRegex: !simpleRegex
-  def: (semana)(\s*)((do|da|de))
-MonthOfRegex: !nestedRegex
-  def: (mes)(\s+)((do|da|de)\s+){MonthRegex}
-  references: [MonthRegex]
+  def: ((del|de|la|el)\s+)?(semana)(\s*)((do|da|del|de))
+MonthOfRegex: !simpleRegex
+  def: (mes)(\s+)((do|da|del|de))
```
guom08: Seems the nested bracket at last is not necessary.
Also is it possible to make this (do|da|del|de) separate regex and reference it in WeekOfRegex, MonthOfRegex and OneWordPeriodRegex? Thanks.

acblacktea: yes, will modify these problem code, thanks

PR #1441, opened 2019-02-02
Removing unnecessary properties from SimpleConsole project

tellarin:


tellarin:
It was bothering me. :)

PR #1452, opened 2019-02-12
Change Portuguese to Brazilian Portuguese

iMicknl:
The README describes that Recognizers-Text supports the Portuguese language, however as far as I know it supports Brazilian Portuguese which is different.


tellarin:
@iMicknl, actually it should cover all variants of Portuguese. So I'll close this  PR, ok?

iMicknl:
@tellarin if that's the case, cool! Thank for confirming.

PR #1461, opened 2019-02-18
[NL NumberRange] Add basic .NET support for NumberRange model

iMicknl:
Still a work in progress, but already here for a long time. Would prefer to merge this already in, so others can build upon the current work.

Recognizer is disabled, only tests are running atm.


tellarin:
It seems 25 tests are failing in .NET in the CI build. I'll start reviewing the current code soon.

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
```diff
@@ -274,13 +274,11 @@ protected override void InitializeConfiguration()
                     AgnosticNumberParserFactory.GetParser(AgnosticNumberParserType.Percentage, new DutchNumberParserConfiguration()),
                     new Dutch.PercentageExtractor(options)));
 
-            /*
             RegisterModel<NumberRangeModel>(
```
tellarin: This is needed to run the corresponding tests, right? But we can't merge it in with the recognizer registered.

iMicknl: I found this out also yesterday, after my PR... I will try to take some time for this this week, to clean it up and debug the failing tests. There are more tests failing than I thought.

Would prefer not to merge it in + enable tests, without almost full coverage.

Specs/Number/Dutch/NumberRangeModel.json
```diff
@@ -286,24 +265,23 @@
   },
   {
     "Input": "Meer dan de helft van de mensen is aanwezig.",
-    "NotSupported": "dotNet",
     "Results": [
       {
-        "Text": "Meer dan de helft",
+        "Text": "meer dan de helft",
```
iMicknl: @tellarin Should the NumberRange recognizer convert everything to lowercase? Or is this a bug and should the specs be changed back to first letter in uppercase.

Specs/Number/Dutch/NumberRangeModel.json
```diff
@@ -286,24 +265,23 @@
...
+    "NotSupported": "dotNet",
     "NotSupportedByDesign": "javascript,python,java"
   },
   {
     "Input": "Meer dan 1/2 van de mensen is aanwezig.",
-    "NotSupported": "dotNet",
     "Results": [
       {
-        "Text": "Meer dan 1/2",
+        "Text": "meer dan 1/2",
```
iMicknl: Same as `meer dan de helft`

iMicknl:
@tellarin I did clean up the PR and all tests are ok. Disabled the recognizer (and the tests).

I will try to take some time for a follow up PR. I am not sure about some current implementations, regarding the boundaries. English also contains a different experimental NumberRange model. Dutch will need some work to support specific Dutch ways and to double check existing implementation.

For now;
Clippy_ShipIt

PR #1465, opened 2019-02-20
Support phone number cases like "000 111 82-2100"

Irrelevances:


.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
```diff
@@ -1,55 +1,55 @@
-//------------------------------------------------------------------------------
```
tellarin: Please be careful to not change all line endings in a file if not really necessary. Can you undo this?

.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
```diff
@@ -1,55 +1,55 @@
...
+//------------------------------------------------------------------------------
+namespace Microsoft.Recognizers.Definitions
+{
+	using System;
+	using System.Collections.Generic;
+
+	public static class BasePhoneNumbers
+	{
+		public const string NumberReplaceToken = @"@builtin.phonenumber";
+		public const string GeneralPhoneNumberRegex = @"(\b(((\d[\s]?){4,12}))(-?[\d\s?]{3}\d)(?!-)\b)|(\(\d{5}\)\s?\d{5,6})|\+\d{2}\(\d\)\d{10}";
```
tellarin: I don't think this is the right place to add support for this number format. It doesn't look like a general/common format to me.

Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -1,4 +1,19 @@
...
+  {
+    "Input": "Tel: 000 111 82-2100.",
+    "Results": [
+      {
+        "Text": "000 111 82-2100",
+        "Start": 5,
+        "End": 19,
+        "TypeName": "phonenumber",
+        "Resolution": {
+          "score": "1",
```
tellarin: You can see the side effect here. This is not a typical number, so it should not have a 1.0 score.
Perhaps SpecialPhoneNumberRegex or somewhere else is a better place to match it. As for the appropriate score, take a look at the other spec cases to see if the score given it to it makes sense compared to others.

.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
Irrelevances: If I made a common regex like "(\b(((\d[\s-]?){7,15}\d))(?!-)\b)|(\(\d{5}\)\s?\d{5,6})|\+\d{2}\(\d\)\d{10}",  "12-12-2012" was recognized as a phone number, how can I distinguish them?

Issue #1170, opened 2019-01-04
[* Sequence] Phone numbers should match without area code


chrimc62:
**Describe the bug**
Phone numbers should match without area or country codes, i.e. in the US "765-6612" should be a phone number.  On LUIS currently, this is not the case.

PR #1487, opened 2019-03-05
Matching US phone number without area code (#1170)

sherryzyy:
Issue: #1170

For example, my phone number is 223-4567.
223-4567 will be recognized and have a score of 0.55. 
Because the first number of the US phone number cannot be 1 or 0, so the number 123-4567 will not be recognized as a phone number.

Support also in JavaScript and python.


.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
```diff
@@ -22,12 +22,13 @@ public static class BasePhoneNumbers
...
+		public const string USPhoneNumberRegex = @"(((((\B\+)|\b)1(\s|-)?)|\b)?(\d{3}\)[-\s]?|\(\d{3}\)[-\s]?|\b\d{3}\s?[-\.]?\s?)|\b)\d{3}\s?[-\.]?\s?\d{4}(\s?(x|X|ext)\s?\d{3,5})?(?!(-\s?\d))\b";
 		public const string CNPhoneNumberRegex = @"((\b00\s?)?\+?86\s?-?\s?)?(((\b|(?<=86))\d{2,5}\s?-?\s?|\(\d{2,5}\)\s?)\d{4}\s?-?\s?\d{4}(\s?-?\s?\d{4})?|(\b|(?<=86))\d{3}\s?-?\s?\d{4}\s?-?\s?\d{4})(?!-)\b";
 		public const string DKPhoneNumberRegex = @"((\(\s?(\+\s?|00)45\s?\)\s?)|(((?<!\d)\+\s?|\b00)45\s?)|\b)(\s?\(0\)\s?)?((\d{8})|(\d{4}\s?-?\s?\d{4,6})|((\d{2}[\s-]){3}\d{2})|(\d{2}\s?-?\s?\d{3}\s?-?\s?\d{3}))(?!-)\b";
 		public const string ITPhoneNumberRegex = @"((\(\s?(\+\s?|00)39\s?\)\s?)|(((?<!\d)\+\s?|\b00)39\s?)|\b)((0[\d\s-]{4,12}\d)|(3[\d\s-]{7,12}\d))(?!-)\b";
 		public const string NLPhoneNumberRegex = @"((((\(\s?(\+\s?|00)31\s?\)\s?)|(((?<!\d)\+\s?|\b00)31\s?))?(((\b|(?<=31))0?\d{1,3}|\(\s?0?\d{1,3}\s?\)|\(0\)[-\s]?\d{1,3})\s?-?[\d\s]{5,11}\d))|\b\d{10,12})(?!-)\b";
 		public const string SpecialPhoneNumberRegex = @"\b(\d{3,4}[/-]\d{1,4}[/-]\d{3,4})\b";
+		public const string SpecialUSPhoneNumberRegex = @"(?<!(-|-\s|\d|\)|\)\s|\.))\d{3}\s?[-\.]\s?\d{4}(?!(-\s?\d))\b";
```
tellarin: It's best to rename these new regexes and vars to something like NoAreaCodeUSNumber. "Special" is a very ambiguous term.

.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
```diff
@@ -22,12 +22,13 @@ public static class BasePhoneNumbers
 		public const string BRPhoneNumberRegex = @"(((\B\(\s?))\d{2,3}(\s?\))|(\b\d{2,3}))\s?\d{4,5}-?\d{3,5}(?!-)\b";
 		public const string UKPhoneNumberRegex = @"(((\b(00)|\B\+)\s?)?(\b\d{2}\s?)?((\s?\(0\)[-\s]?|\b|(?<=(\b^#)\d{2}))\d{2,5}|\(0\d{3,4}\))[/-]?\s?(\d{5,8}|\d{3,4}[-\s]?\d{3,4})(?!-)\b)";
 		public const string DEPhoneNumberRegex = @"((\+\d{2}\s?((\(0\))?\d\s?)?|\b)(\d{2,4}\s?[-/]?[\s\d]{7,10}\d)(?!-)\b)";
-		public const string USPhoneNumberRegex = @"((((\B\+)|\b)1(\s|-)?)|\b)?(\d{3}\)[-\s]?|\(\d{3}\)[-\s]?|\b\d{3}\s?[-\.]?\s?)\d{3}\s?[-\.]?\s?\d{4}(\s?(x|X|ext)\s?\d{3,5})?(?!(-\s?\d))\b";
+		public const string USPhoneNumberRegex = @"(((((\B\+)|\b)1(\s|-)?)|\b)?(\d{3}\)[-\s]?|\(\d{3}\)[-\s]?|\b\d{3}\s?[-\.]?\s?)|\b)\d{3}\s?[-\.]?\s?\d{4}(\s?(x|X|ext)\s?\d{3,5})?(?!(-\s?\d))\b";
```
tellarin: Is this change making the regex also match without area code?

.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
tellarin: Also, this regex is for scoring only (not tagging), right? If so, can you add a comment explaining that in the YAML file?

.NET/Microsoft.Recognizers.Text.Sequence/English/Parsers/PhoneNumberParser.cs
```diff
@@ -81,6 +83,12 @@ public double ScorePhoneNumber(string phoneNumberText)
             // Continue digit deduction
             score -= Math.Max(Regex.Matches(phoneNumberText, continueDigitRegex).Count - 1, 0) * continueDigitDeductionScore;
 
+            // Special award for special USphonenumber, i.e. 123-4567 or 123 - 4567
+            if (SpecialUSphonenumbeRegex.IsMatch(phoneNumberText))
+            {
+                score += specialUSphoneNumberAward;
```
tellarin: If I understand it right, this is to compensate some penalty from other scoring, right? If so, could it be relative to the relevant penalty, instead of a hardcoded value?

.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
sherryzyy: Yes, it's only for scoring. I'll add a comment in the YAML file.

.NET/Microsoft.Recognizers.Definitions.Common/BasePhoneNumbers.cs
sherryzyy: Yes, the change makes the regex also match without the area code.

.NET/Microsoft.Recognizers.Text.Sequence/English/Parsers/PhoneNumberParser.cs
sherryzyy: sure, I'll

PR #1489, opened 2019-03-10
Support Sequence Recognizer for Email in Python

Grey0202:


Python/libraries/recognizers-sequence/recognizers_sequence/sequence/extractors.py
```diff
@@ -114,3 +115,19 @@ def __init__(self):
...
+        return 'email'
+
+    @property
+    def regexes(self) -> List[ReVal]:
+        return self._regexes
+
+    def __init__(self):
+        self._regexes = [
+            ReVal(RegExpUtility.get_safe_reg_exp(BaseEmail.EmailRegex), Constants.EMAIL_REGEX),
+            #EmailRegex2 will break the code as it's not supported in Python, comment out for now
```
tellarin: What is not supported in it?

Python/samples/simple_console/sample.py
```diff
@@ -70,7 +70,11 @@ def parse_all(user_input: str, culture: str) -> List[List[ModelResult]]:
 
         # PhoneNumber recognizer will find any phone number presented
         # E.g "My phone number is ( 19 ) 38294427."
-        Recognizers.recognize_phone_number(user_input, culture)
+        Recognizers.recognize_phone_number(user_input, culture),
+
+        # PhoneNumber recognizer will find any phone number presented
```
tellarin: Can you please update the comment?

Python/tests/test_runner_sequence.py
```diff
@@ -1,12 +1,12 @@
...
 from runner import get_specs
-from recognizers_sequence.sequence.sequence_recognizer import recognize_phone_number
+from recognizers_sequence.sequence.sequence_recognizer import recognize_phone_number, recognize_email
 
 MODEL_FUNCTION = {
     'PhoneNumber': recognize_phone_number,
+    'Email':recognize_email,
 }
 
-
```
tellarin: Nit: No need to remove this, right?

Python/libraries/recognizers-sequence/recognizers_sequence/sequence/extractors.py
Grey0202: The EmailPrefix & EmailSuffix in base_emial.py have a little bug for Regex or Re in Python do not have (?(condition) YES|No) but should use (?(?=condition) YES|No). Here is the code which I tested and passed.
`EmailPrefix = f'(?(?=\"\")(\"\".+?(?<!\\\\)\"\")|(([0-9A-Za-z]((\\.(?!\\.))|[-!#\\$%&\'\\*\\+/=\\?\\^\\{{\\}}\\|~\\w])*)(?<=[0-9A-Za-z])))'`
`EmailSuffix = f'(?(?=\\[)(\\[{IPv4Regex}\\])|{NormalSuffixRegex})'`
But it seems like to change the Base-Email.yaml in Patterns to fixed this bug. C#/Java/Python both have the "?=" and JS are not supported so I think it will work but my disk is not enough for building the entire project. Sorry for it. : (

Python/samples/simple_console/sample.py
Grey0202: Updated

Python/libraries/recognizers-sequence/recognizers_sequence/sequence/extractors.py
tellarin: Got it. Thanks.

Issue #1473, opened 2019-02-26
[Samples] Update BotBuilderRecognizerBot to BotBuilder V4


iMicknl:
**Is your feature request related to a problem? Please describe.**
Currently the BotBuilder sample is using the BotBuilder version 3 SDK and not the newest v4 SDK. Many thing have been changed and creating bots using the V4 version is recommended. 

**Describe the solution you'd like**
Refactor the BotBuilder sample to V4

**Describe alternatives you've considered**
Remove the sample and point to one of the prompt samples in [BotBuilder-samples](https://github.com/Microsoft/Botbuilder-samples) or point to the [BotBuilderCommunity prompts](https://github.com/BotBuilderCommunity/botbuilder-community-dotnet/tree/master/libraries/Bot.Builder.Community.Dialogs.Prompts) which use Recognizers-Text.

**Additional context**
BotBuilder V3 also depends on the Chronic library, where V4 just uses the Recognizers-Text library.

tellarin:
@GasparAcevedoZainSouthworks Can you folks take this and make sure all works correctly?

GasparAcevedoZainSouthworks:
OK @tellarin, we will take this Issue.

iMicknl:
@tellarin is it still necessary to have a BotBuilder example in this repository? Aren't the default prompts making use of `Recognizers-Text`? And the `BotBuilder-Community` packages are providing the missing prompts, using `Recognizers-Text`.

GasparAcevedoZainSouthworks:
Hi @iMicknl, @Tellarin
We created the draft PR #1491  which uses BotBuilder v4 and Recognizers in order to update BotBuilderRecognizerBot sample. We created this as a draft because, as @iMicknl said, the Botbuilder-Dotnet prompts use [recognizers-text](https://github.com/Microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs/Prompts/DatetimePrompt.cs#L58), so, in order to have a better sample, we will be refactoring it.

ceciliaavila:
Hi @iMicknl, @tellarin, 
As you mentioned before, BotBuilder default prompts are making use of Recognizers-Text, so we updated the sample in [PR #1491](https://github.com/Microsoft/Recognizers-Text/pull/1491) to use prompts instead of Recognizers.
As this sample no longer includes Recognizers, we consider that it should be moved to BotBuilder-Samples repository.
[...]

PR #1491, opened 2019-03-12
[DotNet][Samples] Update BotBuilderRecognizerBot sample to BotBuilder v4

ceciliaavila:
### Description
This is a sample that uses **BotBuilder v4** and **Recognizers** in order to update _BotBuilderRecognizerBot_ sample. 
This sample demonstrates how the Recognizers can be used to parse user input. The bot provides a basic experience for ordering roses, it starts by asking the amount of roses and then asks for a delivery date.

This PR is related to ISSUE #1473

### Changes Made
- Update BotBuilderRecognizerBot sample to BotBuilder v4.


ceciliaavila:
@cleemullins, @johnataylor, what do you think of this sample?

tellarin:
One of the reasons for the original sample was to try to make sure changes on the recognizers side wouldn't break usage in BotFramework. Would it make sense to have a sample here for a prompt, that's used by BF? Probably quite redundant with the real BF code, but may help catch breaking API changes.

ceciliaavila:
Hi @tellarin, 
Perfect, we will leave the sample in this repository to help catch breaking API changes. 
Thanks!

> @ceciliaavila One of the reasons for the original sample was to try to make sure changes on the recognizers side wouldn't break usage in BotFramework. Would it make sense to have a sample here for a prompt, that's used by BF? Probably quite redundant with the real BF code, but may help catch breaking API changes.

tellarin:
@ceciliaavila I took a quick look at the code, but it's not clear to me how this will catch breaking changes.
Shouldn't this sample define it's own prompt using the recognizers code directly?

ceciliaavila:
> @ceciliaavila I took a quick look at the code, but it's not clear to me how this will catch breaking changes.
> Shouldn't this sample define it's own prompt using the recognizers code directly?

Hi @tellarin,
You are correct, in order for this sample to detect any breaking changes it should use Recognizers directly. 
We updated the code to use Recognizers for validating user's input as you can see [here](https://github.com/Microsoft/Recognizers-Text/pull/1491/files?file-filters%5B%5D=.asax&file-filters%5B%5D=.config&file-filters%5B%5D=.cs&file-filters%5B%5D=.htm&file-filters%5B%5D=.json&file-filters%5B%5D=.md#diff-03b5edc86e7deca13c141348f3977719R180).

Please, let us know if there is anything else we need to improve.
 
Thanks!

Issue #1494, opened 2019-03-17
[* DateTime] Inconsistent behaviors among platforms affect order of Resolution values


STayinloves:
In .NET platform, the order of the elements in `Resolution.values` relied on the order of C# `Dictionary` which is non-deterministic. So its counterpart in other platforms may produce elements in a different order. But the testing code does examine the order of elements in `Resolution.values`.

So two solution may be effective:
- making the order of elements in `Resolution.values` reliable
- let the testing code not check the order (if it isn't necessary)

tellarin:
@Aliandi could you folks help with this? This is a Python issue where the current test code will fail as the order of resolutions may change. We can fix the test code to not rely on the order.
This would allow us to re-enable many Python tests and make sure features work correctly there. Currently many spec cases have their tests disabled due to this.

Aliandi:
Hey @tellarin! Sure we will look into this and let you know. We will be testing using [this](https://github.com/microsoft/Recognizers-Text/blob/a313d3331a4787ccf08093ba64ed6f4bb368dde1/Specs/DateTime/English/DateTimeModel.json#L10848) test, mentioned in the related PR #1543.

PR #1495, opened 2019-03-17
 Revise French DateTimeExtractor

STayinloves:
For cases like "Je voudrais reserver une salle le vendredi 30 novembre 2019 entre 12h00 et 14h00" and "Ya t'il une salle de dispo à nantes le lundi 24 décembre à 10h, pour 2h".

- revise mergedDateTimeExtractor to make its behavior consistent among all platforms
- the former case is only supported in .NET due to other platforms' lagging behind
- the latter case isn't supported in .NET due to issue #1494


.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -1,785 +1,785 @@
...
-			@"année à ce jour",
-			@"an à ce jour"
-		};
-	}
+﻿//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a tool.
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+//
```
tellarin: Nit

tellarin: BTW, do we really need to re-change all line ends?

PR #1529, opened 2019-04-16
Fix datetime recognition bug when <date> and <year> being separated by <time>

### LINKED ISSUE #1430: [EN DateTimeV2] Year incorrect when long date format used: Wed Oct 26 15:50:06 2016
**Describe the bug**
When trying to recognize Wed Oct 26 15:50:06 2016, the year is reported as 2018 instead of 2016.

**To Reproduce**
1. DateTimeRecognizer.RecognizeDateTime("Wed Oct 26 15:50:06 2016", "en");

In VS immediate window, do:
?(DateTimeRecognizer.RecognizeDateTime("Wed Oct 26 15:50:06 2016", "en")[0].Resolution["values"]) as List<Dictionary<string,string>>)[0]["value"]

2. Returns 2018-10-26 15:50:06

**Expected behavior**
Returns 2016-10-26 15:50:06

**Sample input/output**
Converting the entire resolution structure to JSON:
{"Key":"values","Value":[{"timex":"XXXX-10-26T15:50:06","type":"datetime","value":"2018-10-26 15:50:06"},{"timex":"XXXX-10-26T15:50:06","type":"datetime","value":"2019-10-26 15:50:06"}]}

And a separate result for the date-range which interestingly mentions 2016:
{"Key":"values","Value":[{"timex":"2016","type":"daterange","start":"2016-01-01","end":"2017-01-01"}]}

 - Platform: C#
 - Environment: Nuget package used via console app
 - Version of package Microsoft.Recognizers.Text.DateTime (1.1.4)

**Additional context**
none

Grey0202:
Revise .NET to recognize the DateTime which year is in the after of time which fixes #1430.


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimeParser.cs
```diff
@@ -249,8 +249,25 @@ private DateTimeResolutionResult MergeDateAndTime(string text, DateObject refere
                 ret.Comment = Constants.Comment_AmPm;
             }
 
-            ret.FutureValue = DateObject.MinValue.SafeCreateFromValue(futureDate.Year, futureDate.Month, futureDate.Day, hour, min, sec);
-            ret.PastValue = DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, hour, min, sec);
+            // Handle case like "Wed Oct 26 15:50:06 2016" which year and month separated by time.
```
STayinloves: Can we split this part to a separate function? That will help readability.

Specs/DateTime/English/DateExtractor.json
```diff
@@ -1563,5 +1563,17 @@
     "Input": "the face amount of its 6 1/4% convertible...",
     "Comment": "1/4 shouldn't recognized as date here",
     "Results": []
+  },
+  {
+    "Input": "I'll go back twenty second of June 2017",
```
STayinloves: Better be "twenty-second".

tellarin: Both should work.

Specs/DateTime/English/DateTimeParser.json
```diff
@@ -1337,5 +1337,53 @@
...
+            "dateTime": "2016-10-26 15:50:06"
+          }
+        },
+        "Start": 18,
+        "Length": 24
+      }
+    ]
+  },
+  {
+    "Input": "Wed Oct 26 15:50:06 2016 is not a day in 2019.",
```
tellarin: Can you add this case to DateTimeModel.json and to ComplexCalendar too? We want to make sure the final result is also correct, not only the result for specific Extractor/Parser.

Grey0202: Added.

Specs/DateTime/English/DateExtractor.json
Grey0202: Added

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimeParser.cs
Grey0202: Changed.

STayinloves: Didn't see the change..

Grey0202: > Didn't see the change..

Do you mean I should write a function? I think this one can be here like the "Handle morning, afternoon".

STayinloves: Yes, since this function already contains a large body of code, we should make the function as neat as possible if we can.

Grey0202: I have to use up to 8 local variables here to change the value of ret. Are you sure about this?

STayinloves: Yes, it may need further refactor, but it's fine with this PR.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimeParser.cs
```diff
@@ -251,6 +251,21 @@ private DateTimeResolutionResult MergeDateAndTime(string text, DateObject refere
 
             ret.FutureValue = DateObject.MinValue.SafeCreateFromValue(futureDate.Year, futureDate.Month, futureDate.Day, hour, min, sec);
             ret.PastValue = DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, hour, min, sec);
+
+            // Handle case like "Wed Oct 26 15:50:06 2016" which year and month separated by time.
+            var matchYear = this.config.YearRegex.Match(text.Substring(er2[0].Start + er2[0].Length ?? 0));
+            if (matchYear.Success && pr1.TimexStr.Substring(0, 4) == "XXXX" && pr1.TimexStr.Substring(5, 2) != "XX")
```
tellarin: Can you add a code comment here explaining what this condition means?
Also, don't add hardcoded timex terms. Declare them in some of the existing Timex classes and use here. This makes it easier to maintain in the future, especially if timex expressions change.
The substring comparisons should also be methods in one of the existing timex classes (and substring is not the best way to do this. If XXXX changes to something else, it will break this check, for example.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimeParser.cs
```diff
@@ -251,6 +251,21 @@ private DateTimeResolutionResult MergeDateAndTime(string text, DateObject refere
 
             ret.FutureValue = DateObject.MinValue.SafeCreateFromValue(futureDate.Year, futureDate.Month, futureDate.Day, hour, min, sec);
             ret.PastValue = DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, hour, min, sec);
+
+            // Handle case like "Wed Oct 26 15:50:06 2016" which year and month separated by time.
+            var matchYear = this.config.YearRegex.Match(text.Substring(er2[0].Start + er2[0].Length ?? 0));
```
tellarin: Please declare a var to receive text.Substring(er2[0].Start + er2[0].Length ?? 0) and give a name that describes what this represents.

tellarin:
@Grey0202 Please avoid force pushing to a PR branch once reviews are in. It makes the review process much more confusing. Unless any 'discarded" or "modified" part of the commit history is clearly after the latest PR review/comments.

tellarin:
@Grey0202 The build is failing in JavaScript. Can you take a look?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimeParser.cs
```diff
@@ -253,11 +253,14 @@ private DateTimeResolutionResult MergeDateAndTime(string text, DateObject refere
             ret.PastValue = DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, hour, min, sec);
 
             // Handle case like "Wed Oct 26 15:50:06 2016" which year and month separated by time.
-            var matchYear = this.config.YearRegex.Match(text.Substring(er2[0].Start + er2[0].Length ?? 0));
-            if (matchYear.Success && pr1.TimexStr.Substring(0, 4) == "XXXX" && pr1.TimexStr.Substring(5, 2) != "XX")
+            var timeSuffix = text.Substring(er2[0].Start + er2[0].Length ?? 0);
```
tellarin: Can we do some benchmarking to see the perf impact of this change? @STayinloves can maybe guide you.

tellarin: @Grey0202 unfortunately there are now also some conflicts to resolve.

Grey0202: > Can we do some benchmarking to see the perf impact of this change? @STayinloves can maybe guide you.

I test this many times and the current version takes about 44270ms and PR version takes about 44314ms for all 10314 test on average (including copy the test for 10 times and run the test for more than 10 times they all get similar results). The average length of the sentence is 31.

Issue #1415, opened 2019-01-25
[EN Ordinal] In cases like "last two", "first three" shouldn't be recognized as "Ordinal.Relative"


Irrelevances:
In cases like "last two", "first three", "first", "last" shouldn't be recognized as "Ordinal".

tellarin:
Covered in #1531.

PR #1531, opened 2019-04-16
Fix the "ordinal.relative" entity issue and add new attributes (#1415)

sherryzyy:
1. Fix the issue: cases like "last two", "first three" shouldn't be recognized as "Ordinal.Relative"
input text: last two, next 3, first two
expected output: []
2. Add new type "ordinal.relative" based on type "ordinal"
3. New attributes response
- add new attributes "offset" and "relativeTo"; 
- remove attribute "value" in "ordinal.relative" entity
- remove flag check (enablePreview)
- update the Specs

Not supported in Java, JavaScript and Python.


.NET/Microsoft.Recognizers.Text.Number/Constants.cs
```diff
@@ -47,11 +47,15 @@ public static class Constants
         public const string POWER_SUFFIX = "Pow";
         public const string SPECIAL_SUFFIX = "Spe";
 
+        // RelativeTo
+        public const string RELATIVE_START = "start";
```
tellarin: Why only one relative constant? There can be three of them, right?

.NET/Microsoft.Recognizers.Text.Number/Constants.cs
```diff
@@ -47,11 +47,15 @@ public static class Constants
...
 
+        // RelativeTo
+        public const string RELATIVE_START = "start";
+
         // Number subtypes
         public const string INTEGER = "integer";
         public const string DECIMAL = "decimal";
         public const string FRACTION = "fraction";
         public const string POWER = "power";
+        public const string RELATIVE = "relative";
```
tellarin: Add a comment like "// Odinal subtypes" before this one.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -279,16 +280,23 @@ protected ParseResult TextNumberParse(ExtractResult extResult)
...
+                Metadata = extResult.Metadata,
             };
 
             var handle = extResult.Text.ToLower();
 
             handle = Config.HalfADozenRegex.Replace(handle, Config.HalfADozenText);
 
             // Handling cases like "last", "next one", "previous one"
-            if (Config.RelativeReferenceMap.ContainsKey(extResult.Text))
+            // if (Config.RelativeReferenceMap.ContainsKey(extResult.Text))
```
tellarin: We should not leave commented out code.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -328,6 +336,9 @@ protected ParseResult TextNumberParse(ExtractResult extResult)
                 }
 
                 result.Value = intPartRet + pointPartRet;
+
+                // Get RelativeTo for Ordinal
```
tellarin: Nit: Change the comment to something like: "// Every ordinal number is relative to the start."

.NET/Microsoft.Recognizers.Text/Parsers/IParser.cs
```diff
@@ -30,5 +30,9 @@ public ParseResult(ExtractResult er)
         // Output the value in string format.
         // It is used in some parsers.
         public string ResolutionStr { get; set; } = string.Empty;
+
+        public string Offset { get; set; } = null;
```
tellarin: If we change this here, are you sure they don't appear in other output?
This is specific to Ordinal numbers. Not sure if it should be in all ParseResults. Most entities won't have them.

Specs/Number/English/OrdinalModel.json
```diff
@@ -3,293 +3,596 @@
...
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "11th",
+        "Start": 0,
+        "End": 3,
         "TypeName": "ordinal",
         "Resolution": {
+          "offset": "11",
+          "relativeTo": "",
```
tellarin: This should be start, right?

Specs/Number/English/OrdinalModel.json
```diff
@@ -3,293 +3,596 @@
...
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "21st",
+        "Start": 0,
+        "End": 3,
         "TypeName": "ordinal",
         "Resolution": {
+          "offset": "21",
+          "relativeTo": "",
```
tellarin: Same. And so on.

.NET/Microsoft.Recognizers.Text/Parsers/IParser.cs
sherryzyy: "current" and "end" is in the RelativeReferenceRelativeToMap.
Ordinal and Ordinal.relative have different implementation in this case.

.NET/Microsoft.Recognizers.Text.Number/Constants.cs
sherryzyy: "current" and "end" is in the RelativeReferenceRelativeToMap.
Ordinal and Ordinal.relative have different implementation in this case.

tellarin: I understand, but I think we should define the constants so it's clear what the potential values are.

PR #1533, opened 2019-04-19
Fix "9.am" being treated as ambiguous URL

STayinloves:
For cases like "9.am", we treat it as DateTime string but ambiguous URL, unless it has a specific prefix like "http://9.am".


JavaScript/packages/recognizers-sequence/src/sequence/extractors.ts
```diff
@@ -214,13 +220,22 @@ export class BaseEmailExtractor extends BaseSequenceExtractor {
...
         super();
         this.regexes = new Map<RegExp, string>()
             .set(RegExpUtility.getSafeRegExp(BaseURL.UrlRegex), Constants.URL_REGEX)
             .set(RegExpUtility.getSafeRegExp(BaseURL.UrlRegex2), Constants.URL_REGEX)
             .set(RegExpUtility.getSafeRegExp(BaseURL.IpUrlRegex), Constants.URL_REGEX)
+        this.ambiguousTimeTerm = RegExpUtility.getSafeRegExp(BaseURL.AmbiguousTimeTerm);
+    }
+
+    isValidMatch(match: Match): Boolean {
+        if (RegExpUtility.isMatch(this.ambiguousTimeTerm, match.value)) {
```
tellarin: Nit: We could just directly return the negation of the condition. Code is a little cleaner that way.

JavaScript/packages/recognizers-sequence/src/sequence/extractors.ts
```diff
@@ -26,6 +26,8 @@ export abstract class BaseSequenceExtractor implements IExtractor {
         //Traverse every match results to see each position in the text is matched or not.
         var collections = this.regexes.forEach((typeExtracted, regex) => {
             RegExpUtility.getMatches(regex, source).forEach(match => {
+                if (!this.isValidMatch(match)) return;
```
tellarin: Please always add {} to improve readability.

Specs/Sequence/English/URLModel.json
```diff
@@ -394,5 +394,30 @@
...
+    "NotSupportedByDesign": "python",
+    "Results": []
+  },
+  {
+    "Input": "7.am is more likely a datetime string rather than a valid domain name.",
+    "NotSupportedByDesign": "python",
+    "Results": []
+  },
+  {
+    "Input": "Please visit https://7.am for more information.",
```
tellarin: What about something like "27.pm" or "s7.am"? Those could still be valid URLs. Does the current code account for it?

PR #1535, opened 2019-04-21
Correct wrongly unmerged datetime in Chinese (#1525)

### LINKED ISSUE #1525: [CN DateTimeV2] '2018年12月' wrong extraction
**Describe the bug**
When it parses '2018年10月'，there is two results with '2018' and '10月', rather than '2018年10月'. It works in '2018年大年' / '2018年十月' case.

**To Reproduce**
Parse '2018年10月'

**Expected behavior**
Same result as '2018年十月'

**Sample input/output**

 - Platform: [Python]
 - Environment: 
 - Version of package [latest]

**Additional context**

Grey0202:
Fixes #1525 and added test case.


STayinloves:
Hi @Grey0202, I do see some test cases didn't pass in Python's test, can you fix it at first, so that we can begin the reviewing process?

Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -108,10 +108,10 @@ RelativeMonthRegex: !nestedRegex
   def: (?<relmonth>({DatePeriodThisRegex}|{DatePeriodLastRegex}|{DatePeriodNextRegex})\s*月)
   references: [DatePeriodThisRegex, DatePeriodLastRegex, DatePeriodNextRegex]
 DatePeriodYearRegex: !nestedRegex
-  def: (({YearNumRegex})(\s*年)?|({YearRegex})\s*年)(?=[\u4E00-\u9FFF]|\s|$|\W)
+  def: ((({YearNumRegex})(\s*年)?|({YearRegex})\s*年))
```
STayinloves: Remove duplicated parentheses

Specs/DateTime/Chinese/DatePeriodExtractor.json
```diff
@@ -470,5 +470,38 @@
         "Length": 14
       }
     ]
-  }
+  },
+  {
+    "Input": "2018年10月",
```
STayinloves: Test specs better be complete sentences.

Patterns/Chinese/Chinese-DateTime.yaml
Grey0202: It is necessary otherwise the Python won't work.

Specs/DateTime/Chinese/DatePeriodExtractor.json
Grey0202: It is to test the "2018年10月" and without other parts.

Patterns/Chinese/Chinese-DateTime.yaml
Grey0202: Sorry for this.

PR #1536, opened 2019-04-22
[Time] or later" should be synonym of "since [Time]

sherryzyy:
the following queries should be recognized as the synonym of "since [Time]

- meet me at 3 pm or later

- meet me at 3 pm or later on Monday

Supported in .NET and JAVA
Not supported in JavaScript and Python


Patterns/English/English-DateTime.yaml
```diff
@@ -403,8 +403,11 @@ SpecificTimeFromTo: !nestedRegex
 SpecificTimeBetweenAnd: !nestedRegex
   def: (between\s+)(?<time1>(({TimeRegex2}|{FirstTimeRegexInTimeRange})|({HourRegex}|{PeriodHourNumRegex})(\s*(?<leftDesc>{DescRegex}))?))\s*{RangeConnectorRegex}\s*(?<time2>(({TimeRegex2}|{TimeRegexWithDotConnector}(?<rightDesc>\s*{DescRegex}))|({HourRegex}|{PeriodHourNumRegex})(\s*(?<rightDesc>{DescRegex}))?))
   references: [ TimeRegex2, FirstTimeRegexInTimeRange, TimeRegexWithDotConnector, RangeConnectorRegex, HourRegex, PeriodHourNumRegex, DescRegex, PmRegex, AmRegex ]
-PrepositionRegex: !simpleRegex
-  def: (?<prep>^(at|on|of)(\s+the)?$)
+DateAfterRegex: !simpleRegex
```
STayinloves: If we need to reuse this regex, we'd better change the name.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -173,7 +173,8 @@ public static class DateTimeDefinitions
...
 		public static readonly string SpecificTimeFromTo = $@"({RangePrefixRegex}\s+)?(?<time1>(({TimeRegex2}|{FirstTimeRegexInTimeRange})|({HourRegex}|{PeriodHourNumRegex})(\s*(?<leftDesc>{DescRegex}))?))\s*{TillRegex}\s*(?<time2>(({TimeRegex2}|{TimeRegexWithDotConnector}(?<rightDesc>\s*{DescRegex}))|({HourRegex}|{PeriodHourNumRegex})(\s*(?<rightDesc>{DescRegex}))?))";
 		public static readonly string SpecificTimeBetweenAnd = $@"(between\s+)(?<time1>(({TimeRegex2}|{FirstTimeRegexInTimeRange})|({HourRegex}|{PeriodHourNumRegex})(\s*(?<leftDesc>{DescRegex}))?))\s*{RangeConnectorRegex}\s*(?<time2>(({TimeRegex2}|{TimeRegexWithDotConnector}(?<rightDesc>\s*{DescRegex}))|({HourRegex}|{PeriodHourNumRegex})(\s*(?<rightDesc>{DescRegex}))?))";
-		public const string PrepositionRegex = @"(?<prep>^(at|on|of)(\s+the)?$)";
+		public const string DateAfterRegex = @"\b(((at)\s)?(or|and)\s+(above|after|later|greater)(?!\s+than))\b";
+		public static readonly string PrepositionRegex = $@"(?<prep>^({DateAfterRegex}\s+)?(at|on|of)(\s+the)?$)";
```
tellarin: I don't think this is the correct place to modify.
PrepositionRegex covers prepositions. And DateAfterRegex covers dates after. This change mixes them and breaks their intent.
"<T> or later" seems a special case of dates after. Maybe make the "than" optional or have another usage of a regex similar to DateAfterRegex to use only in the new case are better strategies.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -10570,6 +10570,62 @@
...
+              "type": "timerange",
+              "start": "15:00:00"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Meet me at 3 pm or later on Monday.",
```
tellarin: Could you add these cases to ComplexCalendar too?
The rest LGTM. Thanks!

PR #1538, opened 2019-04-23
[BG Choice] Support Bulgarian choice recognizer in .Net

mitkodi:
Adding Bulgarian (bg-BG) support for Choice recognizer.


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -13,9 +13,9 @@ public class BaseNumberParser : IParser
             new Regex(@"\d+", RegexOptions.Singleline);
 
         private static readonly Regex MultiDecimalSeparatorCultureRegex =
-            new Regex(@"^(en|es|fr)(-)?\b", RegexOptions.Singleline | RegexOptions.Compiled);
+            new Regex(@"^(en|es|fr|bg)(-)?\b", RegexOptions.Singleline | RegexOptions.Compiled);
```
tellarin: Is Bulgarian really multi-decimal separator? For example, are both 5.0 and 5,0 the same number?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -13,9 +13,9 @@ public class BaseNumberParser : IParser
             new Regex(@"\d+", RegexOptions.Singleline);
 
         private static readonly Regex MultiDecimalSeparatorCultureRegex =
-            new Regex(@"^(en|es|fr)(-)?\b", RegexOptions.Singleline | RegexOptions.Compiled);
+            new Regex(@"^(en|es|fr|bg)(-)?\b", RegexOptions.Singleline | RegexOptions.Compiled);
 
-        private static readonly List<string> CompoundNumberLanguages = new List<string> { "de-DE", "nl-NL", "sv-SE" };
+        private static readonly List<string> CompoundNumberLanguages = new List<string> { "de-DE", "nl-NL", "sv-SE", "bg-BG" };
```
tellarin: This represents the difference between English "twenty two" and German "zweiundzwanzig" in having space segmentation for numbers. Is Bulgarian like German? 
Sorry, I have no knowledge of Bulgarian. :)

Specs/Choice/Bulgarian/BooleanModel.json
```diff
@@ -0,0 +1,198 @@
+[
+  {
+    "Input": "Определено!",
+    "NotSupportedByDesign": "python",
```
tellarin: This PR only adds .NET support. 
Could you also add JavaScript? If not, please just add a NotSupported attribute for both JavaScript and Java.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
mitkodi: Thank you @tellarin,
Yes, now days "." and "," are interchangeable as a decimal separator. The formal rule is "," but a decimal point is increasingly used (influenced by electronic devices and internet).
In addition, in regular expressions I've also included variants of common misspelling in some words.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
mitkodi: For "twenty two" in Bulgarian is like "twenty and two", but "two hundred" (and it's true for all hundred numbers) in Bulgarian is like "twohundred" (двеста).

Specs/Choice/Bulgarian/BooleanModel.json
mitkodi: Do you mean to add NotSupported into Bulgarian\BooleanModel.json? (sorry if I haven't understood it right)

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
tellarin: Interesting. Later we'll probably need some customization to properly support that.

Specs/Choice/Bulgarian/BooleanModel.json
tellarin: Yes, basically add that attribute in each spec case in the json file. 
While Python is currently not supported by design and there is an attribute per test to show this  (there is no ChoiceRecognizer for Python yet), Javascript and Java have ChoiceRecognizer. 
So, until we have Bulgarian implementation, each test cases should be marked as NotSupported.

tellarin: Would you be able to contribute JavaScript implementation? If not, no worries. This is a separate question from the review.

PR #1539, opened 2019-04-24
Can't recognize modifier "starting|beginning" date

sherryzyy:
Can't recognize modifier "starting|beginning" date:
Query: Please can you arrange a 2hr Microsoft Teams Meeting **<starting|beginning (on)?>** January 7th to discuss ARM Templates?"
wrong answer: January 7th
right answer: starting on January 7th

Supported in .NET, Java, Javascript, and python.
not fix "from" because of many conflicts.


JavaScript/packages/recognizers-date-time/src/dateTime/baseMerged.ts
```diff
@@ -168,26 +168,33 @@ export class BaseMergedExtractor implements IDateTimeExtractor {
         let lastEnd = 0;
         ers.forEach(er => {
             let beforeStr = source.substr(lastEnd, er.start).toLowerCase();
+            let isSuccess = false;
```
tellarin: Can we give it a better name than "isSuccess". What does success here means?

JavaScript/packages/recognizers-date-time/src/dateTime/baseMerged.ts
```diff
@@ -168,26 +168,33 @@ export class BaseMergedExtractor implements IDateTimeExtractor {
...
                 er.text = source.substr(er.start, er.length);
+                isSuccess = true;
             }
-            let after = this.hasTokenIndex(beforeStr.trim(), this.config.afterRegex);
-            if (after.matched) {
-                let modLength = beforeStr.length - after.index;
-                er.length += modLength;
-                er.start -= modLength;
-                er.text = source.substr(er.start, er.length);
+            if(!isSuccess){
```
tellarin: Nit: Please add whitespaces to match the code style. Also, I suggest adding a blank line between two consecutive ifs.

JavaScript/packages/recognizers-date-time/src/dateTime/baseMerged.ts
tellarin: Not sure what the if/if logic here means. Can you describe in a code comment?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -10592,5 +10592,125 @@
         }
       }
     ]
+  },
+  {
+    "Input": "Please can you arrange a Microsoft Teams Meeting starting on January 7th to discuss ARM Templates?",
```
tellarin: Such cases already worked in .Net with just the regex change? Why do the other plats need the if logic change? To be in sync with .Net or are the implementations diverging?

sherryzyy: Yes, the logic in .NET and other platforms are different in the previous version.

PR #1540, opened 2019-04-25
Fix recognition of "between now and date" as daterange

songwenhao1:
Can't recognize "between now and date" daterange:
Query: between now and November 15th
wrong answer: two text: "now" "November 15th"
right answer: a daterange text: "between now and November 15th"
Supported in .NET,  java, python

 Javascript implementation will come in a future PR


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1257,67 +1262,131 @@ private DateTimeResolutionResult ParseYear(string text, DateObject referenceDate
...
+            ret.Timex = $"({pr1.TimexStr},{pr2.TimexStr},P{(futureEnd - futureBegin).TotalDays}D)";
+            ret.FutureValue = new Tuple<DateObject, DateObject>(futureBegin, futureEnd);
+            ret.PastValue = new Tuple<DateObject, DateObject>(pastBegin, pastEnd);
+            ret.Success = true;
+
+            return ret;
+        }
+
+        // Parse entities that are made up by two time points with now
+        private DateTimeResolutionResult MergeTwoTimePointsWithNow(string text, DateObject referenceDate)
```
tellarin: Why do we need two methods? Also, are the other changes just line ending changes? If so, please revert them to make review easier.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1257,67 +1262,131 @@ private DateTimeResolutionResult ParseYear(string text, DateObject referenceDate
...
         {
             var ret = new DateTimeResolutionResult();
-
+            var nowMatches = this.config.NowRegex.Matches(text);
```
tellarin: Didn't extract already tag "now"? Why do we need to re-extract in Parser?

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -4,6 +4,9 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 VisualStudioVersion = 15.0.27130.2026
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tests", "Tests", "{ED7B6456-AB0A-48CE-8F85-711FE87F09C2}"
+	ProjectSection(SolutionItems) = preProject
```
tellarin: Why is this change necessary?

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseDatePeriodExtractor.java
```diff
@@ -21,6 +21,8 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+
```
tellarin: Nit: Please, be careful to not add unnecessary blank lines like these.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseDatePeriodExtractor.java
```diff
@@ -95,6 +97,15 @@ public BaseDatePeriodExtractor(IDatePeriodExtractorConfiguration config) {
...
 
+        // Handle "now"
+        Match[] matches = RegExpUtility.getMatches(this.config.getNowRegex(), input);
+        if (matches.length != 0) {
+            for (Match match : matches) {
+                ers.add(new ExtractResult(match.index, match.length, match.value, Constants.SYS_DATETIME_DATE));
+            }
+
+            ers.sort(Comparator.comparingInt(arg -> arg.getStart()));
+        }
```
tellarin: Nit: Add a blank line between closing brace and the return.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseDateParser.java
```diff
@@ -177,8 +177,7 @@ private DateTimeResolutionResult parseImplicitDate(String text, LocalDateTime re
 
         if (exactMatch.getSuccess()) {
             int swift = getSwiftDay(exactMatch.getMatch().get().value);
-
-            LocalDateTime value = referenceDate.plusDays(swift);
+            LocalDateTime value = referenceDate.toLocalDate().atStartOfDay().plusDays(swift);
```
tellarin: Is this really the same behaviour as in .NET? Any usage of toLocal() always make me nervous of causing timezone issues.

Specs/DateTime/English/DatePeriodExtractor.json
```diff
@@ -3717,5 +3717,16 @@
         "Length": 17
       }
     ]
+  },
+  {
+    "Input": "I will do my work between now and November 15th",
```
tellarin: Can you also add a case for "between Jan 22 and now"?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
songwenhao1: In the first method, we will re_extract "date" from  "text", "now" is a Date time, so it would not be extracted, could not solve the issue

Specs/DateTime/English/DatePeriodParser.json
```diff
@@ -5595,5 +5595,31 @@
...
+          "Timex": "(2019-04-23,2019-11-15,P206D)",
+          "FutureResolution": {
+            "startDate": "2019-04-23",
+            "endDate": "2019-11-15"
+          },
+          "PastResolution": {
+            "startDate": "2019-04-23",
+            "endDate": "2019-11-15"
+          }
+
```
STayinloves: Remove this blank line

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseDateParser.java
songwenhao1: toLocalDate() just return date information.

public LocalDate toLocalDate() {
    return date;
} 
i find this also used in resolve End Of Day

    private DateTimeResolutionResult resolveEndOfDay(String timexPrefix, LocalDateTime futureDate, LocalDateTime pastDate) {
        String timex = String.format("%sT23:59:59", timexPrefix);
        LocalDateTime futureValue = LocalDateTime.of(futureDate.toLocalDate(), LocalTime.MIDNIGHT).plusDays(1).minusSeconds(1);
        LocalDateTime pastValue = LocalDateTime.of(pastDate.toLocalDate(), LocalTime.MIDNIGHT).plusDays(1).minusSeconds(1);

        DateTimeResolutionResult result = new DateTimeResolutionResult();
        result.setTimex(timex);
        result.setFutureValue(futureValue);
        result.setPastValue(pastValue);
        result.setSuccess(true);

        return result;
    }

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
tellarin: There is also a change in the extractor extracting "now" in this PR, no? Why do we need it again parsing?

tellarin: And it could be part of the same parsing method as a special case, couldn't it?

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseDateParser.java
tellarin: OK.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseDatePeriodParser.java
```diff
@@ -939,42 +940,57 @@ private DateTimeResolutionResult mergeTwoTimePoints(String text, LocalDateTime r
...
+            }
 
-        DateTimeParseResult pr1 = this.config.getDateParser().parse(er.get(0), referenceDate);
-        DateTimeParseResult pr2 = this.config.getDateParser().parse(er.get(1), referenceDate);
+            DateContext dateContext = getYearContext(er.get(0).getText(), er.get(1).getText(), text);
 
-        if (pr1.getValue() == null || pr2.getValue() == null) {
-            return ret;
-        }
+            pr1 = this.config.getDateParser().parse(er.get(0), referenceDate);
```
STayinloves: We should only `getDateParser` once to minimize time&memory cost.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/parsers/BaseDateParser.java
```diff
@@ -177,8 +177,8 @@ private DateTimeResolutionResult parseImplicitDate(String text, LocalDateTime re
 
         if (exactMatch.getSuccess()) {
             int swift = getSwiftDay(exactMatch.getMatch().get().value);
-
-            LocalDateTime value = referenceDate.plusDays(swift);
+
```
STayinloves: Remove trailing spaces.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1325,6 +1345,39 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             return ret;
         }
 
+        // Parse entities that are made up by two time points with now
+        private List<DateTimeParseResult> GetPrWithNow(MatchCollection matches, ExtractResult er, DateObject referenceDate)
+        {
+            var pr = new List<DateTimeParseResult>();
+            pr.Add(this.config.DateParser.Parse(er, referenceDate));
```
tellarin: This method should be only for the "now" code. The other entity (the date result) should not be extracted here.
Also, why call Parse() and not Extract like in the original code?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1262,43 +1262,63 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             var ret = new DateTimeResolutionResult();
 
             var er = this.config.DateExtractor.Extract(text, referenceDate);
-            if (er.Count < 2)
-            {
-                er = this.config.DateExtractor.Extract(this.config.TokenBeforeDate + text, referenceDate);
-                if (er.Count < 2)
-                {
-                    return ret;
```
tellarin: Shouldn't the attempt to extract "now" be here, and if nothing is extracted, then return ret?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1262,43 +1262,63 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
...
-            var match = this.config.WeekWithWeekDayRangeRegex.Match(text);
-            string weekPrefix = null;
-            if (match.Success)
-            {
-                weekPrefix = match.Groups["week"].ToString();
+            var pr1 = new DateTimeParseResult();
+            var pr2 = new DateTimeParseResult();
+
+            // Handle with "now"
+            var nowMatches = this.config.NowRegex.Matches(text);
```
tellarin: We shouldn't have to even do this, if the previous Extract already returned two entities.

tellarin: And everything related to "now", including this regex match should go into the method.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1325,6 +1345,39 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             return ret;
         }
 
+        // Parse entities that are made up by two time points with now
+        private List<DateTimeParseResult> GetPrWithNow(MatchCollection matches, ExtractResult er, DateObject referenceDate)
```
tellarin: Rename to ExtractNowAsDate or ParseNowAsDate.

tellarin: I'd prefer Extract, and Parse later, as needed. Similar to the behaviour in the original method.

songwenhao1: Ok, i will rename the function
 In the Parse stage of the  original method, DateParser couldn't parse the "now" for now is a datetime value, so return a pr about now maybe more simple

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -11060,5 +11060,55 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I will do my work between now and November 15th",
```
tellarin: Please add these same cases in DateTimeComplexCalendar too.

tellarin: Can you also add a case here and in ComplexCalendar like "Let's meet now". 
It should be recognized as a DateTime, not as a Date.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1262,42 +1262,57 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             var ret = new DateTimeResolutionResult();
 
             var er = this.config.DateExtractor.Extract(text, referenceDate);
+            var pr1 = new DateTimeParseResult();
```
tellarin: Why do you need to create two empty objects? Isn't it enough to declare the variable type?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1262,42 +1262,57 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
...
-                er[1].Start -= this.config.TokenBeforeDate.Length;
             }
 
-            var match = this.config.WeekWithWeekDayRangeRegex.Match(text);
-            string weekPrefix = null;
-            if (match.Success)
-            {
-                weekPrefix = match.Groups["week"].ToString();
-            }
+            if (er.Count >= 2)
```
tellarin: Why do we need this if now?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1325,6 +1340,36 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             return ret;
         }
 
+        // Parse entities that are made up by two time points with now
+        private DateTimeParseResult ParseNowAsDate(string text, DateObject referenceDate)
```
tellarin: What happens if there are two "now" in the sentence?
For example: "When I say now, I mean now!". Can you also add that as a spec case?

Specs/DateTime/English/DatePeriodExtractor.json
```diff
@@ -3717,5 +3717,29 @@
...
+    "NotSupported": "python, javascript",
```
tellarin: If you added a Python implementation, why are these cases marked as not supported in it?

Specs/DateTime/English/DatePeriodExtractor.json
```diff
@@ -3717,5 +3717,29 @@
...
+        "Text": "between now and november 15th",
+        "Type": "daterange",
+        "Start": 18,
+        "Length": 29
+      }
+    ]
+  },
+  {
+    "Input": "I have finished my work between Jan 22 and now",
+    "NotSupported": "python, javascript",
```
tellarin: And what's the key reason to not support in JS?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
songwenhao1: We will match "between..and.." in stage Extract, so if just have two "now" and not have "between/from" , the text would not be extracted.
I will add the example as a spec case.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
songwenhao1: In the former case ”er.count<2“, the value of er.count maybe changed in

er = this.config.DateExtractor.Extract(this.config.TokenBeforeDate + text, referenceDate);

so, i  judge the value again. if the value>=2, we should parse the er to pr, otherwise not needed

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
tellarin: But the method usage is not clear. Even the code description.
What about a phrase like "Let's meet between now and Friday; not right now, ok?"?
Please also add it in specs.

Issue #1173, opened 2019-01-07
[EN NumberRange] Cannot recognize "more than half"


HaoYangMS:
When trying to extract "more than half people came here", the Recognizer returns "I found no entities", but it can recognize "more than a half"

![image](https://user-images.githubusercontent.com/45840378/50750751-594fe480-1281-11e9-9aca-fbde671d36b6.png)

PR #1546, opened 2019-04-29
[EN NumberRange] Cannot recognize "more than half" (#1173)

sherryzyy:
Query: more than half people came here
expected result: "half" by itself could be tagged and resolved to 0.5


JavaScript/packages/recognizers-number/src/number/parsers.ts
```diff
@@ -210,7 +210,12 @@ export class BaseNumberParser implements IParser {
             let splitIndex = fracWords.length - 1;
             let currentValue = this.config.resolveCompositeNumber(fracWords[splitIndex]);
             let roundValue = 1;
-
+            
+            if (fracWords.length == 1){
```
tellarin: Nit: code standard.

Issue #1555, opened 2019-05-06
[DE Ordinal] add "letzter", "vorletzter" and similar ordinals from the back


Ablesius:
**Is your feature request related to a problem? Please describe.**
I would like to be able to access ordinals as counted from the back of a list, words like "letzter" ("last") and "vorletzter" ("second to last").

**Describe the solution you'd like**
Please provide me with the necessary information to implement the feature myself; so far I haven't found anything that seems related to that. I've been pointed to https://github.com/Microsoft/Recognizers-Text/blob/master/Patterns/German/German-Numbers.yaml, but I'm not sure what to add there:
* is there a list of allowed keys like "ZeroToNineIntegerRegex" and the other ones? 
* Or are the names completely arbitrary? 
* Where would I have to add them beside the German-Numbers.yaml above?

Any help is going to be appreciated :) cheers
Alex

tellarin:
@Ablesius To implement this you can follow the framework from English that already has this support.
Basically, changes from English (new regexes in YAML or extractor/parser config tweaks) need to be ported/tested in German.
I'll try to track the relevant commits and mention them in this issue.

Ablesius:
@tellarin I've opened a pull request, but the tests fail. If you have the time, maybe you could help me out there: https://github.com/microsoft/Recognizers-Text/pull/1557 :) cheers

tellarin:
@Ablesius, sorry for the delay. I've reviewed the PR and left comments there. Basically the tests need to be marked as "NotSupported" for now. Until the feature is fully implemented in German.

Ablesius:
@tellarin thanks for the pointers and the support :)

PR #1557, opened 2019-05-07
Add German Ordinal.relative spec examples (#1555)

Ablesius:
I'm not sure this is all there needs to be done; please inform me if I have to add more than this :)

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumber_German.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumber_German.cs
@@ -35,6 +35,13 @@ namespace Microsoft.Recognizers.Text.Number.Tests
             TestNumber();
         }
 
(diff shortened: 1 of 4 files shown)


Ablesius:
I can't authorise at CI of Microsoft.Recognizers-BotSys, so no idea why that crashes. The error code in appveyor I don't understand - it's about an OrdinalModel-German.csv, which doesn't seem to be part of the project. Is it auto-generated from the OrdinalModel.json?

tellarin:
BTW, the AppVeyor build should be the primary one to check in case of failures. The other one only builds .NET and run some additional internal checks that are not yet open.

Ablesius:
Thanks for your comments! I was off for a few days, but I'm continuing on this today.

Ablesius:
I tried to make the adaptations in accordance to PR #1550, but I'm not sure whether _all_ ordinal tests from OrdinalModel.json need to be moved to the other test (OrdinalModelEnablePreview.json, exactly as in the referenced PR), or only the ones I added. @tellarin if you could take another look, it would be awesome :)

tellarin:
@Ablesius your change looks good. However, the linter is complaining of mixed use of whitespace and tabs in one of the C# classes. Can you convert all to whitespaces and follow the indentation style of the rest of the file?

Issue #1556, opened 2019-05-06
[CN Number] Incorrect resolution for ''recognizers_number'' with ''Culture.Chinese''


spongejj:
**Describe the bug**
When I used the recognizers_number function with some Chinese, I got some results confused me.
**To Reproduce**
My code:
```python
from recognizers_number import recognize_number
result1 = recognize_number("一万零一十二", Culture.Chinese)
for res in result1:
    print("value: ", res.resolution['value'])
    print("text: ", res.text)

result2 = recognize_number("一万零十二", Culture.Chinese)
for res in result2:
    print("value: ", res.resolution['value'])
    print("text: ", res.text)
```
**Expected behavior**
"一万零一十二" is the same with "一万零十二" in Chinese, I think I should get the same value from recognize_number.

**Platform (please complete the following information):**
 - Platform: Mac
 - Environment: Python 3.6
 - Version of package 1.0.0a0

PR #1558, opened 2019-05-07
[CN Number] Incorrect resolution for ''recognizers_number'' with ''Culture.Chinese'' #1556

songwenhao1:
Issue #1556:
"一万零一十二" is the same with "一万零十二" in Chinese，should get the same value from recognize_number.


JavaScript/packages/recognizers-number/src/number/cjkParsers.ts
```diff
@@ -409,6 +409,8 @@ export class BaseCJKNumberParser extends BaseNumberParser {
                 if (index !== resultStr.length - 1) {
                     if ((currentChar === "零") && !this.config.roundNumberMapChar.has(resultStr.charAt(index + 1))) {
                         roundDefault = 1;
+                    } else if ((currentChar === "零") && (["十", "拾"].indexOf(resultStr.charAt(index + 1)) != -1)) {
```
wgx998877: please create a parameter for this list

Python/libraries/recognizers-number/recognizers_number/number/cjk_parsers.py
```diff
@@ -394,6 +394,8 @@ def get_int_value(self, source: str) -> float:
                 if i != len(result_str)-1:
                     if c == '零' and result_str[i+1] not in self.config.round_number_map_char:
                         round_default = 1
+                    elif c == '零' and result_str[i+1] in ['十', '拾']:
```
wgx998877: same

Specs/Number/Chinese/NumberModel.json
```diff
@@ -4230,5 +4230,19 @@
         }
       }
     ]
+  },
+  {
+    "Input": "一万零十二",
```
wgx998877: please add more case in this test file

Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/parsers/BaseCJKNumberParser.java
```diff
@@ -408,7 +408,7 @@ private double getIntValue(String intStr) {
                 roundDefault = roundRecent / 10;
             } else if (cjkConfig.getZeroToNineMap().containsKey(intStr.charAt(i))) {
                 if (i != intStr.length() - 1) {
-                    if (intStr.charAt(i) == '零' && !roundNumberMapChar.containsKey(intStr.charAt(i + 1))) {
+                    if (intStr.charAt(i) == '零') {
```
wgx998877: can you create a variable for '零' like chars? how about '〇' in Chinese?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -548,7 +550,7 @@ private double GetIntValue(string intStr)
                 {
                     if (i != intStr.Length - 1)
                     {
-                        if (intStr[i] == '零' && !Config.RoundNumberMapChar.ContainsKey(intStr[i + 1]))
+                        if (intStr[i] == roundNumberZero && (!Config.RoundNumberMapChar.ContainsKey(intStr[i + 1]) || (Config.CultureInfo.Name == "zh-CN" && System.Array.IndexOf(roundNumberTen, intStr[i + 1]) != -1)))
```
tellarin: This if could be a call to a method.
Also, it should not reference a culture directly. Whatever culture-specific behaviour is needed should go into the Config class for that culture. Either as attributes or methods.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -476,6 +476,8 @@ private double GetIntValue(string intStr)
 
             var isDozen = false;
             var isPair = false;
+            char roundNumberZero = '零';
```
tellarin: These values should not be hard coded here as the base parser is used for CJK. Any declarations like these two should be made in the YAML file and referenced in the parser config for each culture.

Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/parsers/BaseCJKNumberParser.java
```diff
@@ -361,6 +362,8 @@ private double getIntValue(String intStr) {
 
         boolean isDozen = false;
         boolean isPair = false;
+        char roundNumberZero = '零';
```
tellarin: Same as in .NET.

Specs/Number/Chinese/NumberModel.json
```diff
@@ -4230,5 +4230,47 @@
...
```
tellarin: Are these three new spec cases all situations? Isn't there any false positive we want to make sure doesn't happen?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -477,7 +477,6 @@ private double GetIntValue(string intStr)
             var isDozen = false;
             var isPair = false;
             char roundNumberZero = '零';
```
tellarin: This is still Chinese-specific, right? So it should move to a declaration in the right YAML and referenced in the right culture config.

.NET/Microsoft.Recognizers.Definitions.Common/Korean/NumbersDefinitions.cs
```diff
@@ -112,6 +113,11 @@ public static class NumbersDefinitions
 			'빵',
 			'열'
 		};
+		public static readonly IList<char> TenDirectList = new List<char>
```
wgx998877: please make sure this is correct in Korean and add more examples to test

Specs/Number/Chinese/NumberModel.json
wgx998877: Yes, please add more real examples which are more complex and with more different entities.
Also just like @tellarin's comment, please add more false positive cases.

Specs/Number/Chinese/NumberModel.json
```diff
@@ -4230,5 +4230,191 @@
...
+        "End": 8,
+        "TypeName": "number",
+        "Resolution": {
+          "value": "1010.33333333333"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "零十",
```
wgx998877: create another task for this like case.

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/NumbersDefinitions.cs
```diff
@@ -151,6 +152,11 @@ public static class NumbersDefinitions
 			'兆',
 			'億'
 		};
+		public static readonly IList<char> TenDirectList = new List<char>
```
tellarin: If 零 is ZeroChar, shouldn't  '十', '拾' be named TenChars?

.NET/Microsoft.Recognizers.Definitions.Common/Korean/NumbersDefinitions.cs
```diff
@@ -112,6 +113,10 @@ public static class NumbersDefinitions
 			'빵',
 			'열'
 		};
+		public static readonly IList<char> TenDirectList = new List<char>
+		{
+			' '
```
tellarin: Why is this not an empty list? Won't it cause problems matching to whitespace in Korean?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -132,7 +132,7 @@ protected ParseResult ParseFraction(ExtractResult extResult)
             }
             else
             {
-                intPart = "零";
+                intPart = Config.ZeroChar.ToString();
```
tellarin: If you need to convert to String, why not declare as string in the YAML or in the Config object?
Otherwise you're doing unnecessary conversions all the time.

Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/parsers/BaseCJKNumberParser.java
```diff
@@ -172,7 +172,7 @@ protected ParseResult parsePercentage(ExtractResult extResult) {
 
                     if (intNumberChar == '対' || intNumberChar == '对') {
```
tellarin: I know it's slightly out of scope for this exact PR, but can you move this "FiveChars" also to YAML resources?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
songwenhao1: Yes, i need to convert to String here, because "零" should be a String here, but in the latter when we  judge " intStr[i] == '零' "，零 should be a char

Issue #1545, opened 2019-04-26
[EN Number|DateTimeV2] Bugs in Datetime and Ordinal extractor


sherryzyy:
1. "January first 2000" should be recognized as one date entity
2. cases like "previous 1", "first one" should be recognized as one entity. now it only recognizes previous and first

tellarin:
Addressed in #1636.

PR #1559, opened 2019-05-08
[EN Number|DateTimeV2] Bugs in Datetime and Ordinal extractor #1545

songwenhao1:
1、"January first 2000" should be recognized as one date entity
2、cases like "previous 1" should be recognized as one entity. now it only recognizes previous.


.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -152,6 +152,8 @@ public static class NumbersDefinitions
 		public static readonly Dictionary<string, long> OrdinalNumberMap = new Dictionary<string, long>
 		{
 			{ @"first", 1 },
+			{ @"first one", 1 },
```
tellarin: Can't we do some normalization before checking this map? Instead of adding a lot of redundant values.

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -41,10 +41,10 @@ public static class NumbersDefinitions
 		public static readonly string AllIntRegexWithLocks = $@"((?<=\b){AllIntRegex}(?=\b))";
 		public static readonly string AllIntRegexWithDozenSuffixLocks = $@"(?<=\b)(((half\s+)?a\s+dozen)|({AllIntRegex}\s+dozen(s)?))(?=\b)";
 		public const string RoundNumberOrdinalRegex = @"(hundredth|thousandth|millionth|billionth|trillionth)";
-		public const string NumberOrdinalRegex = @"(first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth|eleventh|twelfth|thirteenth|fourteenth|fifteenth|sixteenth|seventeenth|eighteenth|nineteenth|twentieth|thirtieth|fortieth|fiftieth|sixtieth|seventieth|eightieth|ninetieth)";
-		public const string RelativeOrdinalRegex = @"((next|previous) one|(the second|next) to last|the one before the last( one)?|the last but one|(ante)?penultimate|last|next|previous)";
+		public const string NumberOrdinalRegex = @"(first (one|1)|first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth|eleventh|twelfth|thirteenth|fourteenth|fifteenth|sixteenth|seventeenth|eighteenth|nineteenth|twentieth|thirtieth|fortieth|fiftieth|sixtieth|seventieth|eightieth|ninetieth)";
```
tellarin: Nit: "(one|1)" could be it's own sub-regex declared in the YAML file for reuse.

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -268,6 +271,7 @@ public static class NumbersDefinitions
 			{ @"last", @"end" },
 			{ @"next one", @"current" },
```
tellarin: Otherwise we also need to add for "next", "last", etc, right?

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -41,10 +41,10 @@ public static class NumbersDefinitions
...
-		public const string RelativeOrdinalRegex = @"((next|previous) one|(the second|next) to last|the one before the last( one)?|the last but one|(ante)?penultimate|last|next|previous)";
+		public const string NumberOrdinalRegex = @"(first (one|1)|first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth|eleventh|twelfth|thirteenth|fourteenth|fifteenth|sixteenth|seventeenth|eighteenth|nineteenth|twentieth|thirtieth|fortieth|fiftieth|sixtieth|seventieth|eightieth|ninetieth)";
+		public const string RelativeOrdinalRegex = @"((next|previous) (one|1)|(the second|next) to last|the one before the last( one)?|the last but one|(ante)?penultimate|last|next|previous)";
 		public static readonly string BasicOrdinalRegex = $@"({NumberOrdinalRegex}|{RelativeOrdinalRegex})";
-		public static readonly string RelativeOrdinalFilterRegex = $@"(?<!-)(first|{RelativeOrdinalRegex})\s*({TwoToNineIntegerRegex}|[2-9]+)(?!\s*{RoundNumberIntegerRegex})";
+		public static readonly string RelativeOrdinalFilterRegex = $@"(?<!-)(first|{RelativeOrdinalRegex})\s*({TwoToNineIntegerRegex}|[2-9]\b)(?!\s*{RoundNumberIntegerRegex})";
```
tellarin: Why exactly is this change needed?

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
tellarin: And this fix doesn't cover cases like "third one" and etc. If Ordinal will tag "first one" as it does for "last one", it should also work for other cases of "one|1".
Perhaps rewriting the regex to have an optional "(\s+(one|1))?" at the end.

PR #1563, opened 2019-05-10
JavaScript support for issue #1540 ("between now and date")

songwenhao1:
JavaScript support for #1540


Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -9322,7 +9320,6 @@
     "Context": {
       "ReferenceDateTime": "2019-05-09T12:00:00"
     },
-    "NotSupported": "javascript",
```
tellarin: ComplexCalendar is only supported in .NET, right? We should add "NotSupportedByDesign" for javacript, python, and java (or the subset that's really not supported) to all cases in this spec to make that clear.

PR #1565, opened 2019-05-12
Fix "Friday 4/12 returns Friday 4 as time"

songwenhao1:
Try: "Let us have a meeting this Friday 4/12"
Expected: "this friday 4/12"


Patterns/English/English-DateTime.yaml
```diff
@@ -253,10 +253,10 @@ DateExtractor6: !nestedRegex
   def: (?<={DatePreposition}\s+)({WeekDayRegex}\s+)?{MonthNumRegex}[\-\.]{DayRegex}(?![%])\b
   references: [ MonthNumRegex, DayRegex, WeekDayRegex, DatePreposition ]
 DateExtractor7L: !nestedRegex
-  def: \b({WeekDayRegex}\s+)?{MonthNumRegex}\s*/\s*{DayRegex}((\s+|\s*,\s*|\s+of\s+){DateYearRegex})(?![%])\b
+  def: \b((this\s+)?{WeekDayRegex}\s+)?{MonthNumRegex}\s*/\s*{DayRegex}((\s+|\s*,\s*|\s+of\s+){DateYearRegex})(?![%])\b
```
tellarin: I'm not sure this is the best approach to solve the issue.
This is altering two base regexes. Why not deal with relative terms elsewhere?
There should be a principle behind the fix.

STayinloves: I think this's a reasonable approach for a quick fix, following what the previous PR do. But if for further scalability consideration, we can try to replace the `WeekDayRegex` in all `DateExtractor` regexes with something like a `RelativeWeekDayRegex`, not sure if it'll introduce some side effect.

songwenhao1:
This PR is no longer needed. Fix was implemented differently in #1573.

PR #1567, opened 2019-05-15
Added Dutch DateTime specs, initial versions

LionbridgeCS:
Initial versions of DateTime DatePeriodExtractor and DatePeriodParser

### DIFF
--- a/Specs/DateTime/Dutch/DatePeriodParser.json
+++ b/Specs/DateTime/Dutch/DatePeriodParser.json
@@ -1,6 +1,8 @@
 ﻿[
   {
     "Input": "Ik ben deze maand van 4 tot 22 weg",
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign" : "javascript,python,java",
     "Results": [
       {
         "Text": "deze maand van 4 tot 22",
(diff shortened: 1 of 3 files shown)


Specs/DateTime/Dutch/DatePeriodParser.json
```diff
@@ -0,0 +1,2137 @@
+﻿[
+  {
+    "Input": "Ik ben deze maand van 4 tot 22 weg",
```
tellarin: Thanks for all the new specs!
The build is failing as many of these are not currently supported.
Each spec case should have an attribute "NotSupported" : "dotnet", until the functionality is implemented.
Also, each spec case should have an attribute "NotSupportedByDesign" : "javascript,python,java" to highlight that implementations in those platforms won't exist in the short term.

Specs/DateTime/Dutch/DatePeriodParser.json
```diff
@@ -0,0 +1,2137 @@
...
+      {
+        "Text": "deze maand van 4 tot 22",
+        "Type": "daterange",
+        "Start": 7,
+        "Length": 23
+      }
+    ]
+  },
+  {
+    "Input": "Ik ben volgende maand van 4-23 weg",
```
tellarin: *Parser.json and *Extractor.json specs test sub-modules. But each culture should have an overall spec (DateTimeModel.json) that ensures that the final model output is correct. 
Ideally close parity to the English DateTimeModel.json is desired. But it could be sufficient to at least _sample_ enough cases from all *Parser.json specs and represent them there.

PR #1573, opened 2019-05-21
Improves handling of relative dates and fixes "this friday 4/12"

songwenhao1:
Fix "relative date" and "this friday 4/12", in .NET for now.
Add DateExtractorYearTermRegex; 
Modify DateExtractor1 and merge it with DateExtractor2;
Add RelativeRegex to DateExtractor6/8 to fix "on this 5.6"
Add StrictRelativeRegex in French.

Javascript support will come in a followup PR as there's a missing method implementation in JS.


.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishDateExtractorConfiguration.cs
```diff
@@ -196,12 +199,9 @@ public EnglishDateExtractorConfiguration(IOptionsConfiguration config)
 
             DateRegexList = new List<Regex>
             {
-                // ((this)? Sunday,)? April 5
+                // (Sunday,)? April 5
```
tellarin: Please change all comments for the merged regex to // (Sunday,)? April 5 or (Sunday,)? April 5, 2016 
to illustrate both are covered together.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -24,6 +24,7 @@ public static class DateTimeDefinitions
 		public const string TillRegex = @"(?<till>au|[aà]|et|jusqu'[aà]|avant|--|-|—|——)";
 		public const string RangeConnectorRegex = @"(?<and>et|de la|au|[aà]|et\s*la|--|-|—|——)";
 		public const string RelativeRegex = @"(?<order>prochain|prochaine|de|du|ce|cette|l[ae]|derni[eè]re|pr[eé]c[eé]dente|au\s+cours+(de|du\s*))";
+		public const string StrictRelativeRegex = @"(?<order>prochain|prochaine|ce|cette|derni[eè]re|pr[eé]c[eé]dente|au\s+cours+(de|du\s*))";
```
tellarin: Why declare a new Regex but it's not used anywhere? Shouldn't other pieces reference it?

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -112,16 +112,16 @@ public static class DateTimeDefinitions
...
+		public static readonly string DateExtractor1 = $@"\b({WeekDayRegex}\s*[,-]?\s*)?(({MonthRegex}(\.)?\s*[/\\.,-]?\s*{DayRegex})|(\({MonthRegex}\s*[-.]\s*{DayRegex}\)))(\s*\(\s*{WeekDayRegex}\s*\))?({DateExtractorYearTermRegex}\b)?";
 		public static readonly string DateExtractor3 = $@"\b({WeekDayRegex}(\s+|\s*,\s*))?{DayRegex}(\.)?(\s+|\s*,\s*|\s+of\s+|\s*-\s*){MonthRegex}(\.)?((\s+|\s*,\s*){DateYearRegex})?\b";
 		public static readonly string DateExtractor4 = $@"\b{MonthNumRegex}\s*[/\\\-]\s*{DayRegex}(\.)?\s*[/\\\-]\s*{DateYearRegex}";
 		public static readonly string DateExtractor5 = $@"\b{DayRegex}\s*[/\\\-\.]\s*({MonthNumRegex}|{MonthRegex})\s*[/\\\-\.]\s*{DateYearRegex}";
-		public static readonly string DateExtractor6 = $@"(?<={DatePreposition}\s+)({WeekDayRegex}\s+)?{MonthNumRegex}[\-\.]{DayRegex}(?![%])\b";
-		public static readonly string DateExtractor7L = $@"\b({WeekDayRegex}\s+)?{MonthNumRegex}\s*/\s*{DayRegex}((\s+|\s*,\s*|\s+of\s+){DateYearRegex})(?![%])\b";
+		public static readonly string DateExtractor6 = $@"(?<={DatePreposition}\s+)({RelativeRegex}\s+)?({WeekDayRegex}\s+)?{MonthNumRegex}[\-\.]{DayRegex}(?![%])\b";
```
tellarin: Why use RelativeRegex and not StrictRelativeRegex here?

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchDateExtractorConfiguration.cs
```diff
@@ -69,6 +69,9 @@ public class DutchDateExtractorConfiguration : BaseOptionsConfiguration, IDateEx
         public static readonly Regex RelativeMonthRegex =
             new Regex(DateTimeDefinitions.RelativeMonthRegex, RegexOptions.Singleline);
 
+        public static readonly Regex RelativeRegex =
+            new Regex(DateTimeDefinitions.RelativeRegex, RegexOptions.Singleline);
+
         public static readonly Regex PrefixArticleRegex =
```
tellarin: DateExtractor2 should also be removed from every other language and their YAML files, right?

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Parsers/DutchDateParserConfiguration.cs
```diff
@@ -136,7 +139,7 @@ public DutchDateParserConfiguration(ICommonDateTimeParserConfiguration config)
 
         public IDateTimeUtilityConfiguration UtilityConfiguration { get; }
 
-        public int GetSwiftMonth(string text)
+        public int GetSwiftMonthOrYear(string text)
```
tellarin: This is a rename for consistency with parsers in other cultures?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -134,10 +134,20 @@ private List<Token> BasicRegexMatch(string text)
                 var matches = regex.Matches(text);
                 foreach (Match match in matches)
                 {
-                    // some match might be part of the date range entity, and might be splitted in a wrong way
+                    // some match might be part of the date range entity, and might be split in a wrong way
                     if (ValidateMatch(match, text))
                     {
-                        ret.Add(new Token(match.Index, match.Index + match.Length));
+                        var preText = text.Substring(0, match.Index);
+                        var relativeRegex = this.Config.RelativeRegex.MatchEnd(preText, trim: true);
```
tellarin: Can you add a code comment here explaining why/what this is doing?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -134,10 +134,20 @@ private List<Token> BasicRegexMatch(string text)
...
-                    // some match might be part of the date range entity, and might be splitted in a wrong way
+                    // some match might be part of the date range entity, and might be split in a wrong way
                     if (ValidateMatch(match, text))
                     {
-                        ret.Add(new Token(match.Index, match.Index + match.Length));
+                        var preText = text.Substring(0, match.Index);
+                        var relativeRegex = this.Config.RelativeRegex.MatchEnd(preText, trim: true);
+                        if (relativeRegex.Success)
+                        {
+                            ret.Add(new Token(relativeRegex.Index, match.Index + match.Length));
```
tellarin: You should check that the relative term is right before the detected entity no? Does match end already enforce it?
"I'm not talking about this, but about 4/Jan", for example. Please add it as a spec to show behaviour is correct.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -144,18 +144,40 @@ private DateTimeResolutionResult ParseBasicRegexMatch(string text, DateObject re
...
             {
                 var offset = 0;
+                var swift = -2;
                 var match = regex.Match(trimmedText);
                 if (!match.Success)
                 {
                     match = regex.Match(this.config.DateTokenPrefix + trimmedText);
-                    offset = this.config.DateTokenPrefix.Length;
+
+                    // Handing cases with "on"
```
tellarin: Why is this change needed if you've made extra changes to the base regex for this case?
Sorry, I can't get the meaning of the change just from the code.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -702,13 +724,14 @@ private DateTimeResolutionResult ParseDurationWithAgoAndLater(string text, DateO
...
         // Parse a regex match which includes 'day', 'month' and 'year' (optional) group
-        private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDate)
+        private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDate, int swift)
         {
             var ret = new DateTimeResolutionResult();
 
             var monthStr = match.Groups["month"].Value.ToLower();
             var dayStr = match.Groups["day"].Value.ToLower();
             var yearStr = match.Groups["year"].Value.ToLower();
+            var weekStr = match.Groups["weekday"].Value.ToLower();
```
tellarin: If it's weekday, it's better to call the var weekdayStr and not weekStr.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -702,13 +724,14 @@ private DateTimeResolutionResult ParseDurationWithAgoAndLater(string text, DateO
         }
 
         // Parse a regex match which includes 'day', 'month' and 'year' (optional) group
-        private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDate)
+        private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDate, int swift)
```
tellarin: Why is a new swift necessary?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -144,18 +144,40 @@ private DateTimeResolutionResult ParseBasicRegexMatch(string text, DateObject re
             foreach (var regex in this.config.DateRegexes)
             {
                 var offset = 0;
+                var swift = -2;
```
tellarin: Why -2? What does it mean? Some invalid value?
If so, why is it needed? It should also be declared in Constants so it's easier to see what it means and make sure the right value is reused where necessary.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -3962,14 +3962,9 @@
         "Resolution": {
           "values": [
             {
-              "timex": "XXXX-06-15",
+              "timex": "2018-06-15",
```
tellarin: Why this change? This is incorrect.
The original sentence doesn't force 2018. Why is the timex modified??

.NET/Microsoft.Recognizers.Text.DateTime/French/Extractors/FrenchDateExtractorConfiguration.cs
```diff
@@ -79,6 +79,9 @@ public class FrenchDateExtractorConfiguration : BaseOptionsConfiguration, IDateE
         public static readonly Regex RelativeMonthRegex =
             new Regex(DateTimeDefinitions.RelativeMonthRegex, RegexOptions.Singleline);
 
+        public static readonly Regex RelativeRegex =
```
tellarin: This mismatch shouldn't happen.
A property called RelativeRegex should reference a regex called RelativeRegex, not StrictRelativeRegex.
If we need StrictRelativeRegex, than the prop should be StrictRelativeRegex.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -3952,7 +3952,7 @@
     "Context": {
       "ReferenceDateTime": "2018-06-20T00:00:00"
     },
-    "NotSupported": "python",
+    "NotSupported": "javascript, python, java",
```
tellarin: Until the fix is finalized in .NET, this is great. Once the code is final, then it can propagate to JS and Python before merging the PR.
Thanks!

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
songwenhao1: Yes, the matchEnd already enforce it, so i use the method.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
songwenhao1: Only when the front is "on", "5.12/24-12" will be recognized as date by DateExtractor6/8. So if I don't change the basic regex, "on this 5.12", "5.12" will not be matched, we can't continue with the rest of the operation.

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchDateExtractorConfiguration.cs
songwenhao1: Yes, but I found that we can only remove it from English and Dutch, while in other languages it is defined differently than in English.

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Parsers/DutchDateParserConfiguration.cs
songwenhao1: Yes, the method could also be reused in Year, so i rename the method.

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchDateExtractorConfiguration.cs
tellarin: You mean that in some cultures regex 1 and 2 are not both cases of the same regex with a different ending?

PR #1585, opened 2019-05-22
Additional DateTime Dutch specs

LionbridgeCS:


iMicknl:
Looks good, there are some sentences which don't look like 100% valid Dutch, but could be used by spoken language. Specs could be further enhanced during implementation.

Issue #1575, opened 2019-05-22
[EN DateTimeV2] Cannot recognize "before/after the year 2000"


ZijiaLin:
Only "year 2000" is recognized.

Issue #1574, opened 2019-05-22
[EN DateTimeV2] Incorrectly recognize "10 students" as a Datetime


substill:
**Describe the bug**
All strings containing "10 s****" will be recognized as a Datetime. 
Here, "##" represents a number, "*" represents any character. 
Examples: "10 students", "10 stars", ...

**Expected behavior**
"10 students", "10 stars", should not be recognized as Datetime.

PR #1586, opened 2019-05-23
Fix "10 students/stars" should not be recognized as a datetime.

songwenhao1:
1, Fix issue #1574 "10 students/stars" should not be recognized as a datetime.
2, Fix issue #1575 cannot recognize "after the year 2000"


Patterns/English/English-DateTime.yaml
```diff
@@ -513,7 +513,7 @@ BeforeRegex: !nestedRegex
   references: [ InclusiveModPrepositions ]
 # "starting from" is SinceRegex
 AfterRegex: !nestedRegex
-  def: (\b{InclusiveModPrepositions}?((after|(starting|beginning)(\s+on)?(?!\sfrom)|(?<!no\s+)later than)|(year greater than))(?!\s+or equal to){InclusiveModPrepositions}?\b\s*)|(?<!\w|<)((?<include>>=)|>)
+  def: ((\b{InclusiveModPrepositions}?((after|(starting|beginning)(\s+on)?(?!\sfrom)|(?<!no\s+)later than)|(year greater than))(?!\s+or equal to){InclusiveModPrepositions}?\b\s*?)|(?<!\w|<)((?<include>>=)|>))(\s+the)?
```
STayinloves: `*` means zero or more `\s`, why do we even need a `?` here?

songwenhao1: *? means Lazy quantifier. Appending the ? character to a quantifier makes it lazy; it causes the regular expression engine to match as few occurrences as possible. 
So that we can correctly match the "\s+the" in the behind

STayinloves: That's reasonable.

PR #1587, opened 2019-05-23
Recognize "exceed/surpass" in number range #1578

### LINKED ISSUE #1578: [EN NumberRange] Cannot recognize "exceed/surpass 3000"
The result should be like "> 3000".

songwenhao1:
Fixes #1578.


.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -73,7 +73,7 @@ public static class NumbersDefinitions
 		public static readonly string NumberWithPrefixPercentage = $@"(per cent of|percent of|percents of)(\s*)({BaseNumbers.NumberReplaceToken})";
 		public static readonly string NumberWithPrepositionPercentage = $@"({BaseNumbers.NumberReplaceToken})\s*(in|out\s+of)\s*({BaseNumbers.NumberReplaceToken})";
 		public const string TillRegex = @"(to|through|--|-|—|——|~|–)";
-		public const string MoreRegex = @"((bigger|greater|more|higher|larger)(\s+than)?|above|over|(?<!<|=)>)";
+		public const string MoreRegex = @"((bigger|greater|more|higher|larger|exceed|surpass)(\s+than)?|above|over|(?<!<|=)>)";
```
tellarin: "exceed|surpass" should be with "above|over" in this case. They can't happen before "than"

PR #1594, opened 2019-05-24
Dutch DateTime Specs

LionbridgeCS:
Completed Dutch localization for the DateTime specs

PR #1598, opened 2019-05-27
Dutch specs localized and filtered

LionbridgeCS:
Review of pull request  #1594

### DIFF
--- a/Specs/DateTime/Dutch/HolidayParser.json
+++ b/Specs/DateTime/Dutch/HolidayParser.json
@@ -303,5 +303,345 @@
         "Length": 17
       }
     ]
+  },
+  {
+    "Input": "Ik ga op Pasen terug",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "Pasen",
+        "Type": "date",
+        "Value": {
+          "Timex": "",
+          "FutureResolution": {
+            "date": "0001-01-01"
+          },
+          "PastResolution": {
+            "date": "0001-01-01"
+          }
+        },
+        "Start": 9,
+        "Length": 6
+      }
+    ]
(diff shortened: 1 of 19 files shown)


tellarin:
@LionbridgeCS Please don't close a PR with ongoing reviews to create a new one, unless that's discussed on the original PR for clarity. Otherwise it gets much harder to track if review comments were addressed.
If a commit is mistakenly pushed to a PR, you can update the PR reverting it.

Specs/DateTime/Dutch/HolidayParser.json
```diff
@@ -303,5 +303,345 @@
...
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "Pasen",
+        "Type": "date",
+        "Value": {
+          "Timex": "",
```
tellarin: @iMicknl had asked a question about this, right?
@LionbridgeCS, could you please reply here?

LionbridgeCS: This one was localized from the English counter part which does not have a Timex for Easter.

LionbridgeCS:
@tellarin Thank you for the directive, noted.

Specs/DateTime/Dutch/HolidayParser.json
```diff
@@ -303,5 +303,345 @@
...
+          },
+          "PastResolution": {
+            "date": "2018-05-01"
+          }
+        },
+        "Start": 29,
+        "Length": 7
+      },
+      {
+        "Text": "Int'l Workers Day",
```
iMicknl: This string doesn't exist in `Input` text, so this resolution should be removed.

Specs/DateTime/Dutch/HolidayParser.json
iMicknl: Thanks, had a look at English and Easter is indeed a date that can't have a Timex.

PR #1599, opened 2019-05-27
Korean NumbersWithUnit specs localized

LionbridgeCS:
The requested extra cases for explicit USD, Korean units, and other Korean-specific cases will be added with the implementation in future PRs.

### DIFF
--- a/Specs/NumberWithUnit/Korean/CurrencyModel.json
+++ b/Specs/NumberWithUnit/Korean/CurrencyModel.json
@@ -1,0 +1,1903 @@
+[
+  {
+    "Input": "메릴랜드 주 몽고메리 카운티, 7천 5백만 달러 일반 채권, 시리즈 b,   1989년 매뉴팩처러스 하노버 트러스트에 의해 강화된 공공 개선  채권이다. ",
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "7천 5백만",
+        "TypeName": "currency",
+        "Resolution": {
+          "value": "75000000",
+          "unit": "Dollar"
+        },
+        "Start": 17,
+        "End": 23
+      }
+    ]
+  },
+  {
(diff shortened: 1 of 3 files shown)


Specs/NumberWithUnit/Korean/CurrencyModel.json
```diff
@@ -0,0 +1,1903 @@
...
+    "Input": "메릴랜드 주 몽고메리 카운티, 7천 5백만 달러 일반 채권, 시리즈 b,   1989년 매뉴팩처러스 하노버 트러스트에 의해 강화된 공공 개선  채권이다. ",
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "7천 5백만",
+        "TypeName": "currency",
+        "Resolution": {
+          "value": "75000000",
+          "unit": "Dollar"
```
tellarin: Are these general "dollar" or a term specific to US dollar? Sorry, I'm not familiar with Korean, so I'm double checking.

Specs/NumberWithUnit/Korean/DimensionModel.json
```diff
@@ -0,0 +1,838 @@
+[
```
tellarin: Are there any Korea-specific units that are common enough the recognizers should support?

Specs/NumberWithUnit/Korean/CurrencyModel.json
tellarin: If all these cases are the general "dollar", can you add at least one case with the term specifically for USD too?

LionbridgeCS: These are the general dollar term. The requested will be added with the implementation.

Specs/NumberWithUnit/Korean/DimensionModel.json
LionbridgeCS: The Korean specific units will be added with the implementation.

PR #1600, opened 2019-05-27
Turkish Number specs localized

LionbridgeCS:
The requested extra cases for Turkish-specific cases will be added with the implementation in future PRs.

### DIFF
--- a/Specs/Number/Turkish/NumberModel.json
+++ b/Specs/Number/Turkish/NumberModel.json
@@ -1,0 +1,2466 @@
+[
+  {
+    "Input": "192",
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "javascript,python,java",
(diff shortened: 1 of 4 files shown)


Specs/Number/Turkish/NumberModel.json
```diff
@@ -0,0 +1,2466 @@
...
+      }
+    ]
+  },
+  {
+    "Input": " -123456789101231",
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": " -123456789101231",
```
tellarin: This " " should not be part of the tagged text.

LionbridgeCS:
The test cases here are the translations from the English counterparts. 
The requested will be added in the next PRs

PR #1603, opened 2019-05-29
Check for empty and duplicate inputs in Specs

songwenhao1:


.NET/Build.CI.cmd
```diff
@@ -26,6 +26,21 @@ if EXIST "%VSInstallDir%\MSBuild\!MsBuildVersion!\Bin\MSBuild.exe" (
 	EXIT /B
 )
 
+ECHO.
+ECHO # Check for empty and duplicate inputs in Specs
+Powershell -ExecutionPolicy Bypass "& {buildtools\checkSpec.ps1; exit $LastExitCode }"
+IF %ERRORLEVEL% NEQ 0 (
```
tellarin: This check and multiple code and messages could be eliminated if the .ps1 script already outputs the tight error message an a final error code.

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+$global:emptyFileList = New-Object System.Collections.Generic.List[string]
+$global:duplicateFileList = New-Object System.Collections.Hashtable
+
+function SpecInfo()
+{
+    foreach ($file in $input) 
+    {   
+        $parentName = $file.FullName | Split-Path -parent | Split-Path -leaf
+        $typeFolder = $file.FullName | Split-Path -parent | Split-Path -parent | Split-Path -leaf
+		$ret = Get-Content $file.FullName -encoding utf8 | ConvertFrom-Json
```
tellarin: Please convert all tabs to spaces. :)

tellarin: $ret can be renamed to something like $contents to better represent what it contains.

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+
+    Write-Host("Total empty input:`t" + $global:totalEmpty)
+    Write-Host("Total duplicate input:`t" + $global:totalDuplicate)
+
+    if ($global:totalEmpty)
+    {
+        Write-Host("Contain empty input `"`" in these files")
+        foreach ($case in $global:emptyFileList)
+        {
+            Write-Host($case)
```
tellarin: It would be better if for each spec that "fails" there is a count for each duplicate line.
The current output seems a little ambiguous.

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+        {
+            Write-Host("$($key)")
+            foreach ($case in $global:duplicateFileList[$key])
+            {
+                Write-Host($case)
+            }
+        }
+    }
+
+    if ($global:totalEmpty -OR $global:totalDuplicate)
```
tellarin: No need for multiple codes. This script can output the write message and details, then just fail the build.

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+    $duplicateList = New-Object System.Collections.Generic.List[string]
+    $isDateTime = $false
+    $fileName = "$($type)/$($parent)/$($name)"
+
+    if ($type -eq "DateTime")
+    {
+        $isDateTime = $true
+    }
+
+    foreach ($line in $spec)
```
tellarin: $line could be renamed to $testCase as it's not really a line from the text file and actually a full object for the whole case.

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+    {
+        if ($line.Input -eq "")
```
tellarin: Can you also check for missing "Input" attributes in a test case?

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+        {
+            $emptyNum += 1
+        }
+        else
```
tellarin: This else could be simplified. But I think this is acceptable for now.

.NET/buildtools/checkSpec.ps1
```diff
@@ -0,0 +1,129 @@
...
+        {
+            if ($inputDict.ContainsKey($line.Input))
+            {
+                if ($isDateTime -And $line.Context.ReferenceDateTime) 
+                {   
+                    $timeDict = $inputDict.($line.Input)
```
tellarin: Please rename $timeDict and $referenceDict below to make it clear that they are the same container and that it's not a dictionary.

PR #1610, opened 2019-05-31
Revise Chinese Date Recognizer

Grey0202:
- Added support for "早上/中午/下午/夜里"

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/BaseURL.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/BaseURL.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Recognizers.Definitions
 	{
 		public const string ProtocolRegex = @"((https?|ftp):\/\/)";
 		public const string PortRegex = @"(:\d{1,5})";
-		public const string ExtractionRestrictionRegex = @"(?<=\s|[\'""\(\[:]|^)";
+		public const string ExtractionRestrictionRegex = @"(?<=\s|[\'""\(\[:]|[\u4E00-\u9FFF]|^)";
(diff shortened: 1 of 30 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/BaseURL.cs
```diff
@@ -23,7 +23,7 @@ public static class BaseURL
 	{
 		public const string ProtocolRegex = @"((https?|ftp):\/\/)";
 		public const string PortRegex = @"(:\d{1,5})";
-		public const string ExtractionRestrictionRegex = @"(?<=\s|[\'""\(\[:]|^)";
+		public const string ExtractionRestrictionRegex = @"(?<=\s|[\'""\(\[:]|[\u4E00-\u9FFF]|^)";
```
tellarin: What's the goal of this change? By design we don't support unicode in URLs (unless there is a very strong reason to add it).
There is also no new spec in this PR that relates to it. Any new fix or extension in code needs to at least have a new spec to show that works.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateTimePeriodParserConfiguration.cs
```diff
@@ -429,6 +431,12 @@ private DateTimeResolutionResult ParseSpecificNight(string text, DateObject refe
                 beginHour = 8;
                 endHour = Constants.HalfDayHourCount;
             }
+            else if (NORegex.IsMatch(trimmedText))
+            {
+                timeStr = "TNO";
```
tellarin: There is no such concept as TNO in timex expressions. If this is the same as regular "noon" in English, then it should just be a time.
You can always test similar cases in English to see how it handles things for a ballpark idea of expected outputs.

PR #1611, opened 2019-06-03
Fix cases like "thursday morning est" where timezone is tagged, but not parsed in resolution

songwenhao1:


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
```diff
@@ -372,6 +377,17 @@ protected virtual DateTimeResolutionResult ParseSpecificTimeOfDay(string text, D
                         DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, beginHour, 0, 0),
                         DateObject.MinValue.SafeCreateFromValue(pastDate.Year, pastDate.Month, pastDate.Day, endHour, endMin, endMin));
 
+                // Handle text contain timeZone, like "thursday morning est", try to get the timezone resolution.
+                var zoneErs = this.Config.TimePeriodExtractor.Extract(timeText + ' ' + afterStr, referenceTime);
```
STayinloves: I don't think this implement is suitable.
Timezone is supposed to be extracted in the extractor, it should be firstly extracted alone and merged with TimePeriod afterward.
Please see the [implement](https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs#L45) in TimeExtractor for reference.

PR #1615, opened 2019-06-06
Fix URL issue in Japanese/Chinese

Grey0202:
Now it can extract when no whitespace between URL and the Unicode text.


.NET/Microsoft.Recognizers.Text.Sequence/Chinese/Extractors/ChineseURLExtractor.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Text.RegularExpressions;
+
+using Microsoft.Recognizers.Definitions.Chinese;
+using Microsoft.Recognizers.Text.Matcher;
+
+namespace Microsoft.Recognizers.Text.Sequence
+{
+    public class ChineseURLExtractor : BaseSequenceExtractor
+    {
+        public ChineseURLExtractor()
```
tellarin: The same principle of avoiding duplication should be applied here.
The base URL extractor could receive a config (like, for example HolidayParser) and the value inside the config would be used in the appropriate place.
English would use directly from the Base patterns, Chinese/Japanese from their CJ config (as far as I understand, Korean doesn't need this fix as they use whitespace as word separator).

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -89,7 +89,7 @@ public IModel GetEmailModel(string culture = null, bool fallbackToDefaultCulture
 
         public IModel GetURLModel(string culture = null, bool fallbackToDefaultCulture = true)
         {
-            return GetModel<URLModel>(Culture.English, fallbackToDefaultCulture);
+            return GetModel<URLModel>(culture, fallbackToDefaultCulture);
```
tellarin: This change will introduce error. The current behaviour is that any language receives the base implementation (English). After having a customized version for Chinese/Japanese there maybe would be logic here to chose which of the two existing models to return.

Specs/Sequence/Chinese/URLModel.json
```diff
@@ -0,0 +1,527 @@
...
+        "End": 23,
+        "TypeName": "url",
+        "Resolution": {
+          "value": "www.microsoft.com"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "はmicrosoft.comです",
```
tellarin: Chinese/URLModel.json should only contain cases that we want to validate for Chinese.
Can you move this Japanese case to it's own spec and make sure the tests run for it?

Patterns/Chinese/Chinese-URL.yaml
```diff
@@ -0,0 +1,330 @@
+---
+ProtocolRegex: !simpleRegex
+  def: ((https?|ftp):\/\/)
+PortRegex: !simpleRegex
+  def: (:\d{1,5})
+ExtractionRestrictionRegex: !simpleRegex
+  def: (?<=\s|[\'""\(\[:：]|^|[\u0800-\u9FFF])
```
tellarin: I'm not sure this is the best approach to address the problem.
Adding "|[\u0800-\u9FFF]" will cause a big performance hit. Wouldn't removing the check for whitespace be a better choice?

tellarin:
The build is also currently failing in JS. Once the .NET changes are reviewed, can you propagate the same implementation to JS in this PR?
Thanks.

Patterns/Chinese/Chinese-URL.yaml
tellarin: The desired behaviour is that URLs are extracted if there is a separating whitespace, but also if there isn't one.

.NET/Microsoft.Recognizers.Text.Sequence/Config/ISequenceConfiguration.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.Recognizers.Text.Sequence
+{
+    public interface ISequenceConfiguration
```
tellarin: Why do we need this interface and not just use Options?
If we need this, we can rename it to IOptionsConfiguration for consistency.

.NET/Microsoft.Recognizers.Text.Sequence/Chinese/Extractors/ChineseURLExtractorConfiguration.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Recognizers.Definitions;
+using Microsoft.Recognizers.Definitions.Chinese;
+
+namespace Microsoft.Recognizers.Text.Sequence.Chinese
+{
+    public class ChineseURLExtractorConfiguration : SequenceConfiguration, IURLExtractorConfiguration
+    {
+        public static readonly Regex IpUrlRegex = new Regex(BaseURL.IpUrlRegex, RegexOptions.Compiled);
+
+        public static readonly Regex UrlRegex = new Regex(BaseURL.UnicodeUrlRegex, RegexOptions.Compiled);
```
tellarin: BaseURL.UnicodeUrlRegex seems to be the only difference. Can't the config just have this and the base class get it from the config? Why do we need to repeat everything else?

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -89,7 +90,12 @@ public IModel GetEmailModel(string culture = null, bool fallbackToDefaultCulture
 
         public IModel GetURLModel(string culture = null, bool fallbackToDefaultCulture = true)
         {
-            return GetModel<URLModel>(culture, fallbackToDefaultCulture);
+            if (culture == "zh-cn" || culture == "ko-kr" || culture == "ja-jp")
```
tellarin: I don't think Korean needs this. As far as I know, it uses whitespace as separator everywhere.
Also, this could be an attribute in the language-specific url configs. IsWhitespaceSeparatorCulture. Only Chinese and Japanese would return true.
The attribute could be declared in YAML and exposed through the new config class you added.

Patterns/Base-URL.yaml
```diff
@@ -5,17 +5,25 @@ PortRegex: !simpleRegex
   def: (:\d{1,5})
 ExtractionRestrictionRegex: !simpleRegex
   def: (?<=\s|[\'""\(\[:]|^)
+UnicodeRestrictionRegex: !simpleRegex
```
tellarin: Why add this here?
It would be better if this is the configurable part. Or if the regexes are assembled on startup. Now there are basically 3 pairs of repeated regexes due to only a minor change.

Specs/Sequence/Chinese/URLModel.json
```diff
@@ -1,449 +1,4 @@
 [
-  {
-    "Input": "https://abc.com",
```
tellarin: Please keep all cases as the single URLs should still work in Chinese.
Also, the other cases should be translated to Chinese, ok?

Specs/Sequence/Japanese/URLModel.json
```diff
@@ -0,0 +1,18 @@
+[
+  {
+    "Input": "はmicrosoft.comです",
```
tellarin: Here I think it's OK to leave only one Japanese case for now. The goal of this spec is to show that Japanese will reuse the Chinese model and things will work.

.NET/Microsoft.Recognizers.Text.Sequence/Chinese/Extractors/ChineseURLExtractorConfiguration.cs
Grey0202: > BaseURL.UnicodeUrlRegex seems to be the only difference. Can't the config just have this and the base class get it from the config? Why do we need to repeat everything else?

I am not sure if there will be other special circumstances, so I kept this part first.

tellarin: So let's make the change for the config to only have the difference, ok?

PR #1617, opened 2019-06-06
More holidays improvements

wlifuture:
St. Patrick's day, Earth day, Presidents day, Cyber Monday, Veterans day.


Patterns/English/English-DateTime.yaml
```diff
@@ -986,7 +988,7 @@ HolidayNames: !dictionary
     inaugurationday: [ inaugurationday ]
     groundhougday: [ groundhougday ]
     valentinesday: [ valentinesday ]
-    stpatrickday: [ stpatrickday ]
+    stpatrickday: [ stpatrickday, stpatricksday, st.patrickday, st.patricksday ]
```
tellarin: We should add a normalization in the parser to avoid having to add all forms here.
Do you think you can do that in a future PR?

tellarin:
@wlifuture When creating a new PR, please rebase on top of /master, ok? Otherwise the commit history in the PR contains multiple unrelated items which can make review harder.

Patterns/English/English-DateTime.yaml
wlifuture: We can discuss possible normalization logic. Well for this specific one, I'm creating a new PR to address the improvement.

Issue #1621, opened 2019-06-10
[EN DateTimeV2] "[year]-q[n]" pattern not supported


wlifuture:
Currently e.g. "2017 q2" works well but "2017-q2" doesn't. Low-pri to: @wlifuture

PR #1624, opened 2019-06-11
Fix [year]-q[n]" pattern not supported

songwenhao1:
Fixes issue #1621 : support for cases like "2017-q3", "2016-h1"


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -11745,5 +11745,101 @@
         }
       }
     ]
+  },
+  {
+    "Input": "2017 -q1",
```
tellarin: This is incorrect, the pattern does not include whitespace. It should be like "2017-q1".
Also, can you add full sentences in the specs instead of just the entity? For example: "The sales went up during 2017-q1" or "2019 started as a good year, but 2019-H2 will bring challenges".

tellarin:
Closing this as it was surpassed by #1632.

PR #1625, opened 2019-06-12
Refactoring language characteristics from hardcoded to Number YAML files

tellarin:
- Refactoring language characteristics from hardcoded in BaseParser to Number YAML files;
- Handling of bool declarations in YAML across platforms;
- Re-generating all resources;
- Standardizing all auto-gen resource code with spaces.


JavaScript/packages/resource-generator/lib/code-writers.ts
```diff
@@ -163,6 +176,9 @@ export function GenerateCode(root: any): CodeWriter[] {
         else if (token instanceof Array) {
             lines.push(new ArrayWriter(tokenName, token));
         }
+        else if (typeof token === "boolean") {
```
tellarin: Didn't work with instanceof Boolean

STayinloves: Yes, in javascript `true` is a boolean **primitive**, only boolean objects are instances of Boolean.

PR #1632, opened 2019-06-13
Fix [year]-q[n]" pattern not supported

songwenhao1:
Fixes issue #1621 : support for cases like "2017-q3", "2016-h1", "from 2017-q1 to 2018-q1"


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -265,7 +265,9 @@ public static class DateTimeDefinitions
       public static readonly string DecadeWithCenturyRegex = $@"(the\s+)?(((?<century>\d|1\d|2\d)?(')?(?<decade>\d0)(')?(\s)?s\b)|(({CenturyRegex}(\s+|-)(and\s+)?)?{DecadeRegex})|({CenturyRegex}(\s+|-)(and\s+)?(?<decade>tens|hundreds)))";
       public static readonly string RelativeDecadeRegex = $@"\b((the\s+)?{RelativeRegex}\s+((?<number>[\w,]+)\s+)?decades?)\b";
       public static readonly string YearPeriodRegex = $@"((((from|during|in)\s+)?{YearRegex}\s*({TillRegex})\s*{YearRegex})|(((between)\s+){YearRegex}\s*({RangeConnectorRegex})\s*{YearRegex}))";
-      public static readonly string ComplexDatePeriodRegex = $@"(((from|during|in)\s+)?(?<start>.+)\s*({TillRegex})\s*(?<end>.+)|((between)\s+)(?<start>.+)\s*({RangeConnectorRegex})\s*(?<end>.+))";
+      public static readonly string StrictTillRegex = $@"(?<till>\b(to|till|til|until|thru|through)\b|{BaseDateTime.RangeConnectorSymbolRegex}(?!\s*(h[1-2]|q[1-4])(?!(\s+of|\s*,\s*))))";
```
tellarin: This seems a strange fix and hard to maintain. There may also be a perf hit.
Why not just filter out false positives after a ComplexDatePeriodRegex match?

songwenhao1: Yes, in the early time, i just want to fix cases like "from 2017-q1 to  2017-q2", so i add "?!(h[1-2]|q[1-4])", but i found cases like "from q1 of 2017 - q2 of 2017" whick work before, now will don't work, so i add "?!(\s+of|\s*,\s*)", these modify make the fix looks strange..
From code, we can find the issue is caused by the capture groups couldn't capture right subtext. when the text contain several till/rangeConnector symbol, we couldnot found the right one which should be used. 
If we want to filter out false positives after a ComplexDatePeriodRegex match. 
First we need to judge if the group capture is right . then we need to think how to get the right subtext or we choose directly extract and parse the datepriod in the text... it maybe need to think how to suitable modify the method "ParseComplexDatePeriod".

tellarin: I see. The main issue I have with it is that there is too much repetition. Ideally we should reuse the till/rangeconnector regexes and just add the restriction.
I'm fine with merging it this way if there is no performance hit. Let me run some tests.

tellarin: It seems perf hit is not big. Let's merge as is to unblock it. We can improve re-use later.

PR #1634, opened 2019-06-14
Fix URL issue in Japanese/Chinese

Grey0202:
Now it can extract when no whitespace between URL and the Unicode text.
Refactored .NET and JS URL recognizer
This PR replaces #1615.


.NET/Microsoft.Recognizers.Text.Sequence/Japanese/JapaneseURLExtractorConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions;
+using Microsoft.Recognizers.Definitions.Chinese;
+
+namespace Microsoft.Recognizers.Text.Sequence.Japanese
```
tellarin: Why create a Japanese config and extractor? We don't need these as there is not Japanese-specific change. It's just a CN/JP vs. EN fork, like the previous PR, no?
Keep the Japanese tests though.

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -89,7 +91,7 @@ public IModel GetEmailModel(string culture = null, bool fallbackToDefaultCulture
 
         public IModel GetURLModel(string culture = null, bool fallbackToDefaultCulture = true)
         {
-            return GetModel<URLModel>(Culture.English, fallbackToDefaultCulture);
+            return GetModel<URLModel>(culture, fallbackToDefaultCulture);
```
tellarin: This is a regression compared to the previous PR. It will fail for other languages that are not EN/CN/JP.
Can you add back that if/else that returns the customized extractor for CN/JP and the base English one for others?

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -121,7 +123,21 @@ protected override void InitializeConfiguration()
...
+                    new URLParser(),
+                    new BaseURLExtractor(new EnglishURLExtractorConfiguration(options))));
+
+            RegisterModel<URLModel>(
+                Culture.Chinese,
+                options => new URLModel(
+                    new URLParser(),
+                    new BaseURLExtractor(new ChineseURLExtractorConfiguration(options))));
+
+            RegisterModel<URLModel>(
```
tellarin: Please remove this.

Specs/Sequence/Chinese/URLModel.json
```diff
@@ -0,0 +1,251 @@
...
+        "Text": "http://abc.com:8080",
+        "TypeName": "url",
+        "Resolution": {
+          "value": "http://abc.com:8080"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "网址是microsoft.com",
```
tellarin: Many test cases are missing here when compared to the English spec. Can you add the others and translate them when needed?
Of course you can leave all the new ones here too. Thanks!

PR #1638, opened 2019-06-17
Fix small datetime regression

tellarin:
And yet more consistency in spaces and linebreaks in auto-gen resources...


iMicknl:
@tellarin what about adding an [.editorconfig](https://editorconfig.org/) to help maintain consistency?

tellarin:
The problem was the auto-generated files from YAML. The original code gen was mixing tabs and whitespaces, as well as LF and CRLF; and the repo contained code with different combinations of these.
At least now most should be auto-generated code should be self-consistent. :)
This was making people frequent make mistakes in PRs and messing reviews. The pending items now is codegen for .NET. I can't find a way to have the TT transforms generate only LF. I'll probably add a post-processing script in the build at some point.

PS: Sorry, I had wrongly edited your comment. :-(

tellarin:
Adding an .editorconfig would still be very useful though. Do you want to add it, @iMicknl? ;-)

iMicknl:
@tellarin oops, didn't see all consistency changes were in autogenerated files. :-)

PR #1643, opened 2019-06-18
[Python][Choice] Add recognizers-text-choice to Python implementation

Aliandi:
### Description
This PR adds the support of `recognizers-text-choice` for Python. To the moment, the only ported language is **English**.

### Changes made

- Add all base classes for _extractor_, _parser_ and _model_
- Make tests for Choice available on Python
- Modify the Choice English patterns to use the correct _regex_ for emojis on Python
- Add support for English Choice 
- Add [regex](https://pypi.org/project/regex/), [grapheme](https://pypi.org/project/grapheme/) and [emoji](https://pypi.org/project/emoji/) libraries for Python
### Testing
_Running tests:_

imagen


Python/libraries/recognizers-choice/recognizers_choice/resources/english_choice.py
```diff
@@ -0,0 +1,15 @@
+# ------------------------------------------------------------------------------
+# <auto-generated>
```
tellarin: Why are there two versions of the auto-gen resources code?
Python/libraries/recognizers-choice/recognizers_choice/resources/english_choice.py
and
Python/libraries/recognizers-choice/resources/english_choice.py
The second one seems to be the correct one?

Aliandi: Indeed, I am deleting this file. Thanks!

PR #1648, opened 2019-06-19
Fix regression bug in ordinal.relative

songwenhao1:
Fixes cases like "the one before the last"

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
@@ -108,7 +108,7 @@ namespace Microsoft.Recognizers.Text.Number
                             };
 
                             // Add Metadata information for Ordinal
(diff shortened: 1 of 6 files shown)


Specs/Number/English/OrdinalModel.json
```diff
@@ -678,5 +678,73 @@
...
+        "Resolution": {
+          "offset": "0",
+          "relativeTo": "current",
+          "value": "current+0"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Look at current page",
```
tellarin: Minor: "Look at the current page"

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
```diff
@@ -108,7 +108,7 @@ public virtual List<ExtractResult> Extract(string source)
                             };
 
                             // Add Metadata information for Ordinal
-                            if (ExtractType.Contains(Constants.MODEL_ORDINAL))
+                            if ((Options & NumberOptions.ExperimentalMode) == 0 && ExtractType.Contains(Constants.MODEL_ORDINAL))
```
tellarin: This should be !SuppressedExtendedTypes

PR #1650, opened 2019-06-20
Refactor ordinal.relative code

songwenhao1:


.NET/Microsoft.Recognizers.Text.Number/Models/AbstractNumberModel.cs
```diff
@@ -88,11 +88,6 @@ private ModelResult BuildModelResult(ParseResult pn)
                     if (pn.Metadata != null && pn.Metadata.IsOrdinalRelative)
                     {
                         specificNumberType = Constants.MODEL_ORDINAL_RELATIVE;
```
tellarin: Why can't we also do this somewhere in the parser instead of having strange logic at the model level?
It seems most everything else was already moved.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -161,11 +161,18 @@ public virtual ParseResult Parse(ExtractResult extResult)
             // Add "offset" and "relativeTo" for ordinal
             if (!string.IsNullOrEmpty(ret.Type) && ret.Type.Contains(Constants.MODEL_ORDINAL))
             {
-                if ((this.Config.Options & NumberOptions.SuppressExtendedTypes) == 0 && Config.RelativeReferenceOffsetMap.ContainsKey(extResult.Text) &&
-                    Config.RelativeReferenceRelativeToMap.ContainsKey(extResult.Text))
+                if ((this.Config.Options & NumberOptions.SuppressExtendedTypes) == 0 && ret.Metadata.IsOrdinalRelative)
```
tellarin: Do we really need to set extra metadata to use here? Can't this be represented somewhere else in the extractor?
If not, that's fine. But can't the final type be set here too?

PR #1670, opened 2019-06-27
[JavaScript|Python] Adding try/catch to avoid exceptions crashing consumers

Grey0202:
-Add try/catch to JS .NET Python model file to avoid the crash.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
+++ b/.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
@@ -26,7 +26,6 @@ namespace Microsoft.Recognizers.Text.Choice
             {
                 var extractResults = Extractor.Extract(query);
                 parseResults = extractResults.Select(r => Parser.Parse(r));
-
             }
             catch (Exception)
             {
--- a/.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
+++ b/.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace Microsoft.Recognizers.Text.Choice
@@ -19,17 +20,25 @@ namespace Microsoft.Recognizers.Text.Choice
 
         public List<ModelResult> Parse(string query)
         {
-            var extractResults = Extractor.Extract(query);
-            var parseResults = extractResults.Select(r => Parser.Parse(r));
-
-            return parseResults.Select(pr => new ModelResult()
+            try
             {
-                Start = pr.Start.Value,
-                End = pr.Start.Value + pr.Length.Value - 1,
-                Resolution = GetResolution(pr),
-                Text = pr.Text,
-                TypeName = ModelTypeName,
-            }).ToList();
+                var extractResults = Extractor.Extract(query);
+                var parseResults = extractResults.Select(r => Parser.Parse(r));
+
+                // o can be null, add this part to try.
+                return parseResults.Select(pr => new ModelResult()
+                {
+                    Start = pr.Start.Value,
+                    End = pr.Start.Value + pr.Length.Value - 1,
+                    Resolution = GetResolution(pr),
+                    Text = pr.Text,
+                    TypeName = ModelTypeName,
+                }).ToList();
+            }
+            catch (Exception)
+            {
+                return null;
+            }
         }
 
         protected abstract SortedDictionary<string, object> GetResolution(ParseResult parseResult);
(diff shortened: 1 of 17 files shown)


.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
```diff
@@ -19,17 +20,25 @@ protected ChoiceModel(IParser parser, IExtractor extractor)
...
         {
-            var extractResults = Extractor.Extract(query);
-            var parseResults = extractResults.Select(r => Parser.Parse(r));
+            try
+            {
+                var extractResults = Extractor.Extract(query);
+                var parseResults = extractResults.Select(r => Parser.Parse(r));
 
-            return parseResults.Select(pr => new ModelResult()
+                // o can be null, add this part to try.
```
tellarin: What "o"?

.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
```diff
@@ -19,17 +20,25 @@ protected ChoiceModel(IParser parser, IExtractor extractor)
...
+            }
+            catch (Exception)
             {
-                Start = pr.Start.Value,
-                End = pr.Start.Value + pr.Length.Value - 1,
-                Resolution = GetResolution(pr),
-                Text = pr.Text,
-                TypeName = ModelTypeName,
-            }).ToList();
+                return null;
```
tellarin: Code should never return null. If no extracted entity, we should return an empty list.

JavaScript/packages/recognizers-choice/src/choice/models.ts
```diff
@@ -24,6 +26,10 @@ export abstract class ChoiceModel implements IModel {
                 text: o.text,
                 typeName: this.modelTypeName
             }));
+        }
+        catch(err) {
+            return null;
```
tellarin: Same.

JavaScript/packages/recognizers-choice/src/choice/models.ts
```diff
@@ -12,10 +12,12 @@ export abstract class ChoiceModel implements IModel {
...
 
     parse(source: string): ModelResult[] {
-        let extractResults = this.extractor.extract(source);
-        let parseResults = extractResults.map(r => this.parser.parse(r));
+        try {
+            let extractResults = this.extractor.extract(source);
+            let parseResults = extractResults.map(r => this.parser.parse(r));
 
-        return parseResults
+            // o can be null, add this part to try.
```
tellarin: How can it be null?

JavaScript/packages/recognizers-number-with-unit/src/numberWithUnit/models.ts
```diff
@@ -21,48 +21,56 @@ export abstract class AbstractNumberWithUnitModel implements IModel {
...
-            let extractResults = extractor.extract(query);
-            let parseResults: Array<ParseResult> = [];
-            for (let i = 0; i < extractResults.length; i++) {
-                let r = parser.parse(extractResults[i]);
-                if (r.value !== null) {
-                    if (r.value instanceof Array) {
-                        for (let j = 0; j < r.value.length; j++) {
-                            parseResults.push(r.value[j]);
+
+        try{
```
tellarin: Shouldn't this type of try/catch be in the extract/parser level to make sure individual entity processings don't blow?

Python/libraries/recognizers-choice/recognizers_choice/choice/models.py
```diff
@@ -16,18 +16,24 @@ def __init__(self, parser: Parser, extractor: Extractor):
...
-            model_result.text = o.text
-            model_result.type_name = self.model_type_name
-
-            result.append(model_result)
+
+        try:
+            extract_results = self.extractor.extract(source)
+            parse_results = [self.parser.parse(e) for e in extract_results]
+
+            # o can be null, add this part to try.
```
tellarin: "o" should never be null. When can this happen?

Python/samples/simple_console/sample.py
```diff
@@ -4,7 +4,7 @@
 from recognizers_suite import Culture, ModelResult
 
 # Use English for the Recognizers culture
-DEFAULT_CULTURE = Culture.English
+DEFAULT_CULTURE = Culture.Chinese
```
tellarin: Be careful not to commit such changes. Please revert this.

JavaScript/packages/recognizers-number-with-unit/src/numberWithUnit/models.ts
Grey0202: > Shouldn't this type of try/catch be in the extract/parser level to make sure individual entity processings don't blow?

This part of .NET does the same thing.

.NET/Microsoft.Recognizers.Text.Choice/Models/ChoiceModel.cs
```diff
@@ -20,27 +20,30 @@ protected ChoiceModel(IParser parser, IExtractor extractor)
...
             {
-                return null;
+                // Nothing to do. Exceptions in parse should not break users of recognizers.
+                // No result.
             }
+
+            return parseResults.Select(pr => new ModelResult()
+            {
+                Start = pr.Start.Value,
+                End = pr.Start.Value + pr.Length.Value - 1,
```
tellarin: Won't this return an invalid object and the consuming code will need to filter these out?

PR #1672, opened 2019-06-27
[Python][Style] Autopep8 + Flake8

Aliandi:
## Changes
- Add excluded files to `.flake8`: dist and build
- Change  `.pre-commit-config.yaml` to take the _flake8_ configuration from `.flake8`

## Example 
imagen

_Note:_ autopep8 will not fix issues that have to do with naming conventions as that may end up breaking the code. Those changes will have to be done manually.


Python/.pre-commit-config.yaml
```diff
@@ -4,9 +4,9 @@ repos:
     rev: v1.2.3
     hooks:
     - id: flake8
-      args: ['--config=.flake8', '--ignore=E501,F403,W504,F405,F401,E711,W503,E265,F811,E714,F841,E226,E713,F632,E722,F821']
-#
-#-   repo: https://github.com/pre-commit/mirrors-autopep8
-#    rev: master
-#    hooks:
-#    - id: autopep8
+      args: ['--config=Python/.flake8']
```
tellarin: Aha! :)

tellarin:
@Aliandi my understanding is that re-enabling autopep8 won't cause any problems before we fix the remaining style issues, right?
Also, can we use a config file (pycodestyle) in the pre-hook trigger like the flake8 case? Feel free to rename both config files.
Thanks!

Aliandi:
@tellarin it wouldn't cause any problems as it only changes the staged files! 
Also, yes, we can. We are adding a setup.cfg file to the root of the Python folder, following [this documentation](https://pycodestyle.readthedocs.io/en/latest/intro.html#configuration) where we unify all the configurations related to style

STayinloves:
After this PR, would it be possible to create a PR to only restyle all the python code (using the autopep8)? I did see some mixed styles in the code, in terms of line length limits and indentations.

tellarin:
@Aliandi, we should also delete the pycodestyle file now, right? Does a PR with he current flake8/autopep8 passes (i.e., it ignores those warning marked in config and don't raise others)?
@STayinloves, PR #1668 already applied the automatic autopep8 changes. But as @Aliandi says, some changes still need to be manually made.

Aliandi:
@tellarin yes, we've just deleted it. Also, the PR passes as the build is not configured to fail on flake8 or autopep8 issues. If you look at the AppVeyor build, you'll se that flake8 is being run and detecting the current naming issues. We'll create another PR fixing them manually to merge before this one

tellarin:
@Aliandi Great!  Merging it.

PR #1674, opened 2019-06-28
Fix  wrong recognition for  "from three thirty to four thirty"

songwenhao1:
Fix cases like "from three thirty to four thirty"  which would be wrong extracted in DatePeriodExtractor.
**NOTE**: This PR is a temporary fix, and MatchYearPeriod code will be refactored later.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
@@ -368,13 +368,25 @@ namespace Microsoft.Recognizers.Text.DateTime
                 {
                     var yearMatches = this.config.YearRegex.Matches(match.Value);
                     var allDigitYear = true;
+                    var isValidYear = true;
 
                     foreach (Match yearMatch in yearMatches)
                     {
-                        if (yearMatch.Length != Constants.FourDigitsYearLength)
+                        var year = config.DatePointExtractor.GetYearFromText(yearMatch);
+                        if (!(year >= Constants.MinYearNum && year <= Constants.MaxYearNum))
+                        {
+                            isValidYear = false;
+                            break;
+                        }
+                        else if (yearMatch.Length != Constants.FourDigitsYearLength)
                         {
                             allDigitYear = false;
                         }
+                    }
+
+                    if (!isValidYear)
+                    {
+                        continue;
                     }
 
                     // Cases like "2010-2015"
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -368,15 +368,27 @@ private List<Token> MatchYearPeriod(string text, DateObject referenceDate)
...
+                    var isValidYear = true;
 
                     foreach (Match yearMatch in yearMatches)
                     {
-                        if (yearMatch.Length != Constants.FourDigitsYearLength)
+                        var year = config.DatePointExtractor.GetYearFromText(yearMatch);
+                        if (!(year >= Constants.MinYearNum && year <= Constants.MaxYearNum))
+                        {
+                            isValidYear = false;
+                            break;
```
tellarin: Why this break? Then you are skipping all other matches in yearMatches.

tellarin: Does this call to YearRegex happen for every period with a year? Or only for "year-only" cases?
If the first, then it may also introduce other issues.

songwenhao1:
This PR is a temporary fix, and that MatchYearPeriod will be refactored later.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
songwenhao1: If one year is invalid in a year period, i think we needn't to check another one

tellarin: There may be more than one year involved in a sentence, no?

tellarin: And if you do this ```break```, then the ```if/continue``` never happen.

songwenhao1: I just break the loop of "yearMatches", don't check if another year is valid. why if/continue never happen..If/continue is out of the loop.

tellarin: That's what happens when people review on a tablet :)

PR #1675, opened 2019-06-28
[.NET] Add unit "'" filter in English and Italian

songwenhao1:
Add AmbiguityFiltersDict contains "'" filter in English and Italian.
Fix cases like "i ' m tried", 'm' should not be recognized.


Specs/NumberWithUnit/Italian/DimensionModel.json
```diff
@@ -624,5 +624,15 @@
         }
       }
     ]
+  },
+  {
+    "Input": "l'amministrazione",
```
tellarin: Can you add cases with real usage of sentences in Italian with something like "10l" and "20 l" to show that litter is properly extracted in those?
Or see if any specs currently contain litter examples but are marked as NotSupported in .NET?

PR #1690, opened 2019-07-03
Extending number range recognizer to support cases like "100 or over".

Grey0202:
Fix#1686

Revise English number patterns. Now it can extract "exceeding 150" "100 or over" "up to 700" etc.

Find a bug (issue #1692). It will be fixed in a future PR as a proper solution will take longer.


Patterns/English/English-Numbers.yaml
```diff
@@ -152,9 +152,9 @@ MoreOrEqual: !nestedRegex
   def: (?:({MoreRegex}\s+(or)?\s+{EqualRegex})|({EqualRegex}\s+(or)?\s+{MoreRegex})|{MoreOrEqualPrefix}(\s+(or)?\s+{EqualRegex})?|({EqualRegex}\s+(or)?\s+)?{MoreOrEqualPrefix}|>\s*=)
   references: [ MoreRegex, EqualRegex, LessRegex, MoreOrEqualPrefix ]
 MoreOrEqualSuffix: !simpleRegex
-  def: ((and|or)\s+(more|greater|higher|larger|bigger)((?!\s+than)|(\s+than(?!(\s*\d+)))))
+  def: ((and|or)\s+(((more|greater|higher|larger|bigger)((?!\s+than)|(\s+than(?!(\s*\d+)))))|((over|above)(?!\s+than))))
```
wgx998877: why not move "over and above" to the previous group?

PR #1697, opened 2019-07-09
[DateTime] Fix wrong recognition for "12:30 December 27"

songwenhao1:
Fix cases like "11:30 am to 12:30 december 27th", "30 December" would be wrong recognized
Also fix cases like "$12 december 27th"


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -35,7 +35,7 @@ public static class DateTimeDefinitions
       public const string CenturySuffixRegex = @"(^century)\b";
       public const string ReferencePrefixRegex = @"(that|same)\b";
       public const string FutureSuffixRegex = @"\b(in\s+the\s+)?(future|hence)\b";
-      public const string DayRegex = @"(the\s*)?(?<day>(?:3[0-1]|[1-2]\d|0?[1-9])(?:th|nd|rd|st)?)(?=\b|t)";
+      public const string DayRegex = @"(the\s*)?(?<![:$]\s*)(?<=\b)(?<day>(?:3[0-1]|[1-2]\d|0?[1-9])(?:th|nd|rd|st)?)(?=\b|t)";
```
wgx998877: add a limitation before :, the prefix token should not be "<digital>:...."

PR #1688, opened 2019-07-01
[.NET] Italian Date support

LionbridgeCS:
DateExtractor 107 pass 1 skipped. 
DateParser 89 pass 1 skipped.

Skipped case is "il 10 Settembre, millenovecentouno" partially recognized to "il 10 Settembre"

PR #1706, opened 2019-07-12
[.NET] Italian Holiday support

LionbridgeCS:
HolidayExtractor: All pass. 
HolidayParser: All pass.

***********************************************************************************
Draft as some parts of the YAML are under review in #1688.


.NET/Microsoft.Recognizers.Text.DateTime/Italian/Parsers/ItalianHolidayParserConfiguration.cs
```diff
@@ -12,27 +13,33 @@ public class ItalianHolidayParserConfiguration : BaseHolidayParserConfiguration
         public ItalianHolidayParserConfiguration(IOptionsConfiguration config)
             : base(config)
         {
+            ThisPrefixRegex = new Regex(DateTimeDefinitions.ThisPrefixRegex, RegexOptions.Singleline);
```
tellarin: Please replace the regex options with RegexFlags like in PR #1688.

PR #1710, opened 2019-07-15
[.NET] Italian DateTime Time support

LionbridgeCS:
TimeExtractor: all pass. 
TimeParser: 1 skipped. 

 -'Tornerò alle cinque e mezza meno tre minuti '  (I'll be back three minutes to half past five), the pattern seems not supported by the Extractor (it fails also in English), also it is an unusual way to express time.


.NET/Microsoft.Recognizers.Text.DateTime/Italian/Parsers/ItalianTimeParserConfiguration.cs
```diff
@@ -36,16 +42,16 @@ public void AdjustByPrefix(string prefix, ref int hour, ref int min, ref bool ha
             var deltaMin = 0;
             var trimmedPrefix = prefix.Trim();
 
-            // c'este 8 heures et demie, - "it's half past 8"
-            if (trimmedPrefix.EndsWith("demie"))
+            // "it's half past 8"
+            if (trimmedPrefix.EndsWith("mezza") || trimmedPrefix.EndsWith("mezzo"))
```
tellarin: Ideally these hardcoded words should be moved to YAML. But this is not the highest priority now.
Looks OK to merge once conflicts are resolved.

PR #1711, opened 2019-07-15
[.NET] Turkish Numbers (Percent) support

LionbridgeCS:
Follow up to #1653. 
9 passes 2 skips.

Issues: 
- written negative percentage, "-5%" actual "5%"
"minus five" is captured by BaseNumbers, but in Turkish, the order is different "eksi yüzde beş" directly translates to "minus percent five" since minus sign does not come before the number directly but comes before the number with percent,  it is not resolved. 

Explanation of modified localization: 
NumberWithSuffixPercentage can only be used with a (Num+Suffix)(Percent+Suffix), e.g. 210'un yüzdesi (per cent of 210). However, "yüzde 210" has the same meaning and more common than "210'un yüzdesi". Also, "yüzdesi" is mostly used as "percentage" and used with nouns more frequently than numbers, e.g. "kütle yüzdesi" (meaning "mass percentage").


LionbridgeCS:
Hello @tellarin, this is ready for review.

PR #1050, opened 2018-12-07
Move DataTypes.DataDrivenTests to .Net Core

hansmbakker:
Modernize the `Microsoft.Recognizers.Text.DataTypes.DataDrivenTests` and `Microsoft.Recognizers.Text.DataDrivenTests`.

PR #1714, opened 2019-07-16
Changing to SortedDictionary in NumberWithUnit

STayinloves:
Fix issues mentioned in #1050

- minor: revise python shell build script
- fix ambiguous YAML declarations


STayinloves:
Close for a better solution as discussed offline.

PR #1715, opened 2019-07-16
[DateTime] Modify the begin date of some cases like  "within next week"

songwenhao1:
The timex of "within next week" should be same as "next week".
Also fix "within next month" "within next year"
Support in JAVA, .NET

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -13215,5 +13215,155 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I will handle all the stuff within next week.",
+    "Context": {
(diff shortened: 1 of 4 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -13215,5 +13215,155 @@
...
+              "start": "2019-07-16",
+              "end": "2019-07-23"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "I will handle all the stuff in next two weeks.",
```
tellarin: Does this pr also supports "I will handle all the stuff within the next two weeks." ? Can you add that as a spec case?

PR #1718, opened 2019-07-17
[DateTime] fix incorrectly strips "next" from utterance

songwenhao1:
fix cases like "next week later in the afternoon","week later in the afternoon" should not be recognized.


Patterns/English/English-DateTime.yaml
```diff
@@ -530,8 +530,8 @@ AroundRegex: !simpleRegex
 AgoRegex: !simpleRegex
   def: \b(ago|before\s+(?<day>yesterday|today))\b
 LaterRegex: !nestedRegex
-  def: \b(?:later(?!(\s+in)?\s*{OneWordPeriodRegex})|from now|(from|after) (?<day>tomorrow|tmr|today))\b
-  references: [ OneWordPeriodRegex ]
+  def: \b(?:later(?!((\s+in)?\s*{OneWordPeriodRegex})|(\s+{TimeOfDayRegex}))|from now|(from|after) (?<day>tomorrow|tmr|today))\b
```
tellarin: I'm not sure this regex is the right place for this change.
Time of day has nothing to do with its purpose, which is to represent future times.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -13215,5 +13215,125 @@
         }
       }
     ]
+  },
+  {
+    "Input": "next week later in the afternoon.",
```
tellarin: Please write a full sentence in specs.
Maybe something like: "I'm not available so early. Let's go for coffee next week later in the afternoon."

tellarin: BTW, "next week, later in the afternoon" should also be supported. Can you add a spec for that?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -13215,5 +13215,125 @@
...
+              "timex": "2019-W30",
+              "type": "daterange",
+              "start": "2019-07-22",
+              "end": "2019-07-29"
+            }
+          ]
+        }
+      },
+      {
+        "Text": "in the afternoon",
```
tellarin: This is incorrect, the correct extraction would be "later in the afternoon", and its resolution should include the relevant MOD attribute.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -13215,5 +13215,125 @@
...
+              "timex": "2019-W30",
+              "type": "daterange",
+              "start": "2019-07-22",
+              "end": "2019-07-29"
+            }
+          ]
+        }
+      },
+      {
+        "Text": "in the morning",
```
tellarin: Same.

tellarin: BTW, do we support "late in the morning too"? I think we should.

Patterns/English/English-DateTime.yaml
songwenhao1: Yes， but the issue is caused by "week later in the afternoon" was wrong extracted by "DateExtract". And this "laterRegex" is used in AgoLaterUtil, not used in cases like "later in this week", so we can see behind the "laterRegex"  can not be "OneWordPeriodRegex" which is "dateRange". I think also should limit "timeRange", so i add "timeOfDay" in the limit

Specs/DateTime/English/DateTimeModel.json
songwenhao1: We don't support "later in the afternoon" now.. we support "later in the day/week/month/year" and "in the afternoon"..

tellarin: So this needs to be supported before this PR actually solves the problem it should solve.

Specs/DateTime/English/DateTimeModel.json
tellarin: @songwenhao1, could you update this PR now that there is support for "later in the morning"?

PR #1719, opened 2019-07-17
[.NET] Italian DateTime TimePeriod support

LionbridgeCS:
All tests passing for the type. 

We changed the test case "organizza una riunione dalle 1:30p.m. alle 4 persone" (set up a meeting from 1:30p.m. to 4 people) to "organizza una riunione dalle 2 alle 4 persone" (set up a meeting from 2 to 4 people) for a couple of reasons: 
-we had to relax the restriction that checks for false positive after "to" because in Italian we use the same word "alle" for both "to" and "at" (so there were problems with other patterns)
-the expression "dalle 1:30p.m. alle 4 persone" does not make sense while "dalle 2 alle 4 persone" is a true false positive.


Patterns/Italian/Italian-DateTime.yaml
```diff
@@ -428,7 +429,7 @@ UnspecificEndOfRegex: !simpleRegex
 UnspecificEndOfRangeRegex: !simpleRegex
   def: \b(fine\s*(dell')?anno)\b
 PeriodTimeOfDayRegex: !simpleRegex
-  def: (?<timeOfDay>(((il|la|in|a(l(la)?)?|nel(la)?)\s+)?((?<early>mattin[ao]\s+presto|di\s+prima\s+mattina|pomeriggio\s+presto|primo\s+pomeriggio|sera\s+presto|prima\s+serata)|(?<late>tarda\s+mattinata|mattin[ao]\s+tardi|tardo\s+pomeriggio|pomeriggio\s+inoltrato|pomeriggio\s+tardi|tarda\s+sera|sera\s+tardi|tarda\s+notte|notte\s+fonda))))\b
+  def: \b(((il|la|in|a(l(la)?)?|nel(la)?|di)\s+)?(((mattin[ao]|mattinata|pomeriggio|sera|serata|notte)(?<early>\s+presto)|(?<early>prim[ao]\s+)(mattin[ao]|mattinata|pomeriggio|sera|serata|notte))|((mattin[ao]|mattinata|pomeriggio|sera|serata|notte)(?<late>\s+(tardi|inoltrat[oa]|fonda))|(?<late>tard[oa]\s+)(mattin[ao]|mattinata|pomeriggio|sera|serata|notte))))\b
```
tellarin: Are you sure the usage of the named groups here is correct and match English? Otherwise this will lead to unexpected side effects.

LionbridgeCS: @tellarin Regex is reviewed.

PR #1720, opened 2019-07-17
[DateTime] Support "later in the afternoon"

songwenhao1:
Support "later in the afternoon" same as "late afternoon"

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -13231,6 +13231,7 @@
           "values": [
             {
               "timex": "TAF",
+              "Mod": "end",
               "type": "timerange",
               "start": "14:00:00",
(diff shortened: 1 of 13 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -13215,5 +13215,101 @@
...
+    "Results": [
+      {
+        "Text": "late afternoon",
+        "Start": 17,
+        "End": 30,
+        "TypeName": "datetimeV2.timerange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "TAF",
```
tellarin: Resolutions for such entities need to contain the corresponding MOD attributes.
Are they not being assigned? Or is the test code just not checking them? Both need to be fixed.

PR #1721, opened 2019-07-17
[.NET] Italian DateTime Duration support

LionbridgeCS:
All test cases pass.

### DIFF
--- a/Specs/DateTime/Italian/DurationParser.json
+++ b/Specs/DateTime/Italian/DurationParser.json
@@ -480,6 +480,32 @@
     ]
   },
   {
+    "Input": "Andrò via per l'intero weekend",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "Comment": "whole weekend/ entire weekend is not supported in English either",
+    "NotSupported": "dotnet",
+    "NotSupportedByDesign": "python,javascript,python",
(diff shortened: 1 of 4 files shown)


Specs/DateTime/Italian/DurationParser.json
```diff
@@ -432,20 +449,21 @@
...
-        }
+        },
+        "Start": 17,
+        "Length": 13
       }
     ]
   },
   {
-    "Input": "Andrò via per l'intero weekend",
+    "Input": "Andrò via per l'intera settimana",
```
tellarin: "l'intero weekend" is not a common expression even if informal?
If it's somewhat common, we should support it.

LionbridgeCS: 'l'intero weekend' (the whole weekend) is not supported in English either. 
Also this test case seems a bad translation of the case 'I'll leave for full week' since the expected Timex refers to a week duration.
Should we add a test case for l'intero weekend and skip it?

tellarin: Please do. Thanks.

tellarin:
@LionbridgeCS can you fix the merge conflicts?

LionbridgeCS:
@tellarin, sure thing we will merge and make sure the build passes then update. Thank you!

PR #1724, opened 2019-07-18
Fix conflicts in NumberWithUnits resource files 

STayinloves:
The purpose of this PR is to remove the conflicts existing in `NumberWithUnits` resource files.

- [x] Bulgarian
- [x] Chinese
- [x] Dutch
- [x] English
- [x] French
- [x] German
- [x] Italian
- [x] Japanese
- [x] Korean
- [x] Portuguese
- [x] Spanish
- [x] Swedish
- [x] Turkish


.NET/Microsoft.Recognizers.Text.NumberWithUnit/Utilities/DictionaryUtils.cs
```diff
@@ -31,12 +32,21 @@ public static void BindUnitsString(IDictionary<string, string> sourceDictionary,
...
 
-                sourceDictionary.Add(token, key);
+                // This segment of code is used to prevent duplicated key-values in resource files.
+                try
+                {
+                    sourceDictionary.Add(token, key);
+                }
+                catch (ArgumentException exception)
+                {
+                    // Temporarily catch exception to show where errors occur, should be removed afterward.
```
tellarin: Shouldn't we remove this now, so the build breaks on tests if there are duplicate keys?

Patterns/French/French-NumbersWithUnit.yaml
```diff
@@ -379,12 +379,10 @@ CurrencyPrefixList: !dictionary
      Réal brésilien: r$|brl|reais
      Guaraní paraguayen: ₲|gs.|pyg
      Dollar trinidadien: ttd|titis
-     Yuan renminbi: cny|rmb|¥|元
-     Yen: ¥|jpy
+     Yuan Chinois: cny|rmb|￥|元
```
tellarin: Why is the ¥ not causing issues? Or is this the pending issue and the reason for the exception catch?

Patterns/Italian/Italian-NumbersWithUnit.yaml
```diff
@@ -228,10 +227,10 @@ CurrencySuffixList: !dictionary
     Manat: manats|manat
     Qəpik: qəpik
 #Shilling
-    Somali shilling: scellino somalo|scellini somali|scellino della somalia|scellini della somalia|scellino|shilin|sh.so.|sos
+    Somali shilling: scellino somalo|scellini somali|scellino della somalia|scellini della somalia|shilin|sh.so.|sos
```
tellarin: "shilin" by itself should also move to the general Shiling currency.

Patterns/French/French-NumbersWithUnit.yaml
tellarin: There are three separate unicode characters for it, but all can be used for both yen and yuan.

STayinloves: For now, the PR is keeping the full-width symbol for Chinese and the half-width symbol for Japanese (to avoid conflicts), that's what the most IMEs would do [[reference]](https://en.wikipedia.org/wiki/Yen_sign).

But we do need to find out a better solution.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Utilities/DictionaryUtils.cs
STayinloves: Yes we should.

Patterns/French/French-NumbersWithUnit.yaml
tellarin: I understand the IME case, but in languages like English, I always use the half-width one for yuan. :-)

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestContextExtensions.cs
```diff
@@ -714,9 +714,9 @@ public static IDateTimeExtractor GetItalianExtractor(DateTimeExtractors extracto
                 case DateTimeExtractors.Set:
                     return new BaseSetExtractor(new ItalianSetExtractorConfiguration(config));
                 case DateTimeExtractors.Merged:
-                    return new BaseMergedDateTimeExtractor(new ItalianMergedExtractorConfiguration(config));
+                    return new BaseMergedDateTimeExtractor(new ItalianMergedExtractorConfiguration(DateTimeOptions.None));
```
tellarin: Why wouldn't this one use config?

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -96,6 +97,13 @@ protected override void InitializeConfiguration()
                         new GermanMergedParserConfiguration(new BaseOptionsConfiguration(options))),
                     new BaseMergedDateTimeExtractor(new GermanMergedExtractorConfiguration(options))));
 
+            RegisterModel<DateTimeModel>(
```
tellarin: The Italian DateTime model should not be registered as it is not ready. Only the parser/extractor tests should run by default now, until the whole model can be released.

PR #1736, opened 2019-07-24
[.NET] DateTime suggested solution for BaseDateTimePeriodParser

LionbridgeCS:
The modification in ParseTimePeriod serves to correctly address cases such as "on Dec 9th between 8am and 2pm" so that both particles 'am' and 'pm' are extracted and associated to the corresponding time entities (originally only one particle was extracted and associated to both times causing errors in cases where the two particles differ)

The modification in MergeTwoTimePoints serves to take into account minutes when calculating the duration so that for example "from half past seven to 4pm of tomorrow" returns 'PT8H30M' (the original output was 'PT8H').
Spanish case is an example of this point it was returning PT8H instead of the correct PT8H30M

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
@@ -732,42 +732,27 @@ namespace Microsoft.Recognizers.Text.DateTime
             }
 
             // Parse "pm"
-            string matchPmStr = string.Empty, matchAmStr = string.Empty, descStr = string.Empty;
-            string matchPmStr1 = string.Empty, matchAmStr1 = string.Empty, descStr1 = string.Empty;
-            var matchPmStrGroup = match.Groups[Constants.PmGroupName];
-            var matchAmStrGroup = match.Groups[Constants.AmGroupName];
+            string descStr = string.Empty, beginDescStr = string.Empty, endDescStr = string.Empty;
+            var matchPmStr = match.Groups[Constants.PmGroupName].Value;
+            var matchAmStr = match.Groups[Constants.AmGroupName].Value;
             var descStrGroup = match.Groups[Constants.DescGroupName];
-            if (matchPmStrGroup.Captures.Count > 0)
+
+            if (descStrGroup.Captures.Count > 1)
             {
-                matchPmStr = matchPmStrGroup.Captures[0].Value;
-                if (matchPmStrGroup.Captures.Count > 1)
-                {
-                    matchPmStr1 = matchPmStrGroup.Captures[1].Value;
-                }
+                // first particle am/pm in patterns like "between 8am and 2pm"
+                beginDescStr = descStrGroup.Captures[0].Value;
+
+                // second particle am/pm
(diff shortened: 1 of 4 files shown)


LionbridgeCS:
@tellarin, the intermittent issue happened again here. if the build fails we will close and reopen this tomorrow.

tellarin:
The build failure is not due to the intermittent Java dependency problem.
One of the Python tests now fails as the Python code was not updated. Can you update it? Or mark the test as NotSupported in Python and add a comment for the reason with the PR# that has the corresponding C# fix.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
```diff
@@ -732,36 +732,119 @@ private string ParseTimePeriod(Match match, out int beginHour, out int endHour)
...
-            if (!string.IsNullOrEmpty(matchAmStr) || (!string.IsNullOrEmpty(descStr) && descStr.StartsWith("a")))
+            string matchPmStr = string.Empty, matchAmStr = string.Empty, descStr = string.Empty;
+            string matchPmStr1 = string.Empty, matchAmStr1 = string.Empty, descStr1 = string.Empty;
+            var matchPmStrGroup = match.Groups[Constants.PmGroupName];
+            var matchAmStrGroup = match.Groups[Constants.AmGroupName];
+            var descStrGroup = match.Groups[Constants.DescGroupName];
+            if (matchPmStrGroup.Captures.Count > 0)
             {
-                if (beginHour >= Constants.HalfDayHourCount)
+                matchPmStr = matchPmStrGroup.Captures[0].Value;
```
tellarin: It is better to change the patterns to use named capture groups. Using indices can lead to issues as different patterns are edited.
Also, can you give better names to variables than matchPmStr1? What's the key difference from matchPmStr?
Lastly, why is this change not necessary in English? Doesn't English have corresponding test specs? IF not, please add one sentence to demonstrate this new code also works there, ok?

LionbridgeCS:
@tellarin, we marked the case for now (to make the build pass); we are looking into the feedback and will update accordingly.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
```diff
@@ -732,42 +732,27 @@ private string ParseTimePeriod(Match match, out int beginHour, out int endHour)
...
+            if (descStrGroup.Captures.Count > 1)
             {
-                matchAmStr = matchAmStrGroup.Captures[0].Value;
-                if (matchAmStrGroup.Captures.Count > 1)
-                {
-                    matchAmStr1 = matchAmStrGroup.Captures[1].Value;
-                }
-            }
+                // first particle am/pm in patterns like "between 8am and 2pm"
+                beginDescStr = descStrGroup.Captures[0].Value;
```
tellarin: This is still using capture index groups. Why not just use the new names in the modified regex?

LionbridgeCS:
@tellarin, this is updated. the Python code will take some time so we marked the case as not supported in Python.

PR #1737, opened 2019-07-24
[.NET] Italian DateTime Set support

LionbridgeCS:
SetExtractor: 25 pass. 
SetParser: 27 pass. 

3 moved to DateTimeExtractor.
2 moved to DateTimeParser. 

***********************************************************************
work in progress because of the SetWeekDay change

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
@@ -136,22 +136,30 @@ namespace Microsoft.Recognizers.Text.DateTime
             {
                 if (match.Success)
                 {
-                    var trimmedText = text.Remove(match.Index, match.Length);
-                    trimmedText = trimmedText.Insert(match.Index, match.Groups["weekday"].ToString());
-
-                    var ers = extractor.Extract(trimmedText, reference);
-                    foreach (var er in ers)
+                    if (match.Value.EndsWith("s"))
                     {
-                        if (er.Start <= match.Index && er.Text.Contains(match.Groups["weekday"].Value))
-                        {
-                            var len = (er.Length ?? 0) + 1;
-                            if (match.Groups[Constants.PrefixGroupName].ToString() != string.Empty)
-                            {
-                                len += match.Groups[Constants.PrefixGroupName].ToString().Length;
-                            }
+                        var trimmedText = text.Remove(match.Index, match.Length);
+                        trimmedText = trimmedText.Insert(match.Index, match.Groups["weekday"].ToString());
 
-                            ret.Add(new Token(er.Start ?? 0, er.Start + len ?? 0));
+                        var ers = extractor.Extract(trimmedText, reference);
+                        foreach (var er in ers)
+                        {
+                            if (er.Start <= match.Index && er.Text.Contains(match.Groups["weekday"].Value))
+                            {
+                                var len = (er.Length ?? 0) + 1;
+                                if (match.Groups[Constants.PrefixGroupName].ToString() != string.Empty)
+                                {
+                                    len += match.Groups[Constants.PrefixGroupName].ToString().Length;
+                                }
+
+                                ret.Add(new Token(er.Start ?? 0, er.Start + len ?? 0));
+                            }
                         }
+                    }
+                    else
+                    {
+                        // this function is used to extract SetWeekDay when the plural is not formed adding 's'
+                        config.SetWeekDayExtractor(extractor, text, match, reference, ref ret);
                     }
                 }
             }
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
@@ -1,5 +1,7 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Text.RegularExpressions;
+
 using DateObject = System.DateTime;
 
 namespace Microsoft.Recognizers.Text.DateTime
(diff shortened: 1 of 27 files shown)


LionbridgeCS:
@tellarin, there are many skipped cases because one of the responsible pattern is ambiguous in Italian. above is an explanation.

tellarin:
@LionbridgeCS I think it's fine to always consider such cases as regular DateTimes (i.e., not Set). "la mattina alle 9 faccio colazione" -> 9 in the morning looks OK. Move such spec tests to the corresponding correct type spec.
Having said that, can some of these cases be re-written to make them explicit Sets? Clearly say: "I have breakfast every 9 in the morning" or "I don't have breakfast on Mondays", for example.

LionbridgeCS:
Thank you @tellarin, noted. we will modify the concerned cases.

LionbridgeCS:
@tellarin, in the last update: We introduced the two functions SetWeekDayExtractor and SetWeekDayParser because the original code to handle SetWeekDay cases only supported languages where the plural is formed by adding an 's' at the end of the word (in Italian instead plural is formed by changing the word's final character e.g. mattina -> mattine (morning -> mornings))

tellarin:
@LionbridgeCS are 0 tests skipped/failing now?
We'll review the latest changes ASAP.

LionbridgeCS:
@tellarin, yes all are passing with this change.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -221,12 +221,20 @@ private DateTimeResolutionResult ParseEach(IDateTimeExtractor extractor, IDateTi
             match = this.config.SetWeekDayRegex.Match(text);
             if (match.Success)
             {
-                var trimmedText = text.Remove(match.Index, match.Length);
-                trimmedText = trimmedText.Insert(match.Index, match.Groups["weekday"].ToString());
-                ers = extractor.Extract(trimmedText, refDate);
-                if (ers.Count == 1 && ers.First().Length == trimmedText.Length)
+                if (match.Value.EndsWith("s"))
```
tellarin: Instead of this if and using inheritance, it might be better to have all configs use composition and internally have a class to handle set extract and set parse. This one one base class can be used by default (with the contents of this if) and Italian can instantiate its own.
Then this piece of this method is just a call to config.SetWeekDayParser(). Similarly for the extractor.
You can remove the Draft mark after that.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
```diff
@@ -136,23 +136,31 @@ public List<Token> MatchEach(IDateTimeExtractor extractor, string text, DateObje
...
 
-                            ret.Add(new Token(er.Start ?? 0, er.Start + len ?? 0));
+                                ret.Add(new Token(er.Start ?? 0, er.Start + len ?? 0));
+                            }
                         }
                     }
+                    else
+                    {
+                        // this function is used to extract SetWeekDay when the plural is not formed adding 's'
+                        config.SetWeekDayExtractor(extractor, text, match, reference, ref ret);
```
tellarin: These methods should return the extracted/parsed entity.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/ISetExtractorConfiguration.cs
```diff
@@ -33,5 +35,7 @@ public interface ISetExtractorConfiguration : IOptionsConfiguration
         IDateTimeExtractor TimePeriodExtractor { get; }
 
         IDateTimeExtractor DateTimePeriodExtractor { get; }
+
+        void SetWeekDayExtractor(IDateTimeExtractor extractor, string text, Match match, DateObject reference, ref List<Token> ret);
```
tellarin: Better to use composition instead of inheritance, to avoid most cultures having empty methods.

.NET/Microsoft.Recognizers.Text.DateTime/Italian/Extractors/ItalianSetExtractorConfiguration.cs
```diff
@@ -75,5 +78,60 @@ public ItalianSetExtractorConfiguration(IOptionsConfiguration config)
         Regex ISetExtractorConfiguration.SetWeekDayRegex => SetWeekDayRegex;
 
         Regex ISetExtractorConfiguration.SetEachRegex => SetEachRegex;
+
+         // for SetWeekDay when the plural is not formed by adding 's'
+        public void SetWeekDayExtractor(IDateTimeExtractor extractor, string text, Match match, DateObject reference, ref List<Token> ret)
```
tellarin: Please return the extracted Tokens and add to the list in the calling code instead of use ref.

.NET/Microsoft.Recognizers.Text.DateTime/Italian/Parsers/ItalianSetParserConfiguration.cs
```diff
@@ -141,5 +144,43 @@ public bool GetMatchedUnitTimex(string text, out string timex)
 
             return true;
         }
+
+         // for SetWeekDay when the plural is not formed by adding 's'
+        public void SetWeekDayParser(IDateTimeExtractor extractor, string text, Match match, DateObject reference, ref List<ExtractResult> ers, ref bool success)
```
tellarin: Same here. If nothing is parsed, it can return an empty list. The ```success``` condition can be in the calling code.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
tellarin: Perhaps call the class with both methods something like SetHandler.

LionbridgeCS:
thank you @tellarin, we are reviewing this and will update accordingly.

LionbridgeCS:
@tellarin, we updated the proposed solution. Instead of introducing the Extractor and Parser functions, we kept the original structure in the BaseExtractor/Parser and used new functions only for the weekday group matching (which is the part that really need to be modified).

PR #1738, opened 2019-07-25
[.NET] Revise Chinese date parser (#1731)

### LINKED ISSUE #1731: [CN DateTimeV2] Wrong duration of extracted date range in Chinese
**Describe the bug**
计费周期 2019年6月1日-6月30日

**Expected result**
{
  "Text": "2019年6月1日-6月30日",
  "Start": 5,
  "End": 19,
  "TypeName": "datetimeV2.daterange",
  "Resolution": {
    "values": [
      {
        **"timex": "(2019-06-01,XXXX-06-30,P30D)",**
        "type": "daterange",
        "start": "2019-06-01",
        "end": "2019-06-30"
      },
      {
        "timex": "(2019-06-01,XXXX-06-30,P395D)",
        "type": "daterange",
        "start": "2019-06-01",
        "end": "2020-06-30"
      }
    ]
  }
}

**Actual result**
{
  "Text": "2019年6月1日-6月30日",
  "Start": 5,
  "End": 19,
  "TypeName": "datetimeV2.daterange",
  "Resolution": {
    "values": [
      {
        **"timex": "(2019-06-01,XXXX-06-30,P395D)",**
        "type": "daterange",
        "start": "2019-06-01",
        "end": "2019-06-30"
      },
      {
        "timex": "(2019-06-01,XXXX-06-30,P395D)",
        "type": "daterange",
        "start": "2019-06-01",
        "end": "2020-06-30"
      }
    ]
  }
}

Grey0202:
This fixes #1731.

Now it can parse "2019年6月1日-6月30日" as  "2019-06-01 to 2019-06-30".

Now when parsing "From last Friday to Oct 10", we can get an integer result.

Rename some Chinese & Japanese regex.


.NET/Microsoft.Recognizers.Definitions.Common/Japanese/DateTimeDefinitions.cs
```diff
@@ -47,6 +47,7 @@ public static class DateTimeDefinitions
       public const string DateThisRe = @"这个|这一个|这|这一|本|今";
       public const string DateLastRe = @"上个|上一个|上|上一|去";
       public const string DateNextRe = @"下个|下一个|下|下一|明";
```
tellarin: Ideally we should rename all these regexes to use complete names. No need for such abbreviations, which are also inconsistent with others.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDateExtractorConfiguration.cs
```diff
@@ -118,6 +124,72 @@ public List<ExtractResult> Extract(string text, DateObject referenceTime)
             return Token.MergeAllTokens(tokens, text, ExtractorName);
         }
 
+        public int GetYearFromText(Match match)
```
tellarin: This method would be the same as in Japanese, right?
Why not have it in a base CJK class and just call from here passing the right parser config?

tellarin: Is it also the same as in other languages? If so, we should decide where to place it for reuse.

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -3240,5 +3240,77 @@
         }
       }
     ]
+  },
+  {
+    "Input": "18年10月1日到19年2月3日",
```
tellarin: Can you add at least one equivalent Japanese test case too? Even if not implemented yet, it's good to have the spec to test the future implementation.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -1409,5 +1419,50 @@ private int ToMonthNumber(string monthStr)
         {
             return this.config.MonthOfYear[monthStr] > 12 ? this.config.MonthOfYear[monthStr] % 12 : this.config.MonthOfYear[monthStr];
         }
+
+        private DateContext GetYearContext(string startDateStr, string endDateStr, string text)
```
tellarin: Isn't this implementation still the same?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/AbstractYearExtractor.cs
```diff
@@ -6,6 +6,11 @@ namespace Microsoft.Recognizers.Text.DateTime
 {
     public abstract class AbstractYearExtractor : IDateExtractor
     {
+
+        public AbstractYearExtractor()
```
tellarin: Why do we need this new constructor?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -869,9 +870,18 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
 
             var pr1 = this.config.DateParser.Parse(er[0], referenceDate);
             var pr2 = this.config.DateParser.Parse(er[1], referenceDate);
-            if (pr1.Value == null || pr2.Value == null)
+
+            if (er.Count >= 2)
             {
-                return ret;
+                var dateContext = BaseDatePeriodParser.GetYearContext(config.DatePeriodYearRegex, config.RelativeRegex, new ChineseDateExtractorConfiguration(), er[0].Text, er[1].Text, text);
```
STayinloves: No need to creates a new instance of `ChineseDateExtractorConfiguration`, just use `this.SingleDateExtractor`.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -1409,5 +1419,50 @@ private int ToMonthNumber(string monthStr)
...
+
+            var relativeMatchForStartDate = this.config.RelativeRegex.Match(startDateStr);
+            var relativeMatchForEndDate = this.config.RelativeRegex.Match(endDateStr);
+            isDateRelative = relativeMatchForStartDate.Success || relativeMatchForEndDate.Success;
+
+            if (!isEndDatePureYear && !isDateRelative)
+            {
+                foreach (Match match in config.DatePeriodYearRegex.Matches(text))
+                {
+                    var year = new ChineseDateExtractorConfiguration().GetYearFromText(match);
```
STayinloves: Same.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_German.cs
```diff
@@ -203,11 +203,13 @@ public void DateTimeModel(TestModel testSpec)
             TestDateTime(testSpec);
         }
 
+        /*
```
tellarin: Why this?

.NET/Microsoft.Recognizers.Text.DataDrivenTests/NetCoreTestDataSourceAttribute.cs
```diff
@@ -30,15 +30,6 @@ public IEnumerable<object[]> GetData(MethodInfo methodInfo)
             var rawData = File.ReadAllText(specsFile);
             var specs = JsonConvert.DeserializeObject<IList<TestModel>>(rawData);
 
-            // Adding a pseudo test case to avoid test failure causing by returning empty sequence.
```
tellarin: Please don't undo other people's changes unless they are part of the PR content.

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishDatePeriodParserConfiguration.cs
```diff
@@ -191,7 +191,7 @@ public SpanishDatePeriodParserConfiguration(ICommonDateTimeParserConfiguration c
 
         Regex IDatePeriodParserConfiguration.ThisPrefixRegex => ThisPrefixRegex;
 
-        Regex IDatePeriodParserConfiguration.RelativeRegex => RelativeRegex;
+        Regex ISimpleDatePeriodParserConfiguration.RelativeRegex => RelativeRegex;
```
tellarin: You should not change the interfaces everywhere. Especially mixing types. Please keep the original ones.

tellarin: My understanding from your code is that ISimpleDatePeriodParserConfiguration defines just the common interface needed by your new method. If that's the case, you can make the other interfaces that should fir there to inherit from this simple interface.

Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -13,7 +13,7 @@ MonthNumRegex: !simpleRegex
 TwoNumYear: '50'
 YearNumRegex: !simpleRegex
   def: (?<year>((1[5-9]|20)\d{2})|2100)
-YearRegex: !simpleRegex
+SimpleYearRegex: !simpleRegex
```
tellarin: Why do you have this rename and other changes in Chinese and not in Japanese? Don't they follow the same design?

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishDatePeriodParserConfiguration.cs
tellarin: I see now this is just specifying the property declaration. Please ignore this comment.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -869,9 +870,18 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
 
             var pr1 = this.config.DateParser.Parse(er[0], referenceDate);
             var pr2 = this.config.DateParser.Parse(er[1], referenceDate);
-            if (pr1.Value == null || pr2.Value == null)
+
+            if (er.Count >= 2)
             {
-                return ret;
+                var dateContext = BaseDatePeriodParser.GetYearContext(this.config, er[0].Text, er[1].Text, text);
```
tellarin: This is not ideal. In regular usage BaseDatePeriodParser already has access to a config of the right type. Only this use in Chinese is strange.
It does seem it would need a big refactoring though, so I think it's OK to add this for now with a note here and in the newly static method declaration.

Issue #776, opened 2018-08-08
[Infra] .NET test suite can't handle empty string/arrays


iMicknl:
The .NET test suite can't handle empty strings of the NotSupported property. This is not a big issue, however I think it should be able to handle empty strings during serialisation. 
Low priority.

```
  {
    "NotSupported": "",
  },
```

```
Message: Class Initialization method Microsoft.Recognizers.Text.Number.Tests.TestNumber_Dutch.ClassInitialize threw exception. Newtonsoft.Json.JsonSerializationException: Newtonsoft.Json.JsonSerializationException: Error converting value "" to type 'Microsoft.Recognizers.Text.DataDrivenTests.Platform'. Path '[0].NotSupported', line 4, position 22. ---> System.ArgumentException: Must specify valid information for parsing in the string..
```

Thanks for helping me figure this out, @STayinloves. 👍

Gerappa92:
Hello,

I have created PR #2806.
Review, please.

PR #1741, opened 2019-07-26
Address the failure causing by empty test specs

STayinloves:


iMicknl:
@STayinloves will this also solve #776?

STayinloves:
> @STayinloves will this also solve #776?

No likely. 😅

PR #1665, opened 2019-06-26
[.NET] Dutch Date partial support 

LionbridgeCS:
DateExtractor: 137 passing. Failures are mostly related to partial matches and unrecognized patterns. 
DateParser: 143 passing the skipped are related to partial matches and to to-be modified Regexes. 

Example of partial match: 
 -"woens, 22e van jan". Actual:<22e van jan>

PR #1743, opened 2019-07-26
[.NET] Turkish DateTime (Date) support -Initial-

LionbridgeCS:
Turkish DateTime base code and Date tests activated.
Work in progress, we will keep updating this PR. 
Reported stats: 
DateExtractor : pass: 52 | fail: 4
DateParser:  pass: 23 | fail: 69

Explanation of the fails in Extractor: 
- "Yarından itibaren üç hafta uygun musun?" => The English version is "three weeks from tomorrow" but in Turkish the order is different. "yarından itibaren" means "from tomorrow" and "üç hafta" means "three week" 
- "3 hafta içinde ayrılacağım" => Similar to the above"3 hafta içinde" = "3 week in" ("in three weeks" in the English version). 
The other inputs have 27 extracted while it shouldn't

Turkish specific test cases will be added with the implementation.


LionbridgeCS:
@tellarin, on a somewhat related note, we have Turkish Holiday complete. Should we create a PR already for it? or wait until we reach the sybtype priority order?

tellarin:
@LionbridgeCS if Holiday extractor/parser doesn't depend on the other types, I'm fine with creating a PR for it first for review/merge.

tellarin:
@LionbridgeCS no update here? The PR now has conflicts with /master. Let's try to at least keep it fresh, ok?

LionbridgeCS:
@tellarin, we are working on the parser and as some of the work introduced errors in the extractor test so we are working on fixing.

LionbridgeCS:
@tellarin, we will update this PR tomorrow with the updates. can we merge this as is now -as an initial version- and make a new PR with the new changes using a different account? that would be easier for us.

tellarin:
Same as #1665

PR #1758, opened 2019-08-02
[EN DateTimeV2] Support to recognize "this or next week"

songwenhao1:


Specs/DateTime/English/DateTimeModelExtendedTypes.json
```diff
@@ -1685,5 +1685,48 @@
         }
       }
     ]
+  },
+  {
+    "Input": "Cortana could try to arrange a Skype call this or next week please?",
```
tellarin: Can you add this same test in DateTimeModel to show that there it only catches "next week" as expected;
and add the same case as in this file also to ComplexCalendar to show that parenttext is set correctly (ComplexCalendar includes enabling ExtendedTypes).
Thanks

PR #1765, opened 2019-08-02
[.NET][Java] Revise timezone

Grey0202:
- Add some Spain cities.
- Now it can get right result when cities end with "timezone" or "      time"


.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
```diff
@@ -580,7 +576,6 @@ public static class TimeZoneDefinitions
             { @"pacific", -420 },
             { @"west coast", -420 },
             { @"west coast time", -420 },
-            { @"west coast timezone", -420 },
```
tellarin: Why remove some entries for "* timezone" in the minMapping dict and not the ones for "time" too?
It now look very strange were things are inconsistent. For example, with the new city names ("malaga" time vs. '
"zaragoza").

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimeZoneParser.cs
```diff
@@ -69,6 +69,12 @@ public static string ConvertMinsToRegularFormat(int offsetMins)
             return TimeSpan.FromMinutes(offsetMins).ToString(@"hh\:mm");
         }
 
+        public static string NormalizeText(string text)
+        {
+            text = Regex.Replace(text, @"\s+", " ");
+            return Regex.Replace(text, @"timezone", "time");
```
tellarin: Why this new timezone->time normalization? Wouldn't it be better to actually just remove either word from the end of the match, trim, and then search in the dictionary?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimeZoneParser.cs
```diff
@@ -72,7 +72,9 @@ public static string ConvertMinsToRegularFormat(int offsetMins)
         public static string NormalizeText(string text)
         {
             text = Regex.Replace(text, @"\s+", " ");
-            return Regex.Replace(text, @"timezone", "time");
+            text = Regex.Replace(text, @"timezone$", string.Empty);
+            text = Regex.Replace(text, @"time$", string.Empty);
```
tellarin: All three replaces can be the same regex. It can also be a compiled regex for a little more speed.

Specs/DateTime/English/TimeZoneParser.json
```diff
@@ -718,7 +718,30 @@
           }
         },
         "Start": 11,
-        "Length": 26
+        "Length": 16
+      }
+    ]
+  },
+  {
+    "Input": "I'm in the Madrid ",
```
tellarin: "I'm in Madrid."

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimeZoneParser.cs
```diff
@@ -72,8 +74,7 @@ public static string ConvertMinsToRegularFormat(int offsetMins)
         public static string NormalizeText(string text)
         {
             text = Regex.Replace(text, @"\s+", " ");
```
tellarin: These two regexes can be the same.

Patterns/English/English-TimeZone.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 # Direct UTC Expression
 DirectUtcRegex: !simpleRegex
-  def: \b(utc|gmt)(\s*[+\-\u00B1]?\s*[\d]{1,2}(\s*:\s*[\d]{1,2})?)?\b
+  def: \b(utc|gmt|msk)(\s*[+\-\u00B1]?\s*[\d]{1,2}(\s*:\s*[\d]{1,2})?)?\b
```
tellarin: I don't think it's appropriate to mix gmt|utc behaviour with Russian-specific timezones. These are two semantically different concepts.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimeZoneParser.cs
```diff
@@ -14,7 +14,7 @@ public class BaseTimeZoneParser : IDateTimeParser
         public static readonly Regex TimeZoneEndRegex = new Regex("time$|timezone$", RegexOptions.Singleline);
 
         // Compute UTC offset in minutes from matched timezone offset in text. e.g. "-4:30" -> -270; "+8"-> 480.
-        public static int ComputeMinutes(string utcOffset)
+        public static int ComputeMinutes(string timeZone, string utcOffset)
```
tellarin: This method has a very clear intent to do simple calculations. I don't think passing timezone names into it is a good design.

PR #1766, opened 2019-08-02
[.NET] Italian DateTime Merged support

LionbridgeCSII:
All test cases are passing.
"2017-13-12" test case fixed according to the English counterpart


Patterns/Italian/Italian-DateTime.yaml
```diff
@@ -538,7 +537,10 @@ NightStartEndRegex: !simpleRegex
 InexactNumberRegex: !simpleRegex
   def: \b(un\s+po'\s+di|alcun[ei]|parecchie?|divers[ie]|molt[oaei]|poc[oa]|poch[ei])\b
 InexactNumberUnitRegex: !nestedRegex
-  def: \b({InexactNumberRegex}\s+{DurationUnitRegex}|(?<unit>ore|minuti|secondi))\b
+  def: \b((({InexactNumberRegex}\s+|(?<=(prossim[ei]|scors[ei]|ultim[ei])\s+)){DurationUnitRegex})|({DurationUnitRegex}(?=\s+(seguenti|successiv[ei]|passat[ei]|precedent[ei]))))\b
+  references: [InexactNumberRegex, DurationUnitRegex]
+InexactNumberUnitRegex2: !nestedRegex
```
tellarin: Why declare a new regex with 2 and not declare a new inner regex, but the regex name referenced in the rest of the code the same? For example, rename InexactNumberUnitRegex to InnerInexactNumberUnitRegex and make InexactNumberUnitRegex2 that uses it be the new InexactNumberUnitRegex? This way the outside "interface" for all cultures remains the same.
Or are both used in different circumstances and I didn't notice it in the review?

LionbridgeCSII: The two regex are used in different circumstances (InexactNumberUnitRegex in ItalianDurationExtractorConfiguration and InexactNumberUnitRegex2 in ItalianDurationParserConfiguration). In this way we can address patterns such as "in the next few hours" which in Italian is "nelle prossime ore" (in the next hours) where the InexactNumberRegex match "few" is missing. So in DurationExtractor we use lookarounds to check for the presence of elements like "next/previous" (otherwise many false positive would be extracted) and in DurationParser we match the unit "hours" alone.

tellarin: Can you add a small comment note on the YAML file mentioning the different usage cases?
Comments in YAML start with `#`. One of two sentences before each regex declaration would be enough.
Thanks!

LionbridgeCSII: @tellarin, updated

Specs/DateTime/Italian/MergedParser.json
```diff
@@ -767,39 +785,42 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "NotSupported": "dotnet",
-    "NotSupportedByDesign": "python,javascript,python",
+    "NotSupportedByDesign": "python,javascript,java",
     "Results": [
       {
-        "Text": "Pasqua",
+        "Text": "domenica",
```
tellarin: This one is incorrect, the extracted entity should be "domenica di Pasqua"

Specs/DateTime/Italian/MergedParser.json
```diff
@@ -1408,22 +1473,24 @@
...
+        "Type": "datetimeV2.daterange",
         "Value": {
           "values": [
             {
               "timex": "2020",
               "Mod": "before",
               "type": "daterange",
-              "end": "2020-01-01"
+              "end": "2020-01-01",
+			  "sourceEntity": "datetimerange"
```
tellarin: Nit: Indentation

Specs/DateTime/Italian/MergedParser.json
```diff
@@ -1432,22 +1499,24 @@
...
+        "Type": "datetimeV2.daterange",
         "Value": {
           "values": [
             {
               "timex": "2012",
               "Mod": "before",
               "type": "daterange",
-              "end": "2012-01-01"
+              "end": "2012-01-01",
+			  "sourceEntity": "datetimerange"
```
tellarin: Same

Specs/DateTime/Italian/MergedParser.json
```diff
@@ -1456,22 +1525,24 @@
...
+        "Type": "datetimeV2.daterange",
         "Value": {
           "values": [
             {
               "timex": "2012",
               "Mod": "since",
               "type": "daterange",
-              "start": "2012-01-01"
+              "start": "2012-01-01",
+			  "sourceEntity": "datetimerange"
```
tellarin: Same

PR #1767, opened 2019-08-02
[Javascript] ESLint --fix autofixed files

Aliandi:
### Description
Files with warnings thrown by the current ESLint rules are fixed here. 

### Changes made
The files were affected by the ESLint `--fix` option removing the warnings.
image
### Testing
Build and tests are running well.
image

image


JavaScript/packages/recognizers-date-time/src/dateTime/baseDate.ts
```diff
@@ -35,11 +35,13 @@ export class BaseDateExtractor implements IDateTimeExtractor {
         this.config = config;
     }
 
-    extract(source: string, refDate: Date): Array<ExtractResult> {
-        if (!refDate) refDate = new Date();
+    extract(source: string, refDate: Date): ExtractResult[] {
+        if (!refDate) {
+refDate = new Date();
```
tellarin: It seems the issue @STayinloves mentioned still happens and the auto-fix breaks indentation.

JavaScript/packages/recognizers-date-time/src/dateTime/baseDate.ts
```diff
@@ -325,7 +331,8 @@ export class BaseDateParser implements IDateTimeParser {
                 if (pastDate >= referenceDate) {
                     pastDate.setMonth(pastDate.getMonth() - 1);
                 }
-            } else {
+            }
+ else {
```
tellarin: Same.
Is there some automated way to also fix these? :)

Aliandi:
Hey @tellarin, we applied the feedback manually. We are now looking into a way to automatically do this. We'll keep you updated.

PR #1777, opened 2019-08-06
[CN Number] Fix wrong parse result for '一万五千亿' .

### LINKED ISSUE #759: [CN Number] Wrong parse result for '一万五千亿'
number_results=Recognizers.recognize_number("一万五千亿", Culture.Chinese)
for result in number_results:
    print(json.dumps(result, default=lambda o: o.__dict__, indent='\t', ensure_ascii=False))

the result is:

{
	"start": 0,
	"end": 4,
	"resolution": {
		"value": "500000010000"
	},
	"text": "一万五千亿",
	"type_name": "number"
}

the value is wrong

FeifanLiu98:
To fix #759.


Specs/Number/Chinese/NumberModel.json
```diff
@@ -4633,5 +4633,47 @@
...
+        "End": 4,
+        "TypeName": "number",
+        "Resolution": {
+          "value": "430000000000"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "四万亿",
```
wgx998877: the "四万万" is also supported?

FeifanLiu98: Yes, and should I add this example in spec?

PR #1784, opened 2019-08-06
[.NET] Italian Number support for skipped cases

LionbridgeCSII:
Support for skipped cases:
    [uno e mezzo] (one and a half)
    [undicesimo] (eleventh)

The fix for the second pattern required a modification to BaseNumberParser (switching the order of Cardinal and Ordinal maps for Italian)

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Italian/Parsers/ItalianNumberParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Italian/Parsers/ItalianNumberParserConfiguration.cs
@@ -71,11 +71,11 @@ namespace Microsoft.Recognizers.Text.Number.Italian
                 }
             }
 
-            /*The following piece of code is needed in Italian to correctly compute some fraction patterns
-             * e.g. 'due milioni duemiladuecento quinti' (=2002200/5) which is otherwise interpreted as
-             * 2000000/2205 (in Italian, isolated ordinals <10 have a different form respect to when
-             * they are concatenated to other numbers, so the following lines try to keep them isolated
-             * by concatenating the two previous numbers) */
+            // The following piece of code is needed in Italian to correctly compute some fraction patterns
+            // e.g. 'due milioni duemiladuecento quinti' (=2002200/5) which is otherwise interpreted as
+            // 2000000/2205 (in Italian, isolated ordinals <10 have a different form respect to when
+            // they are concatenated to other numbers, so the following lines try to keep them isolated
+            // by concatenating the two previous numbers)
             var fracLen = fracWords.Count;
             if (fracLen > 2 && this.OneToNineOrdinalRegex.Match(fracWords[fracLen - 1]).Success)
             {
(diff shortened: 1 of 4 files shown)


.NET/Microsoft.Recognizers.Text.Number/Italian/Parsers/ItalianNumberParserConfiguration.cs
```diff
@@ -86,6 +86,19 @@ public override IEnumerable<string> NormalizeTokenSet(IEnumerable<string> tokens
                 }
             }
 
+            /*The following piece of code is needed to compute the fraction pattern number+'e mezzo'
```
tellarin: Nitpick: Please add the start and end multi-line comment markers on their own lines.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -26,6 +26,16 @@ public BaseNumberParser(INumberParserConfiguration config)
                                 GetKeyRegex(this.Config.CardinalNumberMap.Keys) + "|" +
                                 GetKeyRegex(this.Config.OrdinalNumberMap.Keys);
 
+            /*For Italian, we invert the order of Cardinal and Ordinal in singleIntFrac in order to correctly extract
```
tellarin: Nit: Same,

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -26,6 +26,16 @@ public BaseNumberParser(INumberParserConfiguration config)
                                 GetKeyRegex(this.Config.CardinalNumberMap.Keys) + "|" +
```
tellarin: I suggest we extract this to a private method - BuildTextNumberRegex(). 
From `var singleIntFrac` to 'TextNumberRegex = new Regex(textNumberPattern, RegexOptions.Singleline | RegexOptions.Compiled);'. The new method can return the right regex for assignment here.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
tellarin: Please add a `@TODO` code comment here to analyze remodeling the creation of this regex in the future; to call attention to it.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
LionbridgeCSII: @tellarin, we updated the files

tellarin: Not quite what I meant for the comments, but ok. :)
```
/* 
 * It could look
 * like this
 */
```

PR #1790, opened 2019-08-08
[EN DateTimeV2] Avoid some wrong recognize, like "3 this"

songwenhao1:
"ABC-12345-A1B2C3 this is yet to be submitted" would be recognized as "3 this" in ExtendType Mode.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
@@ -383,6 +383,11 @@ namespace Microsoft.Recognizers.Text.DateTime
                 relativeTermsMatches.AddRange(regex.Matches(text).Cast<Match>());
             }
 
+            if (relativeTermsMatches.Count == 0 || (relativeTermsMatches.Count == 1 && config.ThisPrefixRegex.IsMatch(relativeTermsMatches[0].Value)))
+            {
+                return;
+            }
+
             // Remove overlapping matches
             relativeTermsMatches.RemoveAll(m =>
                 ers.Any(e => e.Start <= m.Index && e.Start + e.Length >= m.Index + m.Length));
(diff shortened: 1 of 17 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
```diff
@@ -383,6 +383,11 @@ private void ResolveImplicitRelativeDatePeriod(List<ExtractResult> ers, string t
                 relativeTermsMatches.AddRange(regex.Matches(text).Cast<Match>());
             }
 
+            if (relativeTermsMatches.Count == 0 || (relativeTermsMatches.Count == 1 && config.ThisPrefixRegex.IsMatch(relativeTermsMatches[0].Value)))
```
tellarin: Can you add a code comment here to describe what this 'if' is doing? It will greatly help people better understand the code.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseDateTimeAltExtractor.java
```diff
@@ -197,6 +197,13 @@ public BaseDateTimeAltExtractor(IDateTimeAltExtractorConfiguration config) {
             relativeTermsMatches.addAll(Arrays.asList(RegExpUtility.getMatches(regex, text)));
         }
 
+        List<ExtractResult> results = new ArrayList<>();
+        if (relativeTermsMatches.size() == 0 || (relativeTermsMatches.size() == 1 &&
```
tellarin: Same.

Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -11824,5 +11824,12 @@
     },
     "NotSupported": "javascript, python",
     "Results": []
+  },
+  {
+    "Input": "ABC-12345-A1B2C3 this is yet to be submitted",
```
tellarin: Can you also add this spec in ExtendedTypes.json as it seems to be the source of the original issue?

Issue #1414, opened 2019-01-25
[CN DateTimeV2] Bad cases with characters such as '早', '晚', '到', and Festivals


blazezhazha:
Hi,
I may find some bad cases in data_time recognition, and I split them into two classes. The code I use for testing is simple and shown below.
```python
from recognizers_date_time import DateTimeRecognizer
from recognizers_text import Culture, ModelResult


def test(text):
  recognizer = DateTimeRecognizer(Culture.Chinese)
  model = recognizer.get_datetime_model()
  result = model.parse(text)
  if result:
    for r in result:
      print(r.start, r.end, r.text, r.resolution, r.type_name)
```
#### 1. About '早', '晚', '到'
First, I figure that, when phrases including those three words occur in a sentence, the `parse` function will sometimes throw an error.
For instance, here are some test sentences including phrases/words that contains the character '早':
```python
text1 = '最早去普吉岛的是什么时候'
text2 = '他们官网上明天早上这几班没有取消'
text3 = '我是十号从意大利到土耳其，电子签请务必早办妥'
text4 = '你好我6月15号一早的飞机，出了票了，来得及吗，申请了直接邮寄的方式'
text5 = '8.26的飞机飞泰国现在办早吗'
```
For each of those texts, I run `test` and always get a `KeyError: 'left'`. For `text4`, for example, the complete traceback is:
```shell
Traceback (most recent call last):
  File ".../date_normalize_test.py", line 76, in <module>
    test(text2)
  File ".../date_normalize_test.py", line 18, in test
    result = model.parse(text)
  File "...\recognizers-text\python\libraries\recognizers-date-time\recognizers_date_time\date_time\models.py", line 30, in parse
    parse_result = self.parser.parse(result, reference)
  File "...\recognizers-text\python\libraries\recognizers-date-time\recognizers_date_time\date_time\chinese\merged_parser.py", line 55, in parse
    result = self.config.date_time_period_parser.parse(source, reference)
  File "...\recognizers-text\python\libraries\recognizers-date-time\recognizers_date_time\date_time\chinese\datetimeperiod_parser.py", line 36, in parse
    inner_result = self.merge_date_and_time_periods(source_text, reference)
  File "...\recognizers-text\python\libraries\recognizers-date-time\recognizers_date_time\date_time\chinese\datetimeperiod_parser.py", line 68, in merge_date_and_time_periods
    pr_timeperiod = self.config.time_period_parser.parse(er_timeperiod, reference)
[...]

PR #1794, opened 2019-08-09
[CN DataTimeV2]Fix Bug in“早”, “晚”, “到” and now python can recognize the festival(holiday) such as "春节".

FeifanLiu98:
Fix bug in #1414.
Modify DatePeriodDayRegexInChinese so it won't match blank string anymore. (It is really hard to find this bug. )
Add "早" into MorningTermList &"晚" into "EveningTermList". For example, now "七月五号晚" would be correctly extracted from "台风将于七月五号晚过境".

Regex Week_day_of_month initialized unsuccessfully leading to the problem in festivals. The program would break if a NULL Regex was asked to match a string.
Besides that, the Duration_Extractor is null which shouldn't be null. I think it is another python bug that needs to be fixed in python specifically.


tellarin:
@FeifanLiu98 why not supported in Python? The original issue report is a python one, so ideally it should also be fixed in that platform unless there is a blocking issue for it.

FeifanLiu98:
Okay,  I will try to fix the bug in Python.

Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -100,7 +100,7 @@ DatePeriodTillRegex: !simpleRegex
...
+  def: (?<day>初一|三十|一日|十一日|二十一日|三十一日|二日|三日|四日|五日|六日|七日|八日|九日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|一日|十一日|十日|二十一日|二十日|三十一日|三十日|二日|三日|四日|五日|六日|七日|八日|九日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|十日|二十日|三十日|10日|11日|12日|13日|14日|15日|16日|17日|18日|19日|1日|20日|21日|22日|23日|24日|25日|26日|27日|28日|29日|2日|30日|31日|3日|4日|5日|6日|7日|8日|9日|一号|十一号|二十一号|三十一号|二号|三号|四号|五号|六号|七号|八号|九号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|一号|十一号|十号|二十一号|二十号|三十一号|三十号|二号|三号|四号|五号|六号|七号|八号|九号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|十号|二十号|三十号|10号|11号|12号|13号|14号|15号|16号|17号|18号|19号|1号|20号|21号|22号|23号|24号|25号|26号|27号|28号|29号|2号|30号|31号|3号|4号|5号|6号|7号|8号|9号|一|十一|二十一|三十一|二|三|四|五|六|七|八|九|十二|十三|十四|十五|十六|十七|十八|十九|二十二|二十三|二十四|二十五|二十六|二十七|二十八|二十九|一|十一|十|二十一|二十|三十一|三十|二|三|四|五|六|七|八|九|十二|十三|十四|十五|十六|十七|十八|十九|二十二|二十三|二十四|二十五|二十六|二十七|二十八|二十九|十|二十|三十|廿|卅)
```
FeifanLiu98: The change in this line is at the end of the line. One "||" that can match a blank string was deleted.

Issue #1771, opened 2019-08-05
[JA DateTime] How to suggest spec bug fix?


combacsa:
**Is your feature request related to a problem? Please describe.**
I'm glad that this project has excellent translated Japanese DateTime Spec even though Patterns have some Chinese-only phrases and grammar which is clearly not Japanese ones...

However, I found some bugs on spec. Let's take a look at an [example on DatePeriodParser.json](https://github.com/microsoft/Recognizers-Text/blob/57800ad42419bd8c3898d2b5bc1d6a50f1ee4160/Specs/DateTime/Japanese/DatePeriodParser.json#L894
):

```json
  {
    "Input": "10月2日から10月22日まで不在にします。",
    "Context": {
      "ReferenceDateTime": "2016-11-07T00:00:00"
    },
    "NotSupported": "dotnet",
    "NotSupportedByDesign": "javascript,python,java",
    "Results": [
      {
        "Text": "10月2日から10月22日まで",
        "Type": "daterange",
        "Value": {
          "Timex": "(XXXX-10-02,XXXX-10-22,P20D)",
          "FutureResolution": {
            "startDate": "2017-10-02",
            "endDate": "2017-10-22"
          },
          "PastResolution": {
            "startDate": "2016-10-02",
            "endDate": "2016-10-22"
          }
        },
        "Start": 12,
        "Length": 25
      }
    ]
  },
```

Extracted string doesn't begin at index 12, it's from the beginning. And length is 15, not 22. This kind of bugs are easily fixable, I believe. A bit tough example is something like [this](https://github.com/microsoft/Recognizers-Text/blob/57800ad42419bd8c3898d2b5bc1d6a50f1ee4160/Specs/DateTime/Japanese/DateParser.json#L2612):

```
  {
    "Input": "私は1日の日曜日に戻ります",
    "Context": {
      "ReferenceDateTime": "2017-09-27T00:00:00"
    },
    "NotSupportedByDesign": "dotNet, javascript, python, java",
    "Results": [
      {
        "Text": "1日の日曜日",
        "Type": "date",
        "Value": {
          "Timex": "2017-09-03",
          "FutureResolution": {
            "date": "2017-09-03"
          },
          "PastResolution": {
            "date": "2017-09-03"
          }
        },
        "Start": 2,
        "Length": 6
      }
    ]
  },
```

[...]

PR #1798, opened 2019-08-10
[JA DateTime] Specs amended #1771

combacsa:
Examples of amendments available on #1771 

* Start/Length desgination of extracted text
* Wrong translation
* Etc


Specs/DateTime/Japanese/DateParser.json
```diff
@@ -619,21 +619,22 @@
...
+    "Input": "私は第三火曜日に戻ります",
+    "Comment": "Translated from 'third thursday'.",
     "Context": {
       "ReferenceDateTime": "2017-09-27T00:00:00"
     },
     "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
-        "Text": "3日火曜日",
```
tellarin: Would this case ('3日火曜日") never happen? Even in informal texting?
If it could happen, then we should also have a spec for it. can you add a new spec case in a new PR? This wouldn't block the current one.

combacsa: Grammatically '3日火曜日' just means "third day of that month which is Tuesday" and it's [already on the spec file](https://github.com/combacsa/Recognizers-Text/blob/c25e071fdfb2dce65121bab45da2c373b531151a/Specs/DateTime/Japanese/DateExtractor.json#L512). (27th day of that month which is Sunday). So I don't think it is necessary or probable to add a new spec case.

combacsa: And thanks for the reviewing and merging :)

PR #1801, opened 2019-08-13
[.NET] Turkish DateTime Date support

LionbridgeCSII:
DateExtractor : pass: 54 | fail: 2
DateParser: pass: 89 | fail: 3

Failing Cases:
- "dünden iki gün önce neredeydin" (where were you two days before yesterday?)
the structure of the sentence in Turkish is "yesterday two day before where were you?".
It is not supported because the method ExtractorDurationWithBeforeAndAfter only checks for the group "day" after the duration match "two day before".
- "Yarından itibaren üç hafta sonra uygun musun?" (Are you available three weeks from tomorrow?)
Same as above, 'Yarından' (tomorrow) appears before the duration match 'itibaren üç hafta sonra' (after three weeks from)
- "iki bin yılı Eylül'ün onunda" (on the tenth of September, two thousand)
similarly to the previous cases, here the year 'iki bin yılı' (two thousand) comes before the NumberWithMonth match 'eylül'ün onu' (tenth of September) but the method ExtendWithWeekdayAndYear only checks for years after the match. 

Wondering if we should proceed as we did for Italian Set, moving the involved methods (e.g. NumberWithMonth) to a separate class and using composition to be able to modify these functions only for Turkish.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
@@ -331,13 +331,23 @@ namespace Microsoft.Recognizers.Text.DateTime
                                 // Get week day from text directly, compare it with the weekday generated above
                                 // to see whether they refer to the same week day
                                 var extractedWeekDayStr = matchCase.Groups["weekday"].Value;
-                                var matchLength = result.Start + result.Length - matchCase.Index;
+
+                                // calculate matchLength considering that matchCase can preceed or follow result
+                                var matchLength = matchCase.Index < result.Start ? result.Start + result.Length - matchCase.Index : matchCase.Index + matchCase.Length - result.Start;
 
                                 if (!date.Equals(DateObject.MinValue) &&
                                     numWeekDayInt == Config.DayOfWeek[extractedWeekDayStr] &&
                                     matchCase.Length == matchLength)
                                 {
-                                    ret.Add(new Token(matchCase.Index, result.Start + result.Length ?? 0));
+                                    if (matchCase.Index < result.Start)
+                                    {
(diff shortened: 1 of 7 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -454,6 +464,16 @@ private void ExtendWithWeekdayAndYear(ref int startIndex, ref int endIndex, int
             // Check whether there's a weekday
             var prefix = text.Substring(0, startIndex);
             var matchWeekDay = this.Config.WeekDayEnd.Match(prefix);
+
+            // Check for weekday in the suffix
+            if (!matchWeekDay.Success)
+            {
+                suffix = text.Substring(endIndex);
+                var weekDayStart = @"^\s*,?\s*" + this.Config.WeekDayRegex;
+                Regex weekDayStartRegex = new Regex(weekDayStart, RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
tellarin: @LionbridgeCSII, I made a mistake merging this. Sorry.
Regexes should never be constructed within methods that are called multiple times. Much less with flags like 'Compiled'. This change has caused a massive performance hit.
The correct approach is to introduce a new regex in the relevant config, or to modify regexes only during initialization. In this case, the solution would be to declare a WeekDayStart regex in YAML and config used here (similar scenario to WeekDayEnd referenced a couple lines earlier). 
For Turkish, this regex would be populated, for other languages, it would be kept empty or use a placeholder that never matches.
We'll push a fix soon. Please be aware of such changes in the future.

LionbridgeCSII: ok, we will keep that in mind

PR #1802, opened 2019-08-14
[FR|EN Currency] Use NumberMode info in FractionExtractor to fix cases like "$2000 over 3"

songwenhao1:
Now only support in .NET


.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Extractors/FrenchNumberWithUnitExtractorConfiguration.cs
```diff
@@ -23,7 +24,7 @@ public abstract class FrenchNumberWithUnitExtractorConfiguration : INumberWithUn
         protected FrenchNumberWithUnitExtractorConfiguration(CultureInfo ci)
         {
             this.CultureInfo = ci;
-            this.UnitNumExtractor = NumberExtractor.GetInstance();
+            this.UnitNumExtractor = NumberExtractor.GetInstance(NumberMode.Unit);
```
tellarin: Please enable this in all unit extractors for the other languages.

PR #1803, opened 2019-08-14
[PhoneNumber] Fix boundary issues and add phone number pre-check

Grey0202:
- Fix the case like "March 1 14377 200:00"
- Add a simple pre-check to detect the phone number before extract.


.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BasePhoneNumberExtractor.cs
```diff
@@ -124,11 +124,16 @@ public override List<ExtractResult> Extract(string text)
                             }
                         }
 
+                        ers.Remove(er);
```
tellarin: Can you add a code comment here to describe what types of entities would fall here?

Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -4673,5 +4673,17 @@
   {
     "Input": "+1,032.923 is not a phone number",
     "Results": []
+  },
+  {
+    "Input": "0\t0\t0\t0\t0",
```
tellarin: You can leave only one such case with tabs. What about one that has enough 0 to be a phone number?

PR #1808, opened 2019-08-19
[CN DataTimeV2] Finalize bug fix for “早”, “晚”

FeifanLiu98:
To fix the wrong extractions of "早" in "早稻田" and "晚" in "晚礼服".
Now the single word "早" and “晚” won't be recognized as an entity.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
         public static readonly Regex UntilRegex = new Regex(DateTimeDefinitions.ParserConfigurationUntil, RegexFlags);
         public static readonly Regex SincePrefixRegex = new Regex(DateTimeDefinitions.ParserConfigurationSincePrefix, RegexFlags);
         public static readonly Regex SinceSuffixRegex = new Regex(DateTimeDefinitions.ParserConfigurationSinceSuffix, RegexFlags);
+        public static readonly Regex SingleWordMOAndEVRegex = new Regex(DateTimeDefinitions.SingleWordMorningAndEveningRegex, RegexFlags);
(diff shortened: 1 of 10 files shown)


FeifanLiu98:
#1794

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
```diff
@@ -109,6 +110,12 @@ private static void CheckBlackList(ref List<ExtractResult> extractResults, strin
                     continue;
                 }
 
+                // for cases like "早" in "早稻田" and "晚" in "晚安"
+                if (SingleWordMOAndEVRegex.IsMatch(extractResult.Text))
```
wgx998877: I don't think this design is make sense as we use specify regex to filter that single char entity. I think we can define a HashSet or Dictionary to filter the single char entities.

FeifanLiu98:
I combine the regex which can fix "早" and "晚" with one regex which deals with "12号" into a dictionary. And the code about "12号" was deleted from the program.
Now Chinese has AmbiguityFilter function as well as English, etc.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
```diff
@@ -87,7 +93,6 @@ private static List<ExtractResult> MoveOverlap(List<ExtractResult> dst, ExtractR
         private static void CheckBlackList(ref List<ExtractResult> extractResults, string text)
```
tellarin: Rename do CheckDenyList

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
```diff
@@ -59,6 +63,8 @@ public List<ExtractResult> Extract(string text, DateObject referenceTime)
 
             CheckBlackList(ref ret, text);
```
tellarin: Why do we need to both CheckDenyList and then FilterAmbiguity? Shouldn't they be the same or at least the ambiguity filter be inside the other method?

Issue #556, opened 2018-05-16
[ES DateTimeV2] The entity "la semana pasada" not being detected 


fanidamj:
The entity "la semana pasada" not being detected whereas the entity "esta semana" is recognized.

![spanishdatetimev2](https://user-images.githubusercontent.com/16520557/40131526-03e40e78-58ef-11e8-99fe-61cecae0e578.PNG)

Grey0202:
PR #1818 fix most of the cases in this issue except for "inicio de mes". That will be fixed in coming PR with other improvements.

tellarin:
Marking this as solved. Cases like "inicio de mes" --> start of month will be moved to a new issue.

PR #1818, opened 2019-08-22
[ES DateTimeV2] Fix problem recognizing entities like "la semana pasada"

Grey0202:
1. Fixes issue #556 partially. The "inicio" case is still not supported.
- Now it can extract "el mes pasado" "semana pasada" "la semana pasada".
- Add extra tests.

2. Fix Age issue, such as "Semana" (in Spanish), "Woche" (in German) being wrongly recognized by themselves (workaround)


JavaScript/packages/recognizers-date-time/src/resources/spanishDateTime.ts
```diff
@@ -250,10 +251,10 @@ export namespace SpanishDateTime {
     export const MinusOneDayTerms = [ "ayer","ultimo dia" ];
     export const PlusTwoDayTerms = [ "pasado mañana","dia despues de mañana" ];
     export const MinusTwoDayTerms = [ "anteayer","dia antes de ayer" ];
-    export const MonthTerms = [ "mes","meses" ];
+    export const MonthTerms = [ "mes","meses","mes pasado","mes pasados" ];
```
tellarin: Such relative terms should not be added to the Terms lists. For example, in English, the lists are clean.

Specs/DateTime/Spanish/DatePeriodExtractor.json
```diff
@@ -4611,5 +4611,41 @@
         "Length": 14
       }
     ]
+  },
+  {
+    "Input": "semana pasada",
```
tellarin: "semana" is being wrongly recognized as an Age entity. Can you add a spec for that and look into the cause of the issue? This PR should include that fix too. Thanks.

Specs/DateTime/Spanish/DatePeriodExtractor.json
```diff
@@ -4611,5 +4611,41 @@
...
+        "Text": "semana pasada",
+        "Type": "daterange",
+        "Start": 0,
+        "Length": 13
+      }
+    ]
+  },
+  {
+    "Input": "la semana pasada",
+    "NotSupportedByDesign": "javascript, python",
```
tellarin: Could we enable these tests in js and python too? Or is there a complicating factor?

tellarin:
@Grey0202 please mention in the PR description which issue this is fixing. Also, please add a comment that the "inicio" case will be handled elsewhere.

Specs/DateTime/Spanish/DatePeriodParser.json
```diff
@@ -6950,5 +6949,83 @@
...
+            "endDate": "2019-08-19"
+          }
+        },
+        "Start": 0,
+        "Length": 13
+      }
+    ]
+  },
+  {
+    "Input": "el mes pasado",
```
tellarin: Can you add such cases also in DateTimeModel?

tellarin: Also, will you add the fix for Age in this same PR?

PR #1825, opened 2019-08-26
[DataTimeV2] Support "from 2018"(#1576)

### LINKED ISSUE #1576: [EN DateTimeV2] Cannot recognize "from 2018"
Only "2018" is recognized. Actually "from 2018" should be similar to ">= 2018".

songwenhao1:
This fixes #1576


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -35,17 +35,35 @@ public static bool HasTokenIndex(string text, Regex regex, out int index)
...
+            }
+
             if (HasTokenIndex(beforeStr.TrimEnd(), tokenRegex, out var tokenIndex))
             {
                 var modLength = beforeStr.Length - tokenIndex;
 
                 er.Length += modLength;
                 er.Start -= modLength;
                 er.Text = text.Substring(er.Start ?? 0, er.Length ?? 0);
+                if (er.Metadata == null)
```
tellarin: This if/else could be a simple method AssignModMetadata() or something like that. It would cleanup the current method a little and make it easier to change metadata in the future.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -293,7 +311,7 @@ private List<ExtractResult> AddMod(List<ExtractResult> ers, string text)
 
                 if (!success)
                 {
-                    success = TryMergeModifierToken(er, config.SinceRegex, text);
+                    success = TryMergeModifierToken(er, config.SinceRegex, text, true);
```
tellarin: This is the only place where true is passed?
Can you add a code comment here explaining why? Also, please make it a named parameter in the call to TryMergeModifierToken.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -35,17 +35,35 @@ public static bool HasTokenIndex(string text, Regex regex, out int index)
             return false;
         }
 
-        public static bool TryMergeModifierToken(ExtractResult er, Regex tokenRegex, string text)
+        public bool TryMergeModifierToken(ExtractResult er, Regex tokenRegex, string text, bool ambiguous = false)
```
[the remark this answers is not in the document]
tellarin: For lack of a better suggestion: potentialAmbiguity

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -148,6 +141,18 @@ public List<ExtractResult> Extract(string text, DateObject reference)
             return ret;
         }
 
+        private void AssignModMetadata(ExtractResult er)
```
tellarin: Please mark the parameter as 'ref' in such cases.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -56,14 +56,7 @@ public bool TryMergeModifierToken(ExtractResult er, Regex tokenRegex, string tex
...
                 er.Text = text.Substring(er.Start ?? 0, er.Length ?? 0);
-                if (er.Metadata == null)
-                {
-                    er.Metadata = new Metadata { HasMod = true };
-                }
-                else
-                {
-                    er.Metadata.HasMod = true;
-                }
+                AssignModMetadata(er);
```
tellarin: Nit: Add a blank line before this to make it easier to read.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -309,9 +314,10 @@ private List<ExtractResult> AddMod(List<ExtractResult> ers, string text)
                     success = TryMergeModifierToken(er, config.AfterRegex, text);
                 }
 
+                // The SinceRegex contains potential ambiguity part("from"), so the parameter potentialAmbiguity is true.
```
tellarin: We can re-phrase this to something like:
SinceRegex in English contains the term "from" which is potentially ambiguous with ranges in the form "from X to Y"
Also, please put the code comment before the method call, as it refers to it. Not before the 'if'

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseMergedDateTimeExtractor.java
```diff
@@ -213,6 +213,7 @@ private boolean shouldSkipFromToMerge(ExtractResult er) {
                 modifiedToken = tryMergeModifierToken(er, config.getAfterRegex(), text);
             }
 
+            // SinceRegex in English contains the term "from" which is potentially ambiguous with ranges in the form "from X to Y"
```
tellarin: Comment inside the if, like in .NET

Issue #1826, opened 2019-08-26
[* DateTimeV2] Javascript DateTimeRecognizer throws error with certain strings


jheising:
**Describe the bug**
The javascript date time recognizer throws an error when using the string `"Must deliver sixteen (16) units before the thirty-first day of the month."`

**To Reproduce**
1. Call `DateTimeRecognizers.recognizeDateTime("Must deliver sixteen (16) units before the thirty-first day of the month.", "en-us")`
2. Throws error `TypeError: Cannot set property 'mod' of undefined` at `@microsoft/recognizers-text-date-time/dist/recognizers-text-date-time.umd.js:18976:21`

**Additional context**

The problem appears to begin in `baseDatePeriod.ts` line 330 (see [here](https://github.com/microsoft/Recognizers-Text/blob/master/JavaScript/packages/recognizers-date-time/src/dateTime/baseDatePeriod.ts#L330))

`let resultValue;` is never intialized

The problem appears to manifest itself in `baseMerged.ts` in code blocks that begin with `if (hasBefore && pr.value !== null)` (see [here](https://github.com/microsoft/Recognizers-Text/blob/master/JavaScript/packages/recognizers-date-time/src/dateTime/baseMerged.ts#L316-L323)) when `pr.value` is set to `undefined`.

```javascript
if (hasBefore && pr.value !== null) { // If pr.value is set to undefined, this block will execute
            pr.length += modStr.length;
            pr.start -= modStr.length;
            pr.text = modStr + pr.text;
            let val = pr.value; // val is set to undefined here
            val.mod = TimeTypeConstants.beforeMod; // undefined.mod will throw an error
            pr.value = val;
}
```

The fix appears to be to make sure that `resultValue` is always initialized to `null` in `baseDatePeriod.ts`.

`let resultValue = null;` 

Pull request coming.

tellarin:
@jheising in which version of the code are you seeing this behaviour? The latest npm package, the latest codebase, or both?
This has actually highlighted a behaviour difference between .NET and JavaScript, so we'll analyze further. 
Thanks for reporting!

tellarin:
I'll close this issue as the crash no longer happens and the JS behaviour is now the same as .NET
And I'll create a new issue to properly support parsing/resolving "before the thirty-first day of the month" as a single entity.
Thanks.

PR #1827, opened 2019-08-26
Fixes Javascript DateTimeRecognizer error - Fixes Issue #1826

jheising:
`resultValue` should be initialized to `null`.

Fixes Issue #1826


tellarin:
@jheising I'll close this as there are wider changes to be done to properly fix the root issue.
"before the thirty-first day of the month" should ideally be recognized as one entity and properly parsed/resolved in both .NET and JS with the same behaviour.
The changes from PR #1831 should avoid such crashes for now and at least get the code to behave the same as .NET.
Thanks again for reporting this!

PR #1839, opened 2019-08-29
[.NET] Turkish Date support to skipped cases

LionbridgeCSII:
Added boolean attribute to yaml files to determine if both before and after strings should be checked for match extension and modified methods in AgoLaterUtils to work according to this attribute.
With this modification, all Date test cases (and others in DatePeriod) pass.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Turkish/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Turkish/DateTimeDefinitions.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Recognizers.Definitions.Turkish
 
     public static class DateTimeDefinitions
     {
+      public const bool CheckBothBeforeAfter = true;
       public const string TillRegex = @"(?<till>\b(kadar|dek|değin)\b)";
       public const string RangeConnectorRegex = @"(?<and>\b(ile|ila)\b|(-|—|——|–))";
       public const string RelativeRegex = @"\b(?<order>ertesi|(bir\s+)?sonraki|gelecek|bu|geçen|son|aynı|(bir\s+)?önceki|evvelki|önümüzdeki)\b";
@@ -105,7 +106,7 @@ namespace Microsoft.Recognizers.Definitions.Turkish
       public const string WeekOfRegex = @"(haftası)";
       public const string MonthOfRegex = @"(ayı)";
       public static readonly string DateYearRegex = $@"(?<year>{BaseDateTime.FourDigitYearRegex}|{TwoDigitYearRegex})";
-      public static readonly string YearSuffix = $@"(,?\s*({DateYearRegex}|{FullTextYearRegex}))";
+      public static readonly string YearSuffix = $@"(,?\s*({DateYearRegex}|{FullTextYearRegex})(\s+yılı)?)";
       public const string OnRegex = @"^[\*]";
       public const string RelaxedOnRegex = @"(?<day>(10|20|30|31|(1|2)[1-9])('i|'si|'sı|'ü|'u))(?=(nde|nda))\b";
       public const string PrefixWeekDayRegex = @"(\s*[-—–])";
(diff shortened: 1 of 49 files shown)


LionbridgeCSII:
@tellarin, this is ready for review and the following PRs depend on changes included in this one

.NET/Microsoft.Recognizers.Definitions.Common/Turkish/DateTimeDefinitions.cs
```diff
@@ -21,6 +21,7 @@ namespace Microsoft.Recognizers.Definitions.Turkish
 
     public static class DateTimeDefinitions
     {
+      public const bool CheckBothBeforeAfter = true;
```
tellarin: Does Turkish really needs to check "both sides" in all these situations? I'm wondering if we should have finer-grained flags.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -93,6 +93,30 @@ public List<Token> ExtractRelativeDurationDateWithInPrefix(string text, List<Ext
                         }
                     }
                 }
+                else if (Config.CheckBothBeforeAfter)
+                {
+                    // Check also afterStr
+                    match = Config.InConnectorRegex.MatchBegin(afterStr, trim: true);
```
tellarin: Instead of basically repeating the code in the 'else', could you turn the bodies into a method? 
Ex: call the method with the default direction to check, else if Config.CheckBothBeforeAfter, call the same method with the other direction. 
I see the same in other parts of the code. This would greatly ease maintenance in the future.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
```diff
@@ -458,11 +483,25 @@ private void ExtendWithWeekdayAndYear(ref int startIndex, ref int endIndex, int
                     endIndex += matchYear.Length;
                 }
             }
+            else if (Config.CheckBothBeforeAfter)
```
tellarin: Similar to the comment above. Same elsewhere.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/AgoLaterUtil.cs
```diff
@@ -28,6 +28,19 @@ public static class AgoLaterUtil
 
         public static List<Token> ExtractorDurationWithBeforeAndAfter(string text, ExtractResult er, List<Token> ret,
                                                                       IDateTimeUtilityConfiguration utilityConfiguration)
+        {
+            if (utilityConfiguration.CheckBothBeforeAfter)
+            {
+                return ExtractorDurationWithBeforeAndAfterInBeforeAndAfterString(text, er, ret, utilityConfiguration);
```
tellarin: Can there also be some simplification here? I didn't get why two methods need to exist.
Couldn't it be the same, controlled by a parameter?

LionbridgeCSII: @tellarin, here we have duplicated the method for readability, since the original one is shorter and easier to review. But there is no need for it, we will merge the two.

.NET/Microsoft.Recognizers.Definitions.Common/Turkish/DateTimeDefinitions.cs
LionbridgeCSII: This seems the simpler solution at the moment, since the same methods are used on patterns that have different structures (match in the prefix or match in the suffix).
For example, "iki bin yılı Eylül'ün onunda" (on the tenth of September, two thousand) and "10 eylül bin dokuz yüz bir" (on September 10, nineteen zero one) are both parsed by ParseNumberWithMonth. But in the first, the year "iki bin yılı" is found in the prefix, while in the second the year "bin dokuz yüz bir" is found in the suffix.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
LionbridgeCSII: Done

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
LionbridgeCSII: Done (here and elsewhere)

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/AgoLaterUtil.cs
LionbridgeCSII: merged

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateExtractor.cs
LionbridgeCSII: Note, the file had an inconsistent line ending, it has been changed.

PR #1843, opened 2019-09-01
Added some more very used ChoiceDefinitions in Spanish

warnov:
Some very common words and expression that in Spanish are used as answer for yes/no choices.


tellarin:
@warnov, I'll close this one for now as it's getting stale. We can re-open in the future.

PR #1845, opened 2019-09-02
[CN DateTimeV2] Support to recognize "上半年"和“下半年” in Chinese

### LINKED ISSUE #1824: [CN DateTimeV2] Incorrect "start" and "end" time for "上半年"and "下半年"
```
上半年 {'values': [{'timex': '(2019-08-26,2019-08-26,P0Y)', 'type': 'daterange', 'start': '2019-08-26', 'end': '2019-08-26'}]}
下半年 {'values': [{'timex': '(2019-08-27,2019-08-27,P0Y)', 'type': 'daterange', 'start': '2019-08-27', 'end': '2019-08-27'}]}
```
Why did the start and end time is the same here? Note: my reference day here is 2019-08-26.

songwenhao1:
This fixes #1824
Only supported in .NET now


.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
```diff
@@ -59,18 +59,19 @@ public static class DateTimeDefinitions
...
       public const string DatePeriodThisRegex = @"这个|这一个|这|这一|本";
-      public const string DatePeriodLastRegex = @"上个|上一个|上|上一";
-      public const string DatePeriodNextRegex = @"下个|下一个|下|下一";
+      public const string DatePeriodLastRegex = @"上个|上一个|上(?!半)|上一";
```
wgx998877: why not merge the half logic in this regex? It's not necessary to add a new regex to do that?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -733,6 +733,58 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
...
+                            halfType = 1;
+                        }
+                        else if (!string.IsNullOrEmpty(secondHalf))
+                        {
+                            halfType = 2;
+                        }
+
+                        if (halfType != 0)
+                        {
+                            if (trimmedText.StartsWith("去年"))
```
wgx998877: don't like this type of implementation

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -804,6 +872,18 @@ private DateTimeResolutionResult ParseYear(string text, DateObject referenceDate
                 var endDay = DateObject.MinValue.SafeCreateFromValue(year + 1, 1, 1);
 
                 ret.Timex = year.ToString("D4");
+
+                if (halfType == 1)
```
wgx998877: can we avoid this magic number?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
tellarin: Values should not be hardcoded in code. Values like "去年" should at least be in the YAML file.
The YAML files already have other similar cases that can be used as example.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
tellarin: If such numbers are really needed, please declare as Constants.

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
songwenhao1: Because I think they have different meanings, in the LastRegex, "上" means the last one. But "上" in "上半" means the first half-year, not means "previous half-year", same like "下" in the "下半" means the second half, so I think it should not be merged to LastRegex.

wgx998877: how about “上个半年”, “上一个半年”

wgx998877: why not add a special tag for "半"

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -815,6 +854,10 @@ private DateTimeResolutionResult ParseYear(string text, DateObject referenceDate
             if (match.Success)
             {
                 var tmp = match.Value;
+
+                // Handle like "二零一七年上半年"，"二零一七年下半年"
+                tmp = HandleWithHalfYear(match, tmp, out bool hasHalf, out bool isFirstHalf);
```
wgx998877: have a better name for the tmp

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -837,16 +880,54 @@ private DateTimeResolutionResult ParseYear(string text, DateObject referenceDate
...
+            }
 
-                ret.Timex = year.ToString("D4");
-                ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDay, endDay);
-                ret.Success = true;
+            return ret;
+        }
 
-                return ret;
+        private string HandleWithHalfYear(ConditionalMatch match, string text, out bool hasHalf, out bool isFirstHalf)
```
wgx998877: @tellarin do you think they need better names for those two functions?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -676,6 +676,13 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                         swift = -1;
                     }
 
+                    // Handle cases with "(上|下)半" like "上半月"、 "下半年"
+                    if (!string.IsNullOrEmpty(match.Groups["halfTag"].Value))
+                    {
+                        ret = HandleWithHalfTag(trimmedText, referenceDate, ret, swift);
```
wgx998877: return HandleWith...()

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -837,16 +880,54 @@ private DateTimeResolutionResult ParseYear(string text, DateObject referenceDate
                     year += 2000;
                 }
 
-                var beginDay = DateObject.MinValue.SafeCreateFromValue(year, 1, 1);
-                var endDay = DateObject.MinValue.SafeCreateFromValue(year + 1, 1, 1);
+                ret = HandleYearResult(ret, hasHalf, isFirstHalf, year);
```
wgx998877: return

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -676,6 +676,13 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
...
                     }
 
+                    // Handle cases with "(上|下)半" like "上半月"、 "下半年"
+                    if (!string.IsNullOrEmpty(match.Groups["halfTag"].Value))
+                    {
+                        ret = HandleWithHalfTag(trimmedText, referenceDate, ret, swift);
+                        return ret;
+                    }
+
                     if (trimmedText.EndsWith("周") | trimmedText.EndsWith("星期"))
```
tellarin: Such terms shouldn't be hardcoded. Can you move all to the YAML file?
Other YAML files have examples of this.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -715,6 +722,10 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                     }
                     else if (trimmedText.EndsWith("年"))
```
tellarin: Same

PR #1846, opened 2019-09-02
[* DateTimeV2] Fix some bugs and revise result of invalid date

Grey0202:
- Fix extract decimal before a month
- When the date is invalid, the value will be "not resolve" now.
- Fix cases like "31 Jun. 1903"(invalid) and "6,107.31 August 2019"


Specs/DateTime/English/DateExtractor.json
```diff
@@ -1587,5 +1587,11 @@
         "Length": 26
       }
     ]
+  },
+  {
+    "Input": "6,107.31 August 2019 should not include the decimal",
+    "Comment": "Something wrong with negative lookbehind in java",
```
tellarin: Change this comment to: "Only August 2019 should be extracted as a DateRange, so no output in Date only. Java disabled due to issue in lookbehind."

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -14025,6 +14025,68 @@
         }
       }
     ]
+  },
+  {
+    "Input": "6,107.31 August 2019 should not include the decimal",
```
tellarin: Please also add such cases in ComplexCalendar.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -14025,6 +14025,68 @@
...
+              "type": "date",
+              "value": "not resolved"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "feb 30",
```
tellarin: Please also add the full datetime case: "That's absurd like saying we should meet in feb 30 2019 17:20!"

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -40,9 +40,9 @@ public BaseMergedDateTimeParser(IMergedParserConfiguration configuration)
 
         public static void AddSingleDateTimeToResolution(Dictionary<string, string> resolutionDic, string type, string mod, Dictionary<string, string> res)
         {
+            // If the date is invalid, the result should be invalid.
```
tellarin: Can you make this comment more clear. Maybe include an example? I'm not sure that it actually means.
Also, what that the change in the 'if' below represent?

tellarin: "If an "invalid" Date or DateTime is extracted, it should not have an assigned resolution. Only valid entities should pass this condition."

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -143,6 +143,11 @@ public static void AddPeriodToResolution(Dictionary<string, string> resolutionDi
 
             if (!string.IsNullOrEmpty(start) && !string.IsNullOrEmpty(end))
             {
+                if (start.StartsWith(DateMinString) || end.StartsWith(DateMinString))
```
tellarin: Can you make this into a method? Call it "checkUnresolvedDates".
We should also probably mark the res parameter passed to this method as 'ref' to highlight it will be changed.

PR #1848, opened 2019-09-03
[AR Choice] Added Arabic Choice

ahmedabuamra:
- added specs file.
- added Arabic choice test class.
- added choice definitions `.yaml` file and generated `.cs` from it. 
- added `ArabicBooleanExtractorConfiguration.cs`
- Registered the Arabic choice parser in `ChoiceRecognizer.cs`


.NET/Microsoft.Recognizers.Text/Culture.cs
```diff
@@ -5,6 +5,7 @@ namespace Microsoft.Recognizers.Text
 {
     public sealed class Culture
     {
+        public const string Arabic = "ar";
```
tellarin: Shouldn't this be a specific variant? Or "ar-*"?
The goals should be to cover as many variants as possible without needing separate models.

.NET/Microsoft.Recognizers.Text.Choice/Arabic/Extractors/ArabicBooleanExtractorConfiguration.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.Recognizers.Text.Choice.Arabic
+{
+    public class ArabicBooleanExtractorConfiguration : IBooleanExtractorConfiguration
+    {
+        public static readonly Regex TrueRegex =
+            new Regex(ChoiceDefinitions.TrueRegex, RegexOptions.Singleline);
+
+        public static readonly Regex FalseRegex =
+            new Regex(ChoiceDefinitions.FalseRegex, RegexOptions.Singleline);
```
tellarin: Is it by coincidence that everything works right without telling the regexes to match right-to-left or is the behaviour indeed correct?
Choice is simpler, but in other types, especially DateTime, this will probably lead to issues.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -385,7 +385,7 @@ private void ValidateResults(TestModel testSpec, IEnumerable<string> testResolut
                         continue;
                     }
 
-                    Assert.AreEqual(expected.Resolution[key], actual.Resolution[key], GetMessage(testSpec));
+                    Assert.AreEqual(expected.Resolution[key].ToString(), actual.Resolution[key].ToString(), GetMessage(testSpec));
```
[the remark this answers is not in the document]
ahmedabuamra: The tests fail with the following message:
`Expected:<0.466666666666667>. Actual:<0.466666666666667>. Input: "أعتقد أن هذا سينجح. لذلك نعم سوف أقوم بذلك."`

I think the error here because of the delta issue in the double data type. so I changed them to string to match correctly.

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
```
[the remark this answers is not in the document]
ahmedabuamra: This file is modified automatically by visual studio, I should remove it from the commit.

.NET/Microsoft.Recognizers.Text/Culture.cs
ahmedabuamra: Yes I discussed that with @hazemelh and decided to make it "ar-*" to refer to the Modern Standard Arabic.

.NET/Microsoft.Recognizers.Definitions.Common/Microsoft.Recognizers.Definitions.Common.csproj
```diff
@@ -454,6 +461,10 @@
     <DotNetCliToolReference Include="T5.TextTransform.Tool" Version="1.1.0-*" />
   </ItemGroup>
 
+  <ItemGroup>
```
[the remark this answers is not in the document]
ahmedabuamra: Removed.

.NET/Microsoft.Recognizers.Text.Choice/Arabic/Extractors/ArabicBooleanExtractorConfiguration.cs
ahmedabuamra: I added `RegexOptions.RightToLeft` to regex options.

PR #1861, opened 2019-09-09
[EN DateTimeV2] Improve the method which used to avoid some wrong recognize, cases like "3 this"

songwenhao1:


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeAltExtractor.cs
```diff
@@ -383,11 +383,13 @@ private void ResolveImplicitRelativeDatePeriod(List<ExtractResult> ers, string t
...
             }
 
-            // Filtered out if there is no relative term or the only one found is "this". Like "3 this"
-            if (relativeTermsMatches.Count == 0 || (relativeTermsMatches.Count == 1 && config.ThisPrefixRegex.IsMatch(relativeTermsMatches[0].Value)))
-            {
-                return;
-            }
+            // Filtered out relative term "this" which is behind an extractResult.
+            // Handle cases like "3 this"
+            relativeTermsMatches = relativeTermsMatches.Where(relativeTerm =>
```
tellarin: Is this just filtering out "3 this" as an entity? Why not correct the wrong extraction earlier?
It's also not clear from the change why "3 this" breaks the other entities. Is is because of some parse error, or does it really affect some code related to the next "this week"?

PR #1864, opened 2019-09-10
[EN TimeZone] Add missisng Australian timezones

Grey0202:
- Add Central Daylight Time, Eastern Daylight Time and Australian Eastern Standard Time
- Re-gen definition files (previous changes were not pushed)


.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
```diff
@@ -188,6 +189,7 @@ public static class TimeZoneDefinitions
             @"Australian Central Time",
             @"AUS Eastern Standard Time",
             @"Australian Eastern Time",
+            @"Australian Eastern Standard Time",
```
tellarin: Can you also add:
Australian Central Daylight Time or ACDT has a UTC offset of +10:30.
Australian Eastern Daylight Time or AEDT has a UTC offset of + 11:00.
Please make sure "Eastern Daylight Time" and "Central Daylight Time"have offset as -1000 as they are ambiguous between the US and Australia.

.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
```diff
@@ -35,6 +35,7 @@ public static class TimeZoneDefinitions
             @"ART",
```
tellarin: Can you sort this list alphabetically? It will greatly help spotting missing/incorrect items.

Specs/DateTime/English/TimeZoneParser.json
```diff
@@ -870,5 +870,41 @@
         "Length": 6
       }
     ]
+  },
+  {
+    "Input": "I'm at Eastern Daylight Time",
```
tellarin: Don't forget to add specs to EnablePreview and ComplexCalendar. Please also add cases there where a specific time "merges" with one of the new timezone acronyms.

.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
```diff
@@ -35,6 +35,7 @@ public static class TimeZoneDefinitions
             @"ART",
             @"ADT",
             @"ACST",
```
tellarin: ACDT is still missing in the list.

.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
```diff
@@ -527,6 +534,10 @@ public static class TimeZoneDefinitions
             { @"atlantic standard", -240 },
             { @"aus central standard", 570 },
             { @"aus eastern standard", 600 },
+            { @"australian eastern", 600 },
+            { @"australian eastern standard", 600 },
+            { @"australian central daylight", 630 },
+            { @"australian eastern daylight Time", 660 },
```
tellarin: Why remove "@"central daylight", -300 }," and "@"eastern daylight", -240" from a previous commit?
These names are still timezones we want to catch, but their resolution should be -1000, as they are ambiguous.

tellarin: This entry should not have the word "time", right?

Patterns/English/English-TimeZone.yaml
```diff
@@ -16,6 +16,7 @@ AbbreviationsList: !list
     - ART
     - ADT
     - ACST
+    - AET
     - AEST
     - AEDT
     - AZT
```
tellarin: This list is not sorted. Could you also sort it, like the one for the resolutions? That really benefits checking for inconsistencies.

.NET/Microsoft.Recognizers.Definitions.Common/English/TimeZoneDefinitions.cs
tellarin: In case of such conflicts, the YAML file should contain both definitions, but commented out (with '#'). You can see some examples in the acronym resolution list.

PR #1870, opened 2019-09-12
[EN DateTimeV2] Support cases like "before|after breakfast|lunch"

songwenhao1:


.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -162,6 +162,10 @@ public static class Constants
         public const string Daytime = "TDT";
         public const string Night = "TNI";
         public const string BusinessHour = "TBH";
+        public const string Breakfast = "TMEB";
```
tellarin: Let's rename these to MealtimeBreakfast, MealtimeBrunch, ..., ok?

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -178,7 +178,7 @@ public static class DateTimeDefinitions
       public const string SuffixAfterRegex = @"\b(((at)\s)?(or|and)\s+(above|after|later|greater)(?!\s+than))\b";
       public const string PrepositionRegex = @"(?<prep>^(at|on|of)(\s+the)?$)";
       public const string LaterEarlyRegex = @"((?<early>early(\s+|-))|(?<late>late(r?\s+|-)))";
-      public static readonly string TimeOfDayRegex = $@"\b(?<timeOfDay>((((in\s+(the)?\s+)?{LaterEarlyRegex}?(in\s+(the)?\s+)?(morning|afternoon|night|evening)))|(((in\s+(the)?\s+)?)(daytime|business\s+hour)))s?)\b";
+      public static readonly string TimeOfDayRegex = $@"\b(?<timeOfDay>((((in\s+(the)?\s+)?{LaterEarlyRegex}?(in\s+(the)?\s+)?(morning|afternoon|night|evening)))|(?<=(before|after|around)\s+)(breakfast|brunch|lunch(time)?|dinner(time)?|supper)|(((in\s+(the)?\s+)?)(daytime|business\s+hour)))s?)\b";
```
tellarin: This is not the ideal place to add such terms. This is mixing the new terms that are not entities by themselves with modifiers that shouldn't be in this regex.
For example, there is another definition that already contains synonyms for "around" that should also be supported, but are not enabled here.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -255,6 +255,26 @@ public static TimeOfDayResolutionResult ParseTimeOfDay(string tod)
                     result.EndHour = 23;
                     result.EndMin = 59;
                     break;
+                case Constants.Breakfast:
```
tellarin: Instead of these repeated assignments, can't the cases just fall into the existing ranges?
```
case Constants.Evening:
case Constants.Dinner:
    result.Timex = Constants.Dinner;
```

PR #1872, opened 2019-09-12
[AR Numbers] Adding Arabic Numbers code skeleton and localized specs

ahmedabuamra:


Specs/Number/Arabic/NumberModel.json
```diff
@@ -187,47 +187,62 @@
     ]
   },
   {
-    "Input": "مائة وثلاثة وثلثين",
```
tellarin: Why are specific tests being modified? We should try to make sure that there is some level of parity with English (same cases) to guarantee coverage. 
If you need to add additional cases, or Arabic-specific ones, please append them at the end of the appropriate json files.

Specs/Number/Arabic/NumberModel.json
```diff
@@ -187,47 +187,62 @@
...
-    "NotSupportedByDesign": "javascript,python,java,dotnet",
```
tellarin: In the PR, test cases should only be marked as supported in .NET once the whole model passes. Otherwise tests will fail as they can't obtain the non-registered model.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -714,6 +714,14 @@ private List<string> GetMatches(string input)
 
         private double GetIntValue(List<string> matchStrs)
         {
+            // workarround to solve "و" which means "and" before rounded number in Arabic.
+            // ألف و مائة = one thousand and one hundred
+            // but in arabic there is no integer before hundred, because it's 100 by default.
+            if (matchStrs.Count == 1 && matchStrs.First() == "و")
```
tellarin: This should be a method in the Arabic config for numbers. Other languages should not call it.
Also, please declare "و" in YAML, not as a hardcoded value in code.

Patterns/Arabic/Arabic-Choice - Copy.yaml
```diff
@@ -0,0 +1,8 @@
+LangMarker: Ara
```
tellarin: Please be careful not to include such unnecessary files (* - Copy.*) in PRs.

Patterns/Arabic/Arabic-Numbers - Copy.yaml
```diff
@@ -0,0 +1,519 @@
+---
+LangMarker: Ara
```
tellarin: Same

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -1501,7 +1501,7 @@
     "Context": {
       "ReferenceDateTime": "2019-09-03T00:00:00"
     },
-	"Debug": true,
+    "Debug": true,
```
tellarin: Do not push specs with Debug flags. The PR should represent the code that's eventually going to merge.

Specs/Number/Arabic/NumberModel.json
```diff
@@ -914,28 +930,28 @@
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
-        "Text": "ثلاثة وعشرون وثلاثة أخماس",
+        "Text": "ثلاثة و عشرون و ثلاثة أخماس",
```
tellarin: Why all these changes in specs? Were the original sentences incorrect?
Are they translations of the English specs? We should strive to keep as much coherency of cases as possible (of course, new language-specific cases are very welcome).
There were similar review comments in previous commits in this PR, but no reply.

Specs/Number/Arabic/PercentModel - Copy.json
```diff
@@ -0,0 +1,222 @@
+[
```
tellarin: Same

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
```diff
@@ -80,14 +80,27 @@ public NumberRangeModel GetNumberRangeModel(string culture = null, bool fallback
...
-            /*
             RegisterModel<NumberModel>(
                 Culture.Arabic,
                 (options) => new NumberModel(
                     AgnosticNumberParserFactory.GetParser(AgnosticNumberParserType.Number, new ArabicNumberParserConfiguration(
                                                               new BaseNumberOptionsConfiguration(Culture.Arabic, options))),
                     Arabic.NumberExtractor.GetInstance(NumberMode.PureNumber, options)));
-            */
+
+            RegisterModel<OrdinalModel>(
```
tellarin: I assume this means most tests now pass in .NET for all three types?
If so, please update the PR description with the pass/fail(skipped) numbers per model, ok?
Otherwise, please un-register the models.

tellarin:
@ahmedabuamra build is failing. Can you fix it?

.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/TestNumber_Arabic.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Microsoft.Recognizers.Text.Number.Tests
+{
+    [TestClass]
+    public class TestNumber_Arabic : TestBase
+    {
+        public static TestResources TestResources { get; protected set; }
+
+        /*
+        [NetCoreTestDataSource]
```
wgx998877: don't need them?

tellarin: Temporarily disabled, as commented in the last comit.

tellarin: Temporarily disabled, as commented in the last commit.

Issue #1427, opened 2019-01-28
[DotNet] Move RegexExtensions to RegExpUtility


GasparAcevedoZainSouthworks:
**Is your feature request related to a problem? Please describe.**
It's a bit confusing how regexes work in different languages solutions.

**Describe the solution you'd like**
We need a unique entity that contains all the regexes functionality and that it's easily identified in all the languages implementations. As of today, we have either just RegexExtensions(.net) or RegexExtensions plus RegExpUtility(Java, JavaScript, and Python).

**Describe alternatives you've considered**
We should create (or rename one of the existing ones) a new entity that has all the regexes functionality currently being used in the project. This should have the exact same public methods than their pairs in every language. Then we should analyze all the solutions and refactor them to start using this new entity.

tellarin:
@GasparAcevedoZainSouthworks @Aliandi Could you folks retake this once the Python work is done?

Aliandi:
@tellarin sure, we'll add it to our backlog and take it when we finish Python

PR #1873, opened 2019-09-12
[DotNet] Create Utility class for Regular Expression

Aliandi:
Solves [ISSUE#1427](https://github.com/microsoft/Recognizers-Text/issues/1427)

### Description
This PR will centralize all Regular expressions related methods in a class called `RegExpUtility` in .NET. The idea here is to increase parity to match how it is already implemented in Java, JavaScript and Python. 

### Changes made

- We added four more methods from `RegexExtensions.cs`:
   - `IsExactMatch( )`
   - `MatchExact( )`
   - `MatchEnd( )`
   - `MatchBegin( )`
 
- We changed `MatchEvaluator( )` to `ReplaceMatchGroup( )` because that name was more descriptive

- We moved the `ConditionalMatch` class from `Microsoft.Recognizers.Text.DateTime` to `Microsoft.Recognizers.Text.Utilities`, so the methods we moved before do not break the existing references.
 
- We imported `Microsoft.Recognizers.Text.Utilities` everywhere the code had errors because of the classes and methods migration.

### Testing
Find below a screenshot of the running tests:

image


Aliandi:
@tellarin We fixed the build issue. It was due to an error on some folders on the python's master branch. This PR now contains the fixes for that.

PR #1875, opened 2019-09-18
Fix cases like "人民币5元" “美元 人民币” “dollar RMB”

### LINKED ISSUE #1865: [CN Currency] Parsing cases like "8亿元人民币" causes infinite loop
When passing  "8亿元人民币", "人民币8亿元", "人民币人民币3.50亿元'  these sentences to NumberWithUnitRecognizer, parse function couldn't stop.

I use python in this case

Grey0202:
Fixes #1865
Bugs include:
- Unify the Java decimal point format
- Fix the case "5元人民币" having two results


.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -73,6 +73,11 @@ private static bool CheckUnitsStringContains(string fractionUnitCode, string fra
 
         private string GetResolutionStr(object value)
         {
+            if (value == null)
```
tellarin: Please add a comment here: // Nothing to resolve. This happens when the entity is a currency name only (no numerical value).

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -97,6 +103,8 @@ private ParseResult MergeCompoundUnit(ExtractResult compoundResult)
                 var parseResultValue = parseResult.Value as UnitValue;
                 var unitValue = parseResultValue?.Unit;
 
+                extensibleResult = numberValue == null ? mainUnitValue : string.Empty;
```
tellarin: What is an extensibleResult ?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -163,6 +175,22 @@ private ParseResult MergeCompoundUnit(ExtractResult compoundResult)
                         // If the fraction unit doesn't match the main unit, finish process this group.
                         if (result != null)
                         {
+                            // If the fraction unit is same with main unit and main unit does not have numberValue, combine them.
+                            // For example "人民币50元".
```
tellarin: This doesn't seem a very general solution for compound currencies. If you don't  add this, what's the output?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -163,6 +175,22 @@ private ParseResult MergeCompoundUnit(ExtractResult compoundResult)
...
+                            if (extensibleResult == unitValue && parseResultValue.Number != null)
```
tellarin: Can't this be a method like MergeCompoundCurrency?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -188,6 +216,7 @@ private ParseResult MergeCompoundUnit(ExtractResult compoundResult)
 
                         count = 0;
                         idx -= 1;
+                        numberValue = null;
```
tellarin: Why?

Specs/NumberWithUnit/Chinese/CurrencyModel.json
```diff
@@ -333,5 +333,124 @@
...
+        "Resolution": {
+          "isoCurrency": "CNY",
+          "unit": "Chinese yuan",
+          "value": "50.63"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "人民币 美元",
```
tellarin: If Chinese doesn't use whitespace separators, why have it in this sentence?

Specs/NumberWithUnit/Chinese/CurrencyModel.json
```diff
@@ -333,5 +333,124 @@
...
+        "Resolution": {
+          "isoCurrency": "USD",
+          "unit": "United States dollar",
+          "value": null
+        }
+      }
+    ]
+  },
+  {
+    "Input": "4人民币 5美元",
```
tellarin: Same

Specs/NumberWithUnit/Chinese/CurrencyModel.json
tellarin: Please add some spec examples in English that show the previous issue is now fixed too.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -77,13 +81,20 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
+                }
+
                 result.setResolutionStr(parseResult.getResolutionStr());
                 mainUnitIsoCode = config.getCurrencyNameToIsoCodeMap().containsKey(unitValue) ? config.getCurrencyNameToIsoCodeMap().get(unitValue) : mainUnitIsoCode;
 
                 // If the main unit can't be recognized, finish process this group.
                 if (mainUnitIsoCode == null || mainUnitIsoCode.isEmpty()) {
-                    result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
+                    if (numberValue < 0) {
+                        result.setValue(new CurrencyUnitValue("null", mainUnitValue, mainUnitIsoCode));
```
tellarin: Why this?

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -120,11 +131,22 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
+                            numberValue = Double.parseDouble(parseResultValue.get().number);
+                            result.setValue(new CurrencyUnitValue(String.valueOf(numberValue), mainUnitValue, mainUnitIsoCode));
+                            continue;
+                        }
                         if (mainUnitIsoCode == null || mainUnitIsoCode.isEmpty() ||
                             mainUnitIsoCode.startsWith(Constants.FAKE_ISO_CODE_PREFIX)) {
                             result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
                         } else {
-                            result.setValue(new CurrencyUnitValue(String.valueOf(numberValue), mainUnitValue, mainUnitIsoCode));
+                            if (numberValue < 0) {
```
tellarin: Can't these duplicated pieces of code be a method?

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -145,7 +168,11 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
                 mainUnitIsoCode.startsWith(Constants.FAKE_ISO_CODE_PREFIX)) {
                 result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
             } else {
-                result.setValue(new CurrencyUnitValue(String.valueOf(numberValue), mainUnitValue, mainUnitIsoCode));
+                if (numberValue < 0) {
+                    result.setValue(new CurrencyUnitValue("null", mainUnitValue, mainUnitIsoCode));
```
tellarin: Again

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
Grey0202: It will get "人民币" and “50元”

Grey0202: > This doesn't seem a very general solution for compound currencies. If you don't add this, what's the output?

It will get "人民币" and “50元”

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
Grey0202: > Why?

Set this to the default value in case of next unit without value get previous unit value. Such as "50 dollar RMB", the RMB will get the 50 as value.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -123,6 +123,8 @@ private ParseResult MergeCompoundUnit(ExtractResult compoundResult)
                     };
 
                     mainUnitValue = unitValue;
+
+                    // Nothing to resolve. This happens when the entity is a currency name only (no numerical value).
```
tellarin: I don't think this is the right place for this comment. In the initial review, it was supposed to go where you had a ```return null;```

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -77,13 +84,17 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
+                    NF.format(numberValue);
+                }
+
                 result.setResolutionStr(parseResult.getResolutionStr());
                 mainUnitIsoCode = config.getCurrencyNameToIsoCodeMap().containsKey(unitValue) ? config.getCurrencyNameToIsoCodeMap().get(unitValue) : mainUnitIsoCode;
 
                 // If the main unit can't be recognized, finish process this group.
                 if (mainUnitIsoCode == null || mainUnitIsoCode.isEmpty()) {
-                    result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
+                    result.setValue(new UnitValue(numberValue < 0 ? null : NF.format(numberValue), mainUnitValue));
```
tellarin: Move the new formatting to it's own method and just call here similarly to how the previous String.valueOf was called.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -56,10 +56,12 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
 
         int count = 0;
         ParseResult result = null;
-        double numberValue = 0.0;
+        // Make the default numberValue negative to check if there is no Value.
+        double numberValue = -1;
```
tellarin: Can you declare this as a constant and use min double as it's value?

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -77,13 +79,16 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
+                    numberValue = Double.parseDouble(parseResultValue.get().number);
+                }
+
                 result.setResolutionStr(parseResult.getResolutionStr());
                 mainUnitIsoCode = config.getCurrencyNameToIsoCodeMap().containsKey(unitValue) ? config.getCurrencyNameToIsoCodeMap().get(unitValue) : mainUnitIsoCode;
 
                 // If the main unit can't be recognized, finish process this group.
                 if (mainUnitIsoCode == null || mainUnitIsoCode.isEmpty()) {
-                    result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
+                    result.setValue(new UnitValue(numberValue < 0 ? null : getNumberValue(numberValue), mainUnitValue));
```
tellarin: This check for the invalid number could be moved inside getNumberValue() and improved to check for exactly the new constant value.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -120,37 +125,22 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
-                        if (mainUnitIsoCode == null || mainUnitIsoCode.isEmpty() ||
-                            mainUnitIsoCode.startsWith(Constants.FAKE_ISO_CODE_PREFIX)) {
-                            result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
-                        } else {
-                            result.setValue(new CurrencyUnitValue(String.valueOf(numberValue), mainUnitValue, mainUnitIsoCode));
-                        }
-
-                        results.add(result);
-                        result = null;
+                        addToResults(result,mainUnitIsoCode,numberValue,mainUnitValue,results);
```
tellarin: Code standard. Please add the whitespaces after each comma.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -120,37 +125,22 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
                     }
+                    result = null;
 
                     count = 0;
                     idx -= 1;
+                    numberValue = -1;
```
tellarin: Constant.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -77,13 +79,16 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
                 result = new ParseResult(extractResult.getStart(), extractResult.getLength(), extractResult.getText(), extractResult.getType(), null, null, null);
 
                 mainUnitValue = unitValue;
-                numberValue = parseResultValue.isPresent() ? Double.parseDouble(parseResultValue.get().number) : 0;
+                if (parseResultValue.isPresent() && parseResultValue.get().number != null) {
```
tellarin: What's the benefit of this instead of the previous line?

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -120,37 +125,22 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
-                            result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
-                        } else {
-                            result.setValue(new CurrencyUnitValue(String.valueOf(numberValue), mainUnitValue, mainUnitIsoCode));
-                        }
-
-                        results.add(result);
-                        result = null;
+                        addToResults(result,mainUnitIsoCode,numberValue,mainUnitValue,results);
                     }
+                    result = null;
```
tellarin: This null assignment and the numberValue resest mean the same thing, right? Move this to the same location as in .cs
Nit: Also, add a blankline after the bracket close and before these assignments.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -120,37 +125,22 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
                     count = 0;
                     idx -= 1;
+                    numberValue = -1;
                     continue;
                 }
             }
 
             count++;
         }
-
```
tellarin: Don't remove this blank line. Always keep a blank line between unrelated if brackets.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -120,37 +125,22 @@ private ParseResult mergeCompoundUnit(ExtractResult compoundResult) {
...
         if (result != null) {
-            if (mainUnitIsoCode == null || mainUnitIsoCode.isEmpty() ||
-                mainUnitIsoCode.startsWith(Constants.FAKE_ISO_CODE_PREFIX)) {
-                result.setValue(new UnitValue(String.valueOf(numberValue), mainUnitValue));
-            } else {
-                result.setValue(new CurrencyUnitValue(String.valueOf(numberValue), mainUnitValue, mainUnitIsoCode));
-            }
-
-            results.add(result);
+            addToResults(result,mainUnitIsoCode,numberValue,mainUnitValue,results);
```
tellarin: Code standard.

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
```diff
@@ -171,4 +161,23 @@ private void resolveText(List<ParseResult> prs, String source, int bias) {
...
+
+    private String getNumberValue(double numberValue) {
+        java.text.NumberFormat numberFormat = java.text.NumberFormat.getInstance();
+        numberFormat.setMinimumFractionDigits(0);
+        numberFormat.setGroupingUsed(false);
+        return numberFormat.format(numberValue);
+    }
+
+    private  void addToResults(ParseResult result, String mainUnitIsoCode,
+                               Double numberValue, String mainUnitValue,List<ParseResult> results) {
```
tellarin: Code standard

JavaScript/packages/recognizers-number-with-unit/src/numberWithUnit/parsers.ts
```diff
@@ -254,13 +256,13 @@ export class BaseCurrencyParser implements IParser {
                     if (result !== null) {
                         if (StringUtility.isNullOrEmpty(mainUnitIsoCode) || mainUnitIsoCode.startsWith(Constants.FAKE_ISO_CODE_PREFIX)) {
```
tellarin: Shouldn't this be a method, like the one you created in Java?

Python/libraries/recognizers-number-with-unit/recognizers_number_with_unit/number_with_unit/parsers.py
```diff
@@ -213,26 +213,26 @@ def __merge_compound_unit(self, compound_result: ExtractResult) -> ParseResult:
                     if result:
                         if not main_unit_iso_code or main_unit_iso_code.startswith(Constants.FAKE_ISO_CODE_PREFIX):
```
tellarin: To a method, same as in Java?

Python/libraries/recognizers-number-with-unit/recognizers_number_with_unit/number_with_unit/parsers.py
```diff
@@ -213,26 +213,26 @@ def __merge_compound_unit(self, compound_result: ExtractResult) -> ParseResult:
...
-                    idx = idx - 1
+                    number_value = ''
                     continue
 
             count = count + 1
             idx = idx + 1
 
         if result:
             if not main_unit_iso_code or main_unit_iso_code.startswith(Constants.FAKE_ISO_CODE_PREFIX):
-                result.value = UnitValue(str(number_value), main_unit_value)
```
tellarin: Same

Java/libraries/recognizers-text-number-with-unit/src/main/java/com/microsoft/recognizers/text/numberwithunit/parsers/BaseCurrencyParser.java
Grey0202: > What's the benefit of this instead of the previous line?

It could be null and break the parser.

Specs/NumberWithUnit/Chinese/CurrencyModel.json
```diff
@@ -334,23 +334,6 @@
       }
     ]
   },
-  {
-    "Input": "50元6角三分",
```
tellarin: Why remove this case? If not currently supported, but it will be supported in the future, mark it as not supported in all platforms.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -73,19 +73,50 @@ private static bool CheckUnitsStringContains(string fractionUnitCode, string fra
...
+            {
+                return null;
+            }
+
             return Config.CultureInfo != null ?
                 ((double)value).ToString(Config.CultureInfo) :
                 value.ToString();
         }
 
+        private void AddToResult(ParseResult result, string mainUnitIsoCode, object numberValue,
```
tellarin: Can you move results.Add(result); out of the method? The method could be called something like CreateCurrencyResult and return a result.
It is a bad design to have methods that internally change the state of one passed parameter. Especially if the parameter is not really necessary to the method body.  If a parameter is really necessary, it should at least be explicitly marked as being used by reference.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -73,19 +73,50 @@ private static bool CheckUnitsStringContains(string fractionUnitCode, string fra
...
+            string mainUnitValue, List<ParseResult> results)
+        {
+            if (string.IsNullOrEmpty(mainUnitIsoCode) ||
+                mainUnitIsoCode.StartsWith(Constants.FAKE_ISO_CODE_PREFIX, StringComparison.Ordinal))
+            {
+                result.Value = new UnitValue
```
tellarin: Why not always create CurrencyUnitValue objects? Some with null IsoCurrency?

PR #1884, opened 2019-09-27
[.NET] Turkish DateTime DatePeriod support

LionbridgeCSII:
DatePeriodExtractor : pass: 192 | fail: 1
DatePeriodParser: pass: 173 | fail: 1

Failing case:
"Bu görev dünden 2 gün önce yapılmalıydı." 
This case corresponds to "This task should have been done more than 2 days before yesterday"
but it translates to "This task should have been done 2 days before yesterday"
and there is no meaningful way to express the concept of "more than" in Turkish in this sentence.

Other comments:
There is no difference between "in" and "within" in Turkish, both translate to "içinde", so the case
"[iki hafta içinde] bir toplantı ayarla" 
can be translated to both
"schedule a meeting [in two weeks]" and "schedule a meeting [within two weeks]".
Therefore, we added the result field to this case. 
 
The Date case 
"Aynı gün döneceğim" 
corresponding to "I'll go back the day" has been changed to "O gün döneceğim" which is a more accurate translation. "Aynı" means "the same" and has a different function in DatePeriod.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
@@ -69,24 +69,28 @@ namespace Microsoft.Recognizers.Text.DateTime
 
                 // within "Days/Weeks/Months/Years" should be handled as dateRange here
                 // if duration contains "Seconds/Minutes/Hours", it should be treated as datetimeRange
-                var match = config.WithinNextPrefixRegex.MatchEnd(beforeStr, trim: true);
-
-                if (match.Success)
+                bool inPrefix = true;
+                MatchWithinNextPrexixRegex(text, duration, out int startToken, out int endToken, out bool matchSuccess, inPrefix);
+                if (matchSuccess)
                 {
-                    var startToken = match.Index;
-                    var durationStr = text.Substring(duration.Start, duration.Length);
-                    var matchDate = config.DateUnitRegex.Match(durationStr);
-                    var matchTime = config.TimeUnitRegex.Match(durationStr);
+                    ret.Add(new Token(startToken, endToken));
+                    continue;
+                }
 
-                    if (matchDate.Success && !matchTime.Success)
+                // check also afterStr
+                if (this.config.CheckBothBeforeAfter)
+                {
+                    inPrefix = false;
+                    MatchWithinNextPrexixRegex(text, duration, out startToken, out endToken, out matchSuccess, inPrefix);
+                    if (matchSuccess)
                     {
-                        ret.Add(new Token(startToken, duration.End));
+                        ret.Add(new Token(startToken, endToken));
                         continue;
                     }
                 }
 
                 // Match prefix
-                match = this.config.PreviousPrefixRegex.MatchEnd(beforeStr, trim: true);
+                var match = this.config.PreviousPrefixRegex.MatchEnd(beforeStr, trim: true);
 
                 var index = -1;
 
@@ -170,6 +174,20 @@ namespace Microsoft.Recognizers.Text.DateTime
             {
                 var startIndex = text.LastIndexOf(match.Value);
                 ret.Add(new Token(startIndex, (int)er.Start + (int)er.Length));
+            }
+
+            return ret;
+        }
+
+        // Get token in after string
+        private static List<Token> GetTokenForRegexMatchingAfter(string text, Regex regex, ExtractResult er, int matchStart, int matchLength)
+        {
+            var ret = new List<Token>();
+            var match = regex.Match(text);
+            if (match.Success && text.Trim().StartsWith(match.Value.Trim()))
+            {
+                var length = match.Index + match.Length;
+                ret.Add(new Token(matchStart, matchStart + matchLength + length));
             }
 
             return ret;
@@ -595,6 +613,18 @@ namespace Microsoft.Recognizers.Text.DateTime
                         idx += 2;
                         continue;
                     }
+
+                    // handle "between...and..." case when "between" follows the datepoints
+                    var afterStr = text.Substring(periodEnd, text.Length - periodEnd);
+                    if (this.config.GetBetweenTokenIndex(afterStr, out int afterIndex))
+                    {
+                        periodEnd += afterIndex;
+                        ret.Add(new Token(periodBegin, periodEnd, metadata));
+
+                        // merge two tokens here, increase the index by two
+                        idx += 2;
+                        continue;
+                    }
                 }
 
                 idx++;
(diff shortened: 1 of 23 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDatePeriodExtractorConfiguration.cs
```diff
@@ -168,6 +171,9 @@ public class TurkishDatePeriodExtractorConfiguration : BaseDateTimeOptionsConfig
 
         private const RegexOptions RegexFlags = RegexOptions.Singleline | RegexOptions.ExplicitCapture;
 
+        private static readonly Regex SuffixRegex =
+            new Regex(@"^'\p{L}*(?<match>.*)$", RegexFlags);
```
tellarin: Please move the regex to the YAML file.

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDatePeriodExtractorConfiguration.cs
```diff
@@ -311,14 +317,16 @@ public TurkishDatePeriodExtractorConfiguration(IDateTimeOptionsConfiguration con
...
 
+        bool IDatePeriodExtractorConfiguration.CheckBothBeforeAfter => DateTimeDefinitions.CheckBothBeforeAfter;
+
         string[] IDatePeriodExtractorConfiguration.DurationDateRestrictions => DateTimeDefinitions.DurationDateRestrictions;
 
         public bool GetFromTokenIndex(string text, out int index)
         {
             index = -1;
-            if (text.EndsWith("from"))
+            if (text.EndsWith("itibaren"))
```
tellarin: Please also extract such hardcoded terms to YAML. English has some examples of it being done.

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDatePeriodExtractorConfiguration.cs
```diff
@@ -328,9 +336,27 @@ public bool GetFromTokenIndex(string text, out int index)
...
+            var match = RangePrefixRegex.MatchEnd(text, false);
+
+            if (match.Success)
+            {
+                index = match.Index;
+                return true;
+            }
+
+            string textTrm = text;
+            if (text.StartsWith("'"))
```
tellarin: What is this representing now?

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDatePeriodExtractorConfiguration.cs
LionbridgeCSII: moved and renamed 'ExcludeSuffixRegex'

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDatePeriodExtractorConfiguration.cs
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDatePeriodExtractorConfiguration.cs
LionbridgeCSII: refactored this part to use only ExcludeSuffixRegex, its function is to exclude the suffix (if any) when looking for RangePrefixRegex matches

LionbridgeCSII:
@tellarin, merge conflicts solved.

LionbridgeCSII:
@tellarin, the PR is mergeable.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -682,5 +726,63 @@ private bool IsDateRelativeToNowOrToday(ExtractResult er)
 
             return false;
         }
+
+        // Matches "within (the next)?" part (in beforeStr or afterStr) in "within Days/Weeks/Months/Years"
+        private void MatchWithinNextPrexixRegex(string text, Token duration, out int startToken, out int endToken, out bool matchSuccess, bool inPrefix)
```
tellarin: This should preferably return some structure, instead of multiple 'out' parameters.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -595,6 +613,18 @@ private List<Token> MergeMultipleExtractions(string text, List<ExtractResult> ex
                         idx += 2;
                         continue;
                     }
+
+                    // handle "between...and..." case when "between" follows the datepoints
+                    var afterStr = text.Substring(periodEnd, text.Length - periodEnd);
+                    if (this.config.GetBetweenTokenIndex(afterStr, out int afterIndex))
```
tellarin: Shouldn't this also check for the before/after flag? Or should it really run everytime?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -682,5 +726,63 @@ private bool IsDateRelativeToNowOrToday(ExtractResult er)
...
+                    if (match.Success)
+                    {
+                        startToken = match.Index;
+                    }
+
+                }
+            }
+        }
+
+        private List<Token> ExtractWithinNextPrefix(string subStr, bool inPrefix, ExtractResult extractionResult)
```
tellarin: This is just basically an ExtractToMethod refactor, correct?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
```diff
@@ -175,6 +179,20 @@ private static List<Token> GetTokenForRegexMatching(string text, Regex regex, Ex
             return ret;
         }
 
+        // Get token in after string
+        private static List<Token> GetTokenForRegexMatchingAfter(string text, Regex regex, ExtractResult er, int matchStart, int matchLength)
```
tellarin: I'm not sure I get why this method is written this way? Can you give examples of it and of the regular GetTokenForRegexMatching?
It seems to me both should be combined in some way.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
LionbridgeCSII: ok, modified to return directly a Token

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
LionbridgeCSII: added check for before/after flag.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
LionbridgeCSII: yes, as you suggested in the Date PR, it is used to avoid repeating the same piece of code for beforeStr and afterStr.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDatePeriodExtractor.cs
LionbridgeCSII: the two methods are now combined in GetTokenForRegexMatching

PR #1892, opened 2019-10-09
[.NET] Turkish DateTime TimePeriod support

LionbridgeCSII:
All test cases pass.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Turkish.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Turkish.cs
@@ -112,39 +112,6 @@ namespace Microsoft.Recognizers.Text.DateTime.Tests
         public void TimeParser(TestModel testSpec)
         {
             ExtractorInitialize(Extractors);
-            TestDateTimeExtractor(testSpec);
-        }
-
-        [NetCoreTestDataSource]
-        [TestMethod]
-        public void DatePeriodParser(TestModel testSpec)
-        {
-            ExtractorInitialize(Extractors);
-            ParserInitialize(Parsers);
-            TestDateTimeParser(testSpec);
-        }
-
-        [NetCoreTestDataSource]
-        [TestMethod]
-        public void DurationExtractor(TestModel testSpec)
-        {
-            ExtractorInitialize(Extractors);
-            TestDateTimeExtractor(testSpec);
-        }
-
-        [NetCoreTestDataSource]
-        [TestMethod]
-        public void DurationParser(TestModel testSpec)
-        {
-            ExtractorInitialize(Extractors);
-            ParserInitialize(Parsers);
-            TestDateTimeParser(testSpec);
-        }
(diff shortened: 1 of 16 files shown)


.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Turkish.cs
```diff
@@ -37,10 +109,47 @@ public void DateParser(TestModel testSpec)
 
         [NetCoreTestDataSource]
         [TestMethod]
-        public void HolidayExtractor(TestModel testSpec)
+        public void TimeParser(TestModel testSpec)
```
tellarin: Why is the order of methods in this file changing so much? All test method calls should already exist here and be running the disabled tests, no?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
```diff
@@ -276,6 +277,11 @@ private List<Token> MergeTwoTimePoints(string text, DateObject reference)
                         // Handle "between"
                         periodBegin = betweenIndex;
                     }
+                    else if (this.config.GetBetweenTokenIndex(afterStr, out var afterIndex))
```
tellarin: Only if the before/after flag is on?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
```diff
@@ -297,6 +303,18 @@ private List<Token> MergeTwoTimePoints(string text, DateObject reference)
                         idx += 2;
                         continue;
                     }
+
+                    // handle "between...and..." case when "between" follows the datepoints
+                    var afterStr = text.Substring(periodEnd, text.Length - periodEnd);
+                    if (this.config.GetBetweenTokenIndex(afterStr, out int afterIndex))
```
tellarin: Same?

.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Turkish.cs
LionbridgeCSII: the test methods were not all included until this PR, so the file will not change from now on.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
LionbridgeCSII: yes, flag added

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
LionbridgeCSII: flag added

PR #1893, opened 2019-10-10
[.NET] Turkish DateTime DateTime support

LionbridgeCSII:
All test cases pass.


LionbridgeCSII:
@tellarin, merge conflicts resolved. 
It is not clear what is meant by "add the DateTimeModel.json cases in this PR"

tellarin:
@LionbridgeCSII all the tests in these PRs are for the Extractor and Parsers, but the tests at the model level are missing. These are in DateTimeModel.json and the like.
I made a mistake in that comment, this should not be part of the current PR. It should come either in the PR with the last DateTime sub-type, or at its own PR after that one merges.

PR #1898, opened 2019-10-14
[DE Unit] Update Temperature entries

dNsl9r:
Added entries for kelvin, celsius for better results in LUIS


dNsl9r:
Hi @tellarin - I'm not very familiar with the build process, I will take at look into it when I've got time and will process the requested changes. Thank you!

tellarin:
@dNsl9r adding the test spec cases is the most important, as they should be proper German full sentences. 
We can help with the auto-gen resource files. Thanks!

tellarin:
@dNsl9r, thanks for the specs. I've fixed the json syntax and added the latest auto-gen resource files.
It should be OK to merge as soon as the build finishes.

dNsl9r:
thank you!

PR #1902, opened 2019-10-15
[.NET] Turkish DateTime DateTimePeriod support

LionbridgeCSII:
All test cases pass.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Recognizers.Text.DateTime
 
                     bool inPrefix = true;
                     int length = GetValidConnectorForDateAndTimePeriod(middleStr, inPrefix);
-                    if (length >= 0)
+                    if (length != Constants.INVALID_CONNECTOR_CODE)
                     {
                         var begin = ers[i].Start ?? 0;
                         var end = (ers[j].Start ?? 0) + (ers[j].Length ?? 0);
@@ -118,7 +118,7 @@ namespace Microsoft.Recognizers.Text.DateTime
                         var afterStart = ers[j].Start + ers[j].Length ?? 0;
                         var afterStr = text.Substring(afterStart);
                         length = GetValidConnectorForDateAndTimePeriod(afterStr, inPrefix);
-                        if (length >= 0 && middleStr.Length <= 4)
+                        if (length != Constants.INVALID_CONNECTOR_CODE && middleStr.Length <= 4)
                         {
                             var begin = ers[i].Start ?? 0;
                             var end = (ers[j].Start ?? 0) + (ers[j].Length ?? 0) + length;
@@ -152,7 +152,7 @@ namespace Microsoft.Recognizers.Text.DateTime
         // Valid connector in English for After include: "after", "later than"
         private int GetValidConnectorForDateAndTimePeriod(string text, bool inPrefix)
         {
-            int length = -1;
+            int length = Constants.INVALID_CONNECTOR_CODE;
             var beforeAfterRegexes = new List<Regex>
             {
                 this.config.BeforeRegex,
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
@@ -104,7 +104,8 @@ namespace Microsoft.Recognizers.Text.DateTime
                     var middleStr = text.Substring(middleBegin, middleEnd - middleBegin).Trim();
 
                     bool inPrefix = true;
-                    if (IsValidConnectorForDateAndTimePeriod(middleStr, out int length, inPrefix))
+                    int length = GetValidConnectorForDateAndTimePeriod(middleStr, inPrefix);
+                    if (length >= 0)
                     {
                         var begin = ers[i].Start ?? 0;
                         var end = (ers[j].Start ?? 0) + (ers[j].Length ?? 0);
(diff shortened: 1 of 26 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Parsers/TurkishDateTimePeriodParserConfiguration.cs
```diff
@@ -24,7 +24,8 @@ public class TurkishDateTimePeriodParserConfiguration : BaseDateTimeOptionsConfi
         public TurkishDateTimePeriodParserConfiguration(ICommonDateTimeParserConfiguration config)
             : base(config)
         {
-            TokenBeforeDate = DateTimeDefinitions.TokenBeforeDate;
+            // @TODO replace token with list of tokens (in Turkish many different suffixes can be used)
+            TokenBeforeDate = DateTimeDefinitions.TokenBeforeDate1;
```
tellarin: Will this happen before this PR can merge?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
```diff
@@ -133,8 +148,9 @@ private IEnumerable<Token> MergeDateWithTimePeriodSuffix(string text, List<Extra
         // Cases like "today after 2:00pm", "1/1/2015 before 2:00 in the afternoon"
         // Valid connector in English for Before include: "before", "no later than", "in advance of", "prior to", "earlier than", "sooner than", "by", "till", "until"...
         // Valid connector in English for After include: "after", "later than"
-        private bool IsValidConnectorForDateAndTimePeriod(string text)
+        private bool IsValidConnectorForDateAndTimePeriod(string text, out int length, bool inPrefix)
```
tellarin: If you need the length as output, I'd suggest returning the length, and return an error code (-1) when false.
You can declare a Constants entry for this. Maybe INVALID_CONNECTOR_CODE.

tellarin: It's OK to rename such method to GetValidConnectorIndexForDateAndTimePeriod
and declare a new IsValidConnectorForDateAndTimePeriod that just checks the returned index or code and still returns a bool.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
```diff
@@ -637,5 +672,41 @@ private List<Token> MatchRelativeUnit(string text)
...
+                    var beforeStr = text.Substring(0, duration.Start);
+                    var matchNext = this.config.NextPrefixRegex.Match(beforeStr);
+                    success = match.Success || matchNext.Success;
+                    if (MatchPrefixRegexInSegment(beforeStr, matchNext, true))
+                    {
+                        startToken = matchNext.Index;
+                    }
+                }
+
+                if (success)
```
tellarin: What happens if not a success?

LionbridgeCSII: if not a success, the method returns a token that starts and ends at -1. When the method is called, it is checked if the returned token.Start is >=0.

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Parsers/TurkishDateTimePeriodParserConfiguration.cs
LionbridgeCSII: yes, updated this and modified BaseDateTimePeriodParser to work with a list of tokens (when before/after flag is true)

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
LionbridgeCSII: (the length is needed only when the before/after flag is true).
Modified the method to return the length. 
When the method is called, it is checked if the returned length is >=0 (as for the method in the comment below)

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
```diff
@@ -148,9 +150,9 @@ private IEnumerable<Token> MergeDateWithTimePeriodSuffix(string text, List<Extra
         // Cases like "today after 2:00pm", "1/1/2015 before 2:00 in the afternoon"
         // Valid connector in English for Before include: "before", "no later than", "in advance of", "prior to", "earlier than", "sooner than", "by", "till", "until"...
         // Valid connector in English for After include: "after", "later than"
-        private bool IsValidConnectorForDateAndTimePeriod(string text, out int length, bool inPrefix)
+        private int GetValidConnectorForDateAndTimePeriod(string text, bool inPrefix)
         {
-            length = 0;
+            int length = -1;
```
tellarin: Please declare a constant for it, and use it here and in the return checks.

Patterns/Turkish/Turkish-DateTime.yaml
```diff
@@ -123,7 +123,7 @@ DateUnitRegex: !simpleRegex
 DateTokenPrefix: ''
 TimeTokenPrefix: ''
 TokenBeforeDate: ''
```
tellarin: Can you add a comment here in the YAML file explaining why TokenBeforeDate needs to stay empty and TokenListBeforeDate populated for DateTimePeriod?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
LionbridgeCSII: done.

Patterns/Turkish/Turkish-DateTime.yaml
LionbridgeCSII: Comment added.

PR #1903, opened 2019-10-15
[.NET] Turkish DateTime Set support

LionbridgeCSII:
All test cases pass.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Parsers/TurkishSetParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Parsers/TurkishSetParserConfiguration.cs
@@ -1,11 +1,18 @@
 ﻿using System.Collections.Immutable;
 using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions.Turkish;
 using Microsoft.Recognizers.Text.DateTime.Utilities;
+using Microsoft.Recognizers.Text.Utilities;
 
 namespace Microsoft.Recognizers.Text.DateTime.Turkish
 {
     public class TurkishSetParserConfiguration : BaseDateTimeOptionsConfiguration, ISetParserConfiguration
     {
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Parsers/TurkishSetParserConfiguration.cs
```diff
@@ -80,23 +80,23 @@ public bool GetMatchedDailyTimex(string text, out string timex)
         {
             var trimmedText = text.Trim();
 
-            if (trimmedText.Equals("daily"))
+            if (trimmedText.Equals("günlük"))
```
tellarin: Please move such hardcoded values to YAML.

Patterns/Turkish/Turkish-DateTime.yaml
```diff
@@ -123,6 +123,7 @@ DateUnitRegex: !simpleRegex
 DateTokenPrefix: ''
 TimeTokenPrefix: ''
 TokenBeforeDate: ''
+TokenBeforeDate1: '''de'
```
tellarin: Why declare a new entry here?

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Parsers/TurkishSetParserConfiguration.cs
LionbridgeCSII: done (added named groups to PeriodicRegex and DateUnitRegex and used those regexes to extract the values).

Patterns/Turkish/Turkish-DateTime.yaml
LionbridgeCSII: To reduce merge conflicts, the yaml file already includes the changes from the DateTimePeriod PR. 
This new entry (now changed to TokenListBeforeDate) is required only by DateTimePeriodParser, while all other subtypes need to use the standard empty TokenBeforeDate.

PR #1906, opened 2019-10-20
[PhoneNumber] Add filter for "bank account" prefix

Grey0202:
- Add filter for prefix like "account number" "account #" "account number:" "account"...
- Only in English and PT for now.


Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -4730,5 +4730,32 @@
...
+  {
+    "Input": "account #12345678",
+    "Results": []
+  },
+  {
+    "Input": "account 12345678",
+    "Results": []
+  },
+  {
+    "Input": "card number 1234567 and phone number is +55 11 12345-6789",
```
tellarin: Please also add a more illustrative case like "card number 12345678 and phone number 12345678"

Specs/Sequence/Portuguese/PhoneNumberModel.json
```diff
@@ -944,5 +944,21 @@
         }
       }
     ]
+  },
+  {
+    "Input": "conta número 1234567",
+    "Results": []
+  },
+  {
+    "Input": "conta bancária: 12345678",
```
tellarin: Please add "conta bancária número: 12345678"

tellarin: And "minha conta é 12345678"

Specs/Sequence/English/PhoneNumberModel.json
tellarin: And "My card number is 12345678"

Patterns/English/English-PhoneNumbers.yaml
```diff
@@ -0,0 +1,5 @@
+---
+NumberReplaceToken: '@builtin.phonenumber'
+ForbiddenPrefixRegex: !simpleRegex
```
tellarin: I think this should be renamed to FalsePositivePrefixRegex. And we can also add a named group to represent what kind of false positive (in this case, card number) as in the future we may have multiple.

Patterns/English/English-PhoneNumbers.yaml
```diff
@@ -0,0 +1,5 @@
...
+  def: (account|card)(\s+(#|number))?:?\s*$
```
tellarin: In the specs I mentioned a couple other cases we want to filter, so the regexes would need to be tweaked.

Patterns/Portuguese/Portuguese-PhoneNumbers.yaml
```diff
@@ -0,0 +1,5 @@
+---
+NumberReplaceToken: '@builtin.phonenumber'
+ForbiddenPrefixRegex: !simpleRegex
```
tellarin: Similar comments to English.

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -72,7 +73,7 @@ public IModel GetPhoneNumberModel(string culture = null, bool fallbackToDefaultC
                 return GetModel<PhoneNumberModel>(Culture.Chinese, fallbackToDefaultCulture);
             }
 
-            return GetModel<PhoneNumberModel>(Culture.English, fallbackToDefaultCulture);
+            return GetModel<PhoneNumberModel>(culture, fallbackToDefaultCulture);
```
tellarin: This change effectively breaks every other language. We should implement all cultures, or make any that's not EN and PT still fallback to English.

.NET/Microsoft.Recognizers.Text.Sequence/Portuguese/PortuguesePhoneNumberExtractorConfiguration.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions;
+using Microsoft.Recognizers.Definitions.Portuguese;
+
+namespace Microsoft.Recognizers.Text.Sequence.Portuguese
+{
+    public class PortuguesePhoneNumberExtractorConfiguration : PhoneNumberConfiguration
```
tellarin: Can't we have a BasePhoneNumberExtractorConfiguration and each language extends it?
PT only needs to load the new regex differently from base.

.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BasePhoneNumberExtractor.cs
```diff
@@ -111,6 +111,17 @@ public override List<ExtractResult> Extract(string text)
...
+
+                    if (this.config.ForbiddenPrefixRegex != null &&
+                            this.config.ForbiddenPrefixRegex.IsMatch(front))
+                    {
+                        ers.Remove(er);
+                        i--;
+                        continue;
+                    }
+
+                    front = text.Substring(0, (int)(er.Start - 1));
```
tellarin: Nit: These small changes will make it be called more often than before, which slowly can lead to performance degradation. If not really necessary, I recommend not making such changes in unrelated PRs.

tellarin: And why "front" twice with one index difference? Please try to make the code more readable.

PR #1907, opened 2019-10-20
[SV Numbers] Support for Numbers

pete1854:
This PR adds support for numbers. 
Number model tests Pass/fail/skipped status: 132/0/16
Some tests have been modified to reflect the Swedish language syntax.
This model should be good to go.


tellarin:
@pete1854, this looks quite good. Thanks for the contribution!
Can you mention in the PR description the number of test cases per type that pass/fail (disabled)?
Did you translate all cases from the English spec?

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
```diff
@@ -378,6 +379,20 @@ protected override void InitializeConfiguration()
                     new BaseNumberRangeParser(new HindiNumberRangeParserConfiguration(
                                                   new BaseNumberOptionsConfiguration(Culture.Hindi, options))),
                     new Hindi.NumberRangeExtractor(new BaseNumberOptionsConfiguration(Culture.Hindi, options))));*/
+
+            RegisterModel<NumberModel>(
```
tellarin: Can you add those pass/fail numbers in the description? Models should only be registered when ready-to-ship and we prefer to merge soon and update, if the current coverage of tests is already good.

pete1854:
Test status: Ordinal: Compound ordinals in the hundreds, eg. "Hundratjugosjunde" fails whilst "127:e" and "Hundrade" works. The same goes numbers in the thousands, milllions and so on.
Another exception is "Tredje från slutet" that I haven't fixed yet. The others should work fine.

Numbers: Everything should be covered. Failing tests are tests with whole numbers and fractions, e.g. "En och en halv", "Två och två tredjedelar" that currently resolves to "1" and "1,33" respectively. This error is because currently the parser divides the first two numbers Before adding the fraction part. I'm working on fixing this now.

pete1854:
@tellarin   The build failed because for some reason the file encoding gets messed up if editing the file directly in the WebUI… I think I'll delete this PR and submit a new one instead.

pete1854:
@tellarin I will review my patterns and specs and maybe refactor in some instances. Right now, as I've laid things out, it's quite cumbersome to track what cases are covered and which are not. I'll make some fresh commits shortly to be able to merge this PR.

tellarin:
Thanks! Let us know if some form of voice call could also be of help.

pete1854:
Great! I’m working at the Municipality of Simrishamn, Sweden and we plan to implement a Helpdesk Assistant (basen on the Bot FX). We need to have LUIS better recognize swedish utterances (and extract entities). So, I’m trying to get sv-SE into the recognizer. I’m working solo on this and only got 24h a day 🙂 But, we’ll get there! Btw how long does it usually take for approved new cultures to propagate to LUIS?

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
iMicknl: @tellarin, would it be possible to create a small CLI tool that will output those numbers? I had a hard time looking for those numbers during the work on the Dutch language recognizer. Most of the time I was just counting it manually...

(and it would be awesome if you could run the tool and it would mark testcases as supported, if they are supported)

tellarin: This is a great idea and we'll look into adding that. But it may take a while due to other demands.

pete1854: @tellarin - I decided to pull the Ordinal and Percent models from this PR. I will submit the percent model shortly, followed by the ordinal model. I consider this PR be ready for merging.

Specs/Number/Swedish/PercentModel.json
```diff
@@ -0,0 +1,138 @@
+[
+  {
+    "Input": "100%",
+    "Results": [
```
tellarin: For any new specs for unsupported types under implementation in .NET, we should add:
and for the cases not currently runnable on .NET, "NotSupported": "dotnet". This avoid unintended bug breaks in other platforms and a little easier tracking of what is still unsupported.
I'm OK with merging this PR as-is as it doesn't cause any breaks, but in the future PRs could you add these annotations?

tellarin:
> ... Btw how long does it usually take for approved new cultures to propagate to LUIS?

This is something that needs to be discussed with the LUIS team. Unfortunately we can't give an estimate.

PR #1913, opened 2019-10-25
[.NET] Hindi Numbers Cardinal support

LionbridgeCS:
Support for Numbers Cardinal in Hindi. 
190 pass, 2 skipped, 1 (तीन और पांचवां भाग) removed due to vagueness in translation. 
Added 22 new cases with code-mixing, Indian Numbering System and Devanagari numerals.
 
- The skips: 
[40 000 का मतलब हिंदी में 40000 नहीं होता] => 40 000 not valid in Hindi as they denote two different numbers, needs change in BaseNumber. 
[मैं आपको ३ सौ और २१ युआन दे सकता हूं।] => Issue in extracting composite Devanagari with "और" (will be worked on in future PRs)
********************************************************************************************************
- Explanation of peculiarities in Hindi Number and reasoning for changes made: 

- में से - These type of cases are very common in Hindi. It belongs to fraction unit type. Here any word/char at left of में से acts as denominator and right of it acts as numerator. 
"तीन में से एक" --- third(तीन) (एक)one 
"इक्कीस में से 1" --- (इक्कीस)twenty one  
- Behaviour changes of और - In fraction cases where और is used to connect two words and may not be used as addition of two words from its left and right.  
"तीन और एक का पाँचवाँ भाग" 

- Behaviour of "th" in English vs Hindi: In Hindi "th" is translated as वाँ/वां/वीं/वें which is resolved in two ways:
1) As Ordinal : If it is Ordinal then its value is considered as cardinal. 
2) As a Fraction: if it occurs in fraction context then the word वाँ is always considered a denominator.
बीस और तीन बटे पाँच 
पाँचवाँ 
- Commonly used cases: In Hindi there are commonly used decimal terms such as डेढ़, सवा and ढाई etc. They are basically fraction numbers which are Indian specific. 
- In Hindi, all double digit numbers are formed similar to English but there are specific names (with some variations) for each of the digits from 0 to 100 and they are not formed by its base number translation + zeroToNine number translation.
इकतालीस 
*******************************************************************************************************
Notes: 
- The refactored BaseIndianNumberParser will be useful for the other Indo-Aryan languages.  
*******************************************************************************************************
Queries: 
- If the current handling of composite numbers in the dictionaries is ok we have an extensive list that we can include for better coverage.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/LongFormTestConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/LongFormTestConfiguration.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Recognizers.Text.Number.Tests
             this.CardinalNumberMap = ImmutableDictionary<string, long>.Empty;
             this.OrdinalNumberMap = ImmutableDictionary<string, long>.Empty;
             this.RoundNumberMap = ImmutableDictionary<string, long>.Empty;
+            this.HindiDecimalUnitsList = ImmutableDictionary<string, double>.Empty;
             this.DigitalNumberRegex = new Regex(
                 @"((?<=\b)(hundred|thousand|million|billion|trillion|dozen(s)?)(?=\b))|((?<=(\d|\b))(k|t|m|g|b)(?=\b))", RegexOptions.Singleline);
         }
@@ -33,6 +34,8 @@ namespace Microsoft.Recognizers.Text.Number.Tests
         public ImmutableDictionary<string, long> OrdinalNumberMap { get; }
 
         public ImmutableDictionary<string, long> RoundNumberMap { get; }
+
+        public ImmutableDictionary<string, double> HindiDecimalUnitsList { get; }
 
         public CultureInfo CultureInfo { get; }
 
@@ -75,6 +78,11 @@ namespace Microsoft.Recognizers.Text.Number.Tests
         }
 
         public long ResolveCompositeNumber(string numberStr)
+        {
+            throw new NotImplementedException();
+        }
+
+        public double ResolveUnitCompositeNumber(string numberStr)
         {
             throw new NotImplementedException();
         }
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/LongFormTestConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/LongFormTestConfiguration.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Recognizers.Text.Number.Tests
             this.CardinalNumberMap = ImmutableDictionary<string, long>.Empty;
             this.OrdinalNumberMap = ImmutableDictionary<string, long>.Empty;
             this.RoundNumberMap = ImmutableDictionary<string, long>.Empty;
-            this.HindiDecimalUnitsList = ImmutableDictionary<string, double>.Empty;
             this.DigitalNumberRegex = new Regex(
                 @"((?<=\b)(hundred|thousand|million|billion|trillion|dozen(s)?)(?=\b))|((?<=(\d|\b))(k|t|m|g|b)(?=\b))", RegexOptions.Singleline);
         }
@@ -34,8 +33,6 @@ namespace Microsoft.Recognizers.Text.Number.Tests
         public ImmutableDictionary<string, long> OrdinalNumberMap { get; }
 
         public ImmutableDictionary<string, long> RoundNumberMap { get; }
-
-        public ImmutableDictionary<string, double> HindiDecimalUnitsList { get; }
 
         public CultureInfo CultureInfo { get; }
 
@@ -78,11 +75,6 @@ namespace Microsoft.Recognizers.Text.Number.Tests
         }
 
         public long ResolveCompositeNumber(string numberStr)
-        {
-            throw new NotImplementedException();
-        }
-
-        public double ResolveUnitCompositeNumber(string numberStr)
         {
             throw new NotImplementedException();
         }
--- a/.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
@@ -8,8 +8,6 @@ namespace Microsoft.Recognizers.Text.Number
 {
     public class BaseIndianNumberParser : BaseNumberParser
     {
-        private readonly bool isMultiDecimalSeparatorCulture = false;
-
         private readonly bool isCompoundNumberLanguage = false;
 
         public BaseIndianNumberParser(INumberParserConfiguration config)
@@ -17,7 +15,6 @@ namespace Microsoft.Recognizers.Text.Number
         {
             this.Config = config as IIndianNumberParserConfiguration;
 
-            this.isMultiDecimalSeparatorCulture = config.IsMultiDecimalSeparatorCulture;
             this.isCompoundNumberLanguage = config.IsCompoundNumberLanguage;
 
             TextNumberRegex = BuildTextNumberRegex();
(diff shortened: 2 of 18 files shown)


.NET/Microsoft.Recognizers.Text.DataDrivenTests/Number/LongFormTestConfiguration.cs
```diff
@@ -16,6 +16,7 @@ public LongFormTestConfiguration(char decimalSep, char nonDecimalSep)
             this.CardinalNumberMap = ImmutableDictionary<string, long>.Empty;
             this.OrdinalNumberMap = ImmutableDictionary<string, long>.Empty;
             this.RoundNumberMap = ImmutableDictionary<string, long>.Empty;
+            this.HindiDecimalUnitsList = ImmutableDictionary<string, double>.Empty;
```
tellarin: Why this? It should use a form that's not language specific, or need to be overwritten in specific languages.
But a base attribute can't directly reference a specific culture.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -597,6 +598,24 @@ protected double GetDigitalValue(string digitsStr, double power)
                 {
                     isNegative = true;
                 }
+
+                // Used for Hindi to resolve Devanagari numerals, C# char.isDigit accouts for Devanagari Numbers
```
tellarin: Please never break a if/else with comments in the middle. Comments can always go inside the clause.
However, such changes shouldn't be added to base code if Hindi-specific. Hindi could extend BaseNumberParser and override specific methods. Or new overridable methods can be created and Hindi config implements them differently from the default config.
Urudu will probably reuse these Hindi-specific changes, right?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -890,6 +920,18 @@ private Regex BuildTextNumberRegex()
                                     GetKeyRegex(this.Config.CardinalNumberMap.Keys);
             }
 
+            // For Hindi, there is a need for another NumberMap of the type double to handle values like 1.5.
+            // As this cannot be included into either Cardinal or Ordinal NumberMap as they are of the type long,
+            // HindiDecimalUnitsList (type double) takes care of these entries and it needs to be added to the singleIntFrac
+            // for the extractor to extract them
+            if (this.Config.CultureInfo.Name == "hi-IN")
```
tellarin: Language-specific initialization is acceptable. But it would be preferable if this could be a method in config.
As Italian already added this workaround, no need to change Hindi only at this stage though.

.NET/Microsoft.Recognizers.Text.Number/Parsers/INumberParserConfiguration.cs
```diff
@@ -20,6 +20,8 @@ public interface INumberParserConfiguration
 
         ImmutableDictionary<string, string> RelativeReferenceRelativeToMap { get; }
 
+        ImmutableDictionary<string, double> HindiDecimalUnitsList { get; }
```
tellarin: Same.

.NET/Microsoft.Recognizers.Text.Number/Parsers/INumberParserConfiguration.cs
```diff
@@ -210,5 +216,16 @@ public virtual string ResolveSpecificString(string numberStr)
 
             return string.Empty;
         }
+
+        // resolve decimal cases for Hindi
+        public virtual double ResolveUnitCompositeNumber(string numberStr)
```
tellarin: This and the dictionary can't be in the general interface.

LionbridgeCS:
@tellarin: feedback noted, we are working on handling it.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
LionbridgeCS: Noted. 
The majority of Indian languages e.g. Bengali, Urdu, Gujarati, Marathi, Punjabi and other Indo-Aryan and Dravidian languages have the same behavior. 

We will extend BaseNumberParser for Hindi for now.

LionbridgeCS:
@tellarin: PR ready for further review, we have some queries and we will update this according to your feedback.

.NET/Microsoft.Recognizers.Text.Number/Parsers/IIndianNumberParserConfiguration.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        Regex AdditionTermsRegex { get;  }
+
+        Regex FractionPrepositionInverseRegex { get; }
+
+        double ResolveUnitCompositeNumber(string numberStr);
+
+    }
+
+    public class BaseIndianNumberParserConfiguration : IIndianNumberParserConfiguration
```
tellarin: It's preferable to have BaseIndianNumberParserConfiguration in it's own file.

.NET/Microsoft.Recognizers.Text.Number/Parsers/IIndianNumberParserConfiguration.cs
```diff
@@ -0,0 +1,180 @@
...
+using System.Collections.Immutable;
+using System.Globalization;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Recognizers.Text.Number
+{
+    public interface IIndianNumberParserConfiguration : INumberParserConfiguration
+    {
+        ImmutableDictionary<string, double> DecimalUnitsMap { get; }
```
tellarin: Can you add comments on this interface to describe why these extra properties are needed over INumberParserConfiguration. This may be helpful for future developers.

.NET/Microsoft.Recognizers.Text.Number/Parsers/IIndianNumberParserConfiguration.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        public Regex NegativeNumberSignRegex { get; set; }
+
+        public bool IsCompoundNumberLanguage { get; set; }
+
+        public bool IsMultiDecimalSeparatorCulture { get; set; }
+
+        public Regex AdditionTermsRegex { get; set; }
+
+        public virtual long ResolveCompositeNumber(string numberStr)
```
tellarin: Can you document this resolution and give examples in the code?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
```diff
@@ -0,0 +1,838 @@
...
+            this.isCompoundNumberLanguage = config.IsCompoundNumberLanguage;
+
+            TextNumberRegex = BuildTextNumberRegex();
+        }
+
+        protected new IIndianNumberParserConfiguration Config { get; private set; }
+
+        protected new Regex TextNumberRegex { get; }
+
+        public override ParseResult Parse(ExtractResult extResult)
```
tellarin: Is the flow of the whole Parse substantially different from the BaseParser? Otherwise we should insert specific overridable points.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
```diff
@@ -0,0 +1,838 @@
...
+
+                pointPartRet += GetPointValue(matchStrs);
+            }
+
+            result.Value = intPartRet + pointPartRet;
+
+            return result;
+        }
+
+        protected ParseResult ParseFracLikeNumber(ExtractResult extResult)
```
tellarin: This should override FracLikeNumberParse (and one of the two should be renamed for consistency). This new name is better.

tellarin: I also see many similarities between this method and FracLikeNumberParse. What are the key differences?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
```diff
@@ -0,0 +1,838 @@
...
+            if (ret != null)
+            {
+                ret.Type = DetermineType(extResult);
+                ret.Text = ret.Text.ToLowerInvariant();
+            }
+
+            return ret;
+        }
+
+        protected ParseResult ParseTextNumber(ExtractResult extResult)
```
tellarin: Similar comments to ParseFracLikeNumber

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
```diff
@@ -0,0 +1,838 @@
...
+                else
+                {
+                    result.Value = (intValue + numerValue) / denominator;
+                }
+            }
+
+            return result;
+        }
+
+        protected ParseResult ParseDigitNumber(ExtractResult extResult)
```
tellarin: Similar comments for most method from here on. Some seem exactly the same as in the base class.

tellarin: Unfortunately I won't have time in the next couple days to do an exact comparison. Sorry if I missed something obvious.

tellarin:
@LionbridgeCS how big is the "extensive list" mentioned in the comments?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
LionbridgeCS: @tellarin: feedback noted, we will work on it. the main hindrance is that the said functions in base have a protection level that does not allow them to be overridden, hence why we included them as they are needed. The behavior is not not that much different from the base but only some peculiarities are addressed in Indian refactored. We are not straying too far from the base to stick to base behavior.  We will review these one by one.

LionbridgeCS:
@tellarin: for cardinal we have around 320 total entries for CardinalNumberMap and around 1390 for OrdinalNumberMap. If this is fine we can review the lists and add more.

tellarin:
> @tellarin: for cardinal we have around 320 total entries for CardinalNumberMap and around 1390 for OrdinalNumberMap. If this is fine we can review the lists and add more.

1390 seems a little much, but it should not be an issue. Is this expected to grow further? If so, can't we do some kind of normalization or semi-parsing before querying the dictionary?

LionbridgeCS:
> 
> 
> > @tellarin: for cardinal we have around 320 total entries for CardinalNumberMap and around 1390 for OrdinalNumberMap. If this is fine we can review the lists and add more.
> 
> 1390 seems a little much, but it should not be an issue. Is this expected to grow further? If so, can't we do some kind of normalization or semi-parsing before querying the dictionary?

@tellarin: we will review the lists and once we have the final number we will explore normalization or pre-parsing if possible (if not in this PR, it will be in a future PR)

LionbridgeCS:
@tellarin: Ready for further review. We further refactored the changes in BaseIndianNumberParser and commented the functions that we need for the specific handling. We changed the protection level of some of the functions in BaseNumberParser to avoid duplication and make them accessible in the BaseIndianNumberParser. We are trying to not stray from the base behavior to make further development easier. 

If agreed, this PR can be merged as is and we will include the lists for the better coverage (with a possible pre-parsing rule to reduce the number of entries) and few enhancements and bug fixes in a follow up PR. This will enable us to make PRs for the other completed subtypes.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseIndianNumberParser.cs
tellarin: Is there any special behaviour in this Parse?
Why not just override DigitNumberParse, FracLikeNumberParse, TextNumberParse, GetIntValue, and GetDigitalValue? Are there other tweeks besides replacing these methods that I didn't notice?

tellarin:
@LionbridgeCS, the latest push is much more readable. Thanks.
My only concern now is the apparent duplication of parse that maybe could be replaced only by overriding the specific methods.

LionbridgeCS:
> 
> 
> @LionbridgeCS, the latest push is much more readable. Thanks.
> My only concern now is the apparent duplication of parse that maybe could be replaced only by overriding the specific methods.

Nothing special in Parse, we will override the methods, and change the protection level for the original so that they are "overridable"

LionbridgeCS:
@tellarin: updated, pending the checks.

PR #1914, opened 2019-10-25
[.NET] Hindi Numbers Ordinal support

LionbridgeCS:
All ordinal cases pass. 
Added 24 case with code-mixing and Devanagari numerals. 

***************************************************************************************************
Work in progress as it includes base from #1913; and #1913 needs to be merged first.


LionbridgeCS:
@tellarin: PR is ready

PR #1915, opened 2019-10-26
[.NET] Turkish DateTime DateTimeModel support

LionbridgeCSII:
pass: 230
fail: 68

Skipped cases:
- Cases with weekdays and numbers, e.g. "Friday 31", "Friday the 31st". In English these cases are treated differently, but there is no such difference in Turkish (their translations are identical). So, currently, they are all treated as the second one (Friday the 31st). We are still considering if this is the best solution (it leads to some big difference in the results in certain cases).
- Cases that require adjustments in the regular expressions because some specific pattern was not yet included.
- Cases that require adjustments in the "Input" or "Result" fields because of wrong translation, ambiguity or different behaviour/meaning of some expressions in Turkish.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
@@ -282,33 +282,33 @@ namespace Microsoft.Recognizers.Text.DateTime
                 var equalMatch = Config.EqualRegex.MatchBegin(er.Text, trim: true);
 
                 // check also after match
-                if (this.Config.CheckBothBeforeAfter)
+                if (this.Config.CheckBothBeforeAfter && er.Data != null && er.Data.Equals(Constants.HAS_MOD))
                 {
-                    if (!beforeMatch.Success && !er.Type.EndsWith("datetimerange"))
+                    if (!beforeMatch.Success)
                     {
                         beforeMatch = Config.BeforeRegex.MatchEnd(er.Text, trim: true);
                         matchIsAfter = matchIsAfter || beforeMatch.Success;
                     }
 
-                    if (!afterMatch.Success && !er.Type.EndsWith("datetimerange"))
+                    if (!afterMatch.Success)
                     {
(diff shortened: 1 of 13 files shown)


LionbridgeCSII:
pass: 230
fail: 68

tellarin:
@LionbridgeCSII please rebase PR and remove 'Draft' mark for review.
Thanks.

LionbridgeCSII:
@tellarin, PR rebased and ready to review.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseMergedDateTimeParser.cs
```diff
@@ -282,33 +282,33 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refTime)
                 var equalMatch = Config.EqualRegex.MatchBegin(er.Text, trim: true);
 
                 // check also after match
-                if (this.Config.CheckBothBeforeAfter)
+                if (this.Config.CheckBothBeforeAfter && er.Data != null && er.Data.Equals(Constants.HAS_MOD))
```
tellarin: I'm not sure if this change is not causing side-effects. We'll need to run more experiments.
Is there a specific reason the previous code was not correct for Turkish?
The semantics of the new has_mod and the previous type checks don't match.

LionbridgeCSII: It turned out that the previous solution was specific to the Merged test cases and did not work for new test cases in DateTimeModel. The new approach is more general. 
The problem is that, in Turkish, these modifiers are usually found at the end of the match whether they are part of the already extracted subentity or added by MergedExtractor, but they need to be parsed separately in MergedParser only in the second case. 
Take, for example, "tomorrow before 10am" and "before tomorrow 10am", in Turkish, "before" would be at the end of the match in both cases.

tellarin: We're running some regression tests across languages to see if there's any side-effect. Thanks.

PR #1925, opened 2019-11-05
Missing month week format

HiltonGiesenow:
For an example like "second week of August", Luis returns me a datetime that the TimexResolver brings back as "XXXX-08-WXX-2". This format seems to be missing in this file. When I add in my fix, I now get:
    [0]: {[0, XXXX-08-WXX-2]}
    [1]: {[month, 08]}
    [2]: {[weekOfMonth, 2]}

which I think is correct?

Incidentally, the actual DateTimeSpec coming back from Luis is "DateTimeSpec(daterange, [XXXX-08-W02]", but it's sending back a Resolution with "XXXX-08-WXX-2" (in this example), so arguably that's where the issue is, and should be fixed there. This regex just helps to deal with that.


tellarin:
The correct timex returned for "second week of August" is "XXXX-08-W02". Can you describe how you get to "XXXX-08-WXX-2"? This last timex means a Tuesday in August with no specific date.

For example, "second monday of august" returns "XXXX-08-WXX-1-#2", where "XXXX-08-WXX-1" means it's a Monday and the "#2" indicates it's the second one.

HiltonGiesenow:
It does indeed, when I looked further, look like this shouldn't have happened - it seems like a bug in the resolver itself that gave me this timex. I've submitted a more recent PR that deals with all the issues I was having until then, including a unit test. Please ignore my first two PRs and review that most recent one.

Just fyi, I also found a problem parsing UTC times, and I've got a test and a fix, but it's dependent on my latest PR first, then I'll send a further PR so we can look at that on it's own properly.

tellarin:
Closing according to discussion.

PR #1926, opened 2019-11-05
Updating "date" to include time where it appears

HiltonGiesenow:
For an example like "every tuesday at 9:15:27", I'm getting back the following DateTimeSpec from Luis: XXXX-WXX-2T09:15:27 . There are no regex's that match, so I've added what I've encountered so far.


HiltonGiesenow:
I thought about that, but the problem is there are a bunch for date and a bunch for time. Combining would result in a set of duplicates of everything, and each of which would need to be maintained, bug fixed, etc.

PR #1927, opened 2019-11-05
Adding "datetime"

HiltonGiesenow:
This relates to my previous PR (#1926) - this is possibly a better way to resolve it. Here's the context from that PR:

When I send something like the following to a Luis endpoint "every tuesday at 9:15:27", I'm getting back the following DateTimeSpec: XXXX-WXX-2T09:15:27 . This class can't deal with both "date" and "time" together, so it just returns a piece of them - my suggested alternative is to handle all scenarios - keep doing "date" and "time" as is, but if the user (developer) wants both, (s)he can request "datetime" and it will run both searches and return the combined results. Using my suggested fix, the previous Timex returns now:

[0]: {[0, T09:15:27]}
[1]: {[dayOfWeek, 2]}
[2]: {[hour, 09]}
[3]: {[minute, 15]}
[4]: {[second, 27]}

I've added the option for "datetime", and removed the "^" and "$" start and end markers appropriately. This could be changed (removing "^" and "$") to only doing it if "datetime" is specified - perhaps it depends on whether there's a risk not having it in every run?

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
@@ -1,11 +1,12 @@
-using System;
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 using System.Collections.Generic;
-using System.Text;
 using System.Text.RegularExpressions;
 
-namespace ChitChattr.BotFramework.Recognizers
+namespace Microsoft.Recognizers.Text.DataTypes.TimexExpression
 {
-    public static class TimexRegex2
+    public static class TimexRegex
     {
         const string DateTimeCollectionName = "datetime";
         const string DateCollectionName = "date";
--- a/.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
@@ -1,49 +1,54 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
+using System;
 using System.Collections.Generic;
+using System.Text;
 using System.Text.RegularExpressions;
 
-namespace Microsoft.Recognizers.Text.DataTypes.TimexExpression
+namespace ChitChattr.BotFramework.Recognizers
 {
-    public static class TimexRegex
+    public static class TimexRegex2
     {
+        const string DateTimeCollectionName = "datetime";
+        const string DateCollectionName = "date";
+        const string TimeCollectionName = "time";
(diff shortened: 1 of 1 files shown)


.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
```diff
@@ -1,49 +1,54 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License.
-
```
tellarin: Please don't strip copyright notices.

.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
```diff
@@ -1,49 +1,54 @@
...
+using System;
 using System.Collections.Generic;
+using System.Text;
 using System.Text.RegularExpressions;
 
-namespace Microsoft.Recognizers.Text.DataTypes.TimexExpression
+namespace ChitChattr.BotFramework.Recognizers
```
tellarin: Namespaces and class names should not be changed.

.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
HiltonGiesenow: ah sorry, copy-pasting late at night - this was just a working version from my own project

.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexRegex.cs
HiltonGiesenow: same as above: copy-pasting late at night - this was just a working version from my own project, of course this should stay as it was

HiltonGiesenow:
In terms of context, this relates to my previous PR (https://github.com/microsoft/Recognizers-Text/pull/1926) - this is possibly a better way to resolve it. Here's the context from that PR:

When I send something like the following to a Luis endpoint "every tuesday at 9:15:27", I'm getting back the following DateTimeSpec: XXXX-WXX-2T09:15:27 . This class can't deal with both "date" and "time" together, so it just returns a piece of them - my suggested alternative is to handle all scenarios - keep doing "date" and "time" as is, but if the user (developer) wants both, (s)he can request "datetime" and it will run both searches and return the combined results. Using my suggested fix, the previous Timex returns now:

    [0]: {[0, T09:15:27]}
    [1]: {[dayOfWeek, 2]}
    [2]: {[hour, 09]}
    [3]: {[minute, 15]}
    [4]: {[second, 27]}

HiltonGiesenow:
Files fixed and comments updated - hope that makes more sense?

tellarin:
Closing due to #1928

PR #1930, opened 2019-11-07
[.NET | EN Currency] Fix currency cases with potential entity overlaps

Grey0202:
Fix cases like "1 $ 10" "$ 20,000.00 $ 37,305.97"


tellarin:
I'll close this PR as the modified fix will probably look very different.
Please add all the new specs in the new PR, ok? Also, if the result from a spec is not correct, it it still worth it to add it as NotSupproted and add a Comment attribute describing what's incorrect.
Thanks.

PR #1941, opened 2019-11-13
[.NET | EN Currency] Fix currency cases with potential entity overlaps

Grey0202:
- Fix cases like "1 $ 10 $" "$ 20,000.00 $ 37,305.97"
- Add a select method for curreny unit

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
@@ -100,6 +100,7 @@ namespace Microsoft.Recognizers.Text.NumberWithUnit
                     var maxFindPref = Math.Min(maxPrefixMatchLen, number.Start.Value);
                     var maxFindSuff = sourceLen - start - length;
 
+                    var closeMatch = false;
                     if (maxFindPref != 0)
                     {
                         // Scan from left to right, find the longest match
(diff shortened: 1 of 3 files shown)


.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -161,7 +190,7 @@ public List<ExtractResult> Extract(string source)
                                 Type = this.config.ExtractType,
                             };
 
-                            if (prefixUnit != null)
+                            if (prefixUnit != null && !this.config.ExtractType.Contains("currency"))
```
tellarin: Please move this harcoded string to Constants and declare a method to check the extractor type so you don't need to repeat this line everywhere. 
Even better would be if there was another way to represent the type of the extractor in the config in an easier way to use.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -128,6 +129,34 @@ public List<ExtractResult> Extract(string source)
                     }
 
                     mappingPrefix.TryGetValue(start, out PrefixUnitResult prefixUnit);
+
+                    // For currency unit, add the prefixUnit with number here.
+                    if (prefixUnit != null && !prefixMatched && this.config.ExtractType.Contains("currency"))
```
tellarin: What do you mean add it "here"? Can you give some example?
Any why would this code only run for Currency?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -427,5 +462,68 @@ private List<ExtractResult> FilterAmbiguity(List<ExtractResult> extractResults,
 
             return extractResults;
         }
+
+        private List<ExtractResult> SelectCandidate(string source, List<ExtractResult> extractResults, List<bool> unitIsPrefix)
```
tellarin: What's the performance impact of this new method?

Specs/NumberWithUnit/English/CurrencyModel.json
```diff
@@ -1924,5 +1924,300 @@
...
+          "isoCurrency": "EUR",
+          "unit": "Euro",
+          "value": "500"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "1 $10 dollars bill",
+    "Comment": "This is to show the current behaviour.",
```
tellarin: This duplicated case breaks the build.

Specs/NumberWithUnit/English/CurrencyModel.json
```diff
@@ -1924,5 +1924,300 @@
...
+          "unit": "Dollar",
+          "value": "10"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "1 $10 dollars bill",
+    "Comment": "This is to show the ideally behaviour.",
+    "NotSupported": "dotnet, javascript, python, java",
```
tellarin: Isn't this PR supposed to correct this?

Specs/NumberWithUnit/English/CurrencyModel.json
tellarin: Please remove as this is not the desired behaviour anyways. In the comment for the correct test, you can show what reason for it being NotSupproted is.

Specs/NumberWithUnit/English/CurrencyModel.json
```diff
@@ -1924,5 +1924,300 @@
...
+        "Resolution": {
+          "unit": "Dollar",
+          "value": "10"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "1 $10 dollars bill",
+    "Comment": "This is to show the ideally behaviour.",
```
tellarin: Why would this one fail? My understanding was that "1 $ 10 dollars bill" below is the ambiguous case.

Specs/NumberWithUnit/English/CurrencyModel.json
```diff
@@ -1924,5 +1924,300 @@
...
+        "Resolution": {
+          "unit": "Dollar",
+          "value": "10"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "1 $ 10 dollars bill",
+    "Comment": "This is to show the ideally behaviour.",
```
tellarin: Here you can  describe the incorrect current behaviour and why this is currently NotSupported.

PR #1947, opened 2019-11-22
[.NET] Turkish Number NumberRangeModel support

LionbridgeCSII:
59 cases passing
2 cases failing

also included fix for one of the failing cases in NumberModel: "bir buçuk" (one and a half).

Failing cases:
- "x, 10'dan büyük ve 20'den küçüktür. y, 50'den fazla değil ve 20'den az değildir." (x is larger than 10 and smaller than 20. y is no more than 50 and no less than 20.)
- "Onun skoru 5000'e eşit veya 5000'den az" (His score is equal to 5000 or less than 5000)

The problem is due to the fact that, in Turkish, the modifiers follow the number entity, e.g. "5000'den az" (less than 5000). For this reason, the method ValidateMatchAndGetStartAndLength in BaseNumberRangeExtractor returns as start index the first index of the number entity (line 122: start = source.IndexOf(numberStr) + extractNum.Start ?? 0;) and since the number entities are equal, they both get the same start index. We have not modified the method because the solutions we tried break other test cases (in Turkish and other languages).


tellarin:
Re-triggering build.

Issue #1940, opened 2019-11-12
[TimexExpression] Timex.to_natural_language() does not return correct relative value


tracyboehrer:
**Describe the bug**
Given:

```
reference_date = datetime.datetime.now()
timex_date = Timex.from_date(datetime.datetime.now())
print(f"{timex_date.timex_value()} : {timex_date.to_natural_language(reference_date)}")
```

Outputs:
```
2019-11-12 : 12nd November 2019
```

**Expected behavior**
The .Net version returns: 

```
2019-11-12 : today
```

**Platform (please complete the following information):**
 - Platform: Python
 - Version of package 1.0.2a2

PR #1949, opened 2019-11-22
[TimexExpression] Modify TimexDateHelpers.date_part_equal() to fix Timex.to_natural_language() 

Aliandi:
Issue #1940
### Changes Made
- Ignore time when comparing dates
- Add `now` when recognizing present types
### Testing

In order to test this issue, we created a small script using the repro steps from the original issue.

imagen


Aliandi:
@tellarin We updated the PR and added a test for this.

PR #1950, opened 2019-11-24
[EN DateTimeV2] Add support for cases like "Every other Friday"

CcKeima:
Support for cases like "Every other Friday".
**The problem:** "Every other Friday" was recognized as "Friday".
**Changes as below:**

- Match "WeekDay" after "every other"

- Add "every other" as SetEachRegex


CcKeima:
It will be same as "Every Friday" like this:
图片1

tellarin:
@CcKeima, the build is failing in Python. Such new test cases don't pass. You can disable the test cases in Python only by marking them "NotSupported" : "python", in case you can't fix the issue.

CcKeima:
I think it because DateTimeModel's result is not allow upper case letters.

PR #1971, opened 2019-12-03
[EN DateTimeV2] Revise ImplicitDate

Grey0202:
.NET only for now. Pending performance test.

- Add test
- Support cases like "1st week of Oct. on Friday"

Draft due to performance drop.


tellarin:
This is currently blocked as perf was it, right @Grey0202? @LionbridgeCSII, could you take a look and try other approach to solve the source issue?

LionbridgeCSII:
@tellarin, a simpler approach would be to include the new pattern directly into WeekDayOfMonthRegex, in this way there is no need to modify the base code.
I could not see any performance issue using this method, but maybe I am missing something, do you have any guidelines on how analyse performance?

tellarin:
Closing as it was replaced by #2120.

PR #1977, opened 2019-12-04
[.NET] Hindi DateTime Date support

LionbridgeCS:
Support for Hindi Date. 
Extractor (100 Pass - 0 Fail)
Parser (85 Pass - 2 Skipped)
********************************************************************************************************
Notes: 
- Includes DateTime skeleton and base for Time subtype along with other localizations. 
- All test cases are localized from English.
- Change in the IndianBaseNumberParser to account for Ordinals before cardinal (तीसरे=third but तीस=thirty)
- Change in the BaseDateParser to check if the extraction is empty or a relative ordinal (e.g. "next", "previous").  without this extraction could result in a relative ordinal and then try to convert it to int; this is relevant in Hindi because cases like "the 20th of next month" have the structure "next month the 20th", so 'next' is extracted before 20th. 
- Skipps in the DateParser test are related to "yesterday|tomorrow" which in Hindi are the same and are disambiguated only by context. कल and परसों are ambiguous as they refer to one day later and two days later respectively both forward and backward in time; this has to be disambiguated based on the context and other words indicating grammatical tense/aspect of the sentence.


tellarin:
> Skips in the DateParser test are related to "yesterday|tomorrow" which in Hindi are the same and are disambiguated only by context. कल and परसों are ambiguous as they refer to one day later and two days later respectively both forward and backward in time; this has to be disambiguated based on the context and other words indicating grammatical tense/aspect of the sentence.

Could any frequent usage cases of these be disambiguated by common terms immediately next to them?

.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Hindi.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        [NetCoreTestDataSource]
+        [TestMethod]
+        public void DateExtractor(TestModel testSpec)
+        {
+            ExtractorInitialize(Extractors);
+            TestDateTimeExtractor(testSpec);
+        }
+
+        ////[NetCoreTestDataSource]
```
tellarin: All these test calls can be uncommented, as long as the specs are properly marked as NotSupported during dev.

.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDateTimeAltExtractorConfiguration.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions.Hindi;
+
+namespace Microsoft.Recognizers.Text.DateTime.Hindi
+{
+    public class HindiDateTimeAltExtractorConfiguration : BaseDateTimeOptionsConfiguration, IDateTimeAltExtractorConfiguration
```
tellarin: No need to implement/test cases for this specific type until all basic DateTime items are complete.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Hindi.cs
LionbridgeCS: we will uncomment with the PRs for each of the subtypes (that include the specs)

tellarin: @LionbridgeCS please uncomment all at once and add the spec files even if empty or with only one disabled case per file. 
The recommendation is to always add either all new language specs marked as NotSupported, or to add the skeleton of specs and test code for the whole type.
There could be a simple PR just for this. The main goal here is to show that all skeleton code runs and executes it's tests and there is no infra-structure failure.

.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDateTimeAltExtractorConfiguration.cs
LionbridgeCS: > 
> 
> > Skips in the DateParser test are related to "yesterday|tomorrow" which in Hindi are the same and are disambiguated only by context. कल and परसों are ambiguous as they refer to one day later and two days later respectively both forward and backward in time; this has to be disambiguated based on the context and other words indicating grammatical tense/aspect of the sentence.
> 
> Could any frequent usage cases of these be disambiguated by common terms immediately next to them?

There are such terms, we will add a couple of cases with such.

LionbridgeCS:
> 
> 
> > Skips in the DateParser test are related to "yesterday|tomorrow" which in Hindi are the same and are disambiguated only by context. कल and परसों are ambiguous as they refer to one day later and two days later respectively both forward and backward in time; this has to be disambiguated based on the context and other words indicating grammatical tense/aspect of the sentence.
> 
> Could any frequent usage cases of these be disambiguated by common terms immediately next to them?

There are terms that can indicate, we will add a couple of cases with such.

.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Hindi.cs
LionbridgeCS: @tellarin: noted. we will make a simple PR with this.

PR #1979, opened 2019-12-05
[EN DateTimeV2] Add support for "lunch time" "at lunch"

Grey0202:
- Add support for cases like "lunch time" "dinner time" "at dinner"
- Add test

### DIFF
--- a/Patterns/English/English-DateTime.yaml
+++ b/Patterns/English/English-DateTime.yaml
@@ -415,7 +415,7 @@ PrepositionRegex: !simpleRegex
 LaterEarlyRegex: !simpleRegex
   def: ((?<early>early(\s+|-))|(?<late>late(r?\s+|-)))
 MealTimeRegex: !simpleRegex
-  def: \b(?<mealTime>breakfast|brunch|lunch(time)?|dinner(time)?|supper)\b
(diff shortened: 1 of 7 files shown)


Patterns/English/English-DateTime.yaml
```diff
@@ -415,7 +415,7 @@ PrepositionRegex: !simpleRegex
 LaterEarlyRegex: !simpleRegex
   def: ((?<early>early(\s+|-))|(?<late>late(r?\s+|-)))
 MealTimeRegex: !simpleRegex
-  def: \b(?<mealTime>breakfast|brunch|lunch(time)?|dinner(time)?|supper)\b
+  def: \b(at\s+)?(?<mealTime>breakfast|brunch|lunch(\s*time)?|dinner(\s*time)?|supper)\b
```
tellarin: What is LunchRegex now used for?
And should the regex in ambiguity filter be removed?

Grey0202: LunchRegex is to judge if the time is am or pm. Such as "11:00" could be both am and pm, but "11:00 lunchtime" is am time.
I revised the ambiguity filter, I think we'd better not to remove it as we still need to filter pure  "lunch".

PR #1980, opened 2019-12-05
Fix to handle UTC on timex resolving (e.g. XXXX-WXX-2T15)

HiltonGiesenow:
For a scenario like "tuesday at 3:00 pm" the timex was fine, but even though the resolver accepts a date it doesn't do any UTC check. Now, if a UTC time is passed in, it will take the resolved time, which is effectively in "local" time, and, if the baseline datetime passed in is in UTC, it will convert it to UTC.


HiltonGiesenow:
.

PR #1983, opened 2019-12-05
[.NET] Hindi DateTime Time support

LionbridgeCSII:
Support for Hindi Time.
Extractor (77 Pass - 0 Fail)
Parser (76 Pass - 1 Skipped)

Notes:
- The skipped case corresponds to "I'll go back three mins to five thirty" that in Hindi has a different structure (half past five three mins to).
- The modification in BaseTimeParser is needed to parse Devanagari numbers (since they are treated as standard numbers by the regexes but not recognized by int.Parse())


Specs/DateTime/Hindi/DateTimeModel.json
```diff
@@ -0,0 +1,3 @@
+[
```
tellarin: I recommend populating the DateTimeModel specs whenever possible, as they provide visibility on overall entity coverage. But this doesn't block PRs until the final one.

Specs/DateTime/Hindi/TimePeriodParser.json
```diff
@@ -0,0 +1,1327 @@
+[
+  {
+    "Input": "午後5時から6時まで不在にします。",
```
tellarin: Can you remove these Japanese specs?

tellarin: BTW, why was Japanese used as source? English should be considered the master version for localization.

Specs/DateTime/Hindi/TimeExtractor.json
```diff
@@ -0,0 +1,918 @@
+[
+  {
+    "Input": "मैं 7 बजे तक वापस आ जाउंगा",
+        "NotSupportedByDesign": "javascript,python,java",
```
tellarin: Nit: Please fix indentation in json file.

Specs/DateTime/Hindi/TimePeriodParser.json
LionbridgeCSII: sorry, did not notice this. Not sure why the initial spec was Japanese, will check

Specs/DateTime/Hindi/TimeExtractor.json
LionbridgeCSII: fixed

PR #1985, opened 2019-12-06
[.NET] Hindi DateTime Holiday support

LionbridgeCSII:
All test cases pass


tellarin:
Thanks! We'll continue the review after #1983 merges.

tellarin:
@LionbridgeCSII, can you resolve the conflicts first to we can review/merge? Thanks.

LionbridgeCSII:
@tellarin, conflicts solved

.NET/Microsoft.Recognizers.Definitions.Common/Hindi/DateTimeDefinitions.cs
```diff
@@ -994,30 +992,233 @@ public static class DateTimeDefinitions
...
             { @"allsoulsday", new string[] { @"allsoulsday" } },
             { @"guyfawkesday", new string[] { @"guyfawkesday" } },
             { @"veteransday", new string[] { @"veteransday" } },
             { @"christmaseve", new string[] { @"christmaseve" } },
-            { @"newyeareve", new string[] { @"newyearseve", @"newyeareve" } }
+            { @"newyeareve", new string[] { @"newyearseve", @"newyeareve", @"नएसालकीशाम", @"नववर्षकीपूर्वसंध्या", @"न्यूइयरईव" } }
+        };
+      public static readonly Dictionary<int, IEnumerable<int>> HoliDiwaliDates = new Dictionary<int, IEnumerable<int>>
+        {
+            { 1900, new int[] { 3, 16, 10, 23 } },
```
tellarin: Isn't there a way to calculate these instead of adding them all here?
Not blocker.

tellarin:
> @tellarin, conflicts solved

We'll merge once the builds pass. Thanks!

.NET/Microsoft.Recognizers.Definitions.Common/Hindi/DateTimeDefinitions.cs
LionbridgeCSII: we tried to come up with a formula based on the average duration of the lunar month, but it is difficult to get a reliable result. Using tabulated dates seems a safer solution at the moment. We will update the method in case we find the solution to the problem

tellarin: If it helps, the codebase has some code to calculate lunar holidays. But I have no knowledge of how Diwali is calculated. And it can also vary in different regions, right?

LionbridgeCSII: ah interesting, will check the codebase! 
Diwali lasts a week and the start day is the same everywhere, but certain states celebrates maybe only one day.

PR #1987, opened 2019-12-06
[* DateTimeV2] Fixes ranges incorrectly interpreted with dash-formatted dates

### LINKED ISSUE #1981: [EN DateTimeV2] Ranges incorrectly interpreted in format MM-DD-YYYY H:MMam/pm to/until MM-DD-YYYY H:MMam/pm
When you attempt to recognize in LUIS or in the Recognizer Nuget the following query (any datetime range with indicated format will repro)

12-04-2019 6:00am until 12-05-2019 5:00pm

You will not get a correct recognizer result

timex (2019-12-05T00,2019-12-05T04,PT4H)

type datetimerange

start 2019-12-05 00:00:00

end 2019-12-05 04:00:00

timex (2019-12-05T12,2019-12-05T16,PT4H)

type datetimerange

start 2019-12-05 12:00:00

end 2019-12-05 16:00:00

Or at least I cannot make sense of it. It doesn't even talk about the 4th anywhere I can tell.

It can recognize 12-04-2019 6:00am just fine by itself
It can recognize 12-05-2019 5:00pm just fine by itself

1. Use Recognizer Nuget in .NET CORE application
var result = Microsoft.Recognizers.Text.DateTime.DateTimeRecognizer.RecognizeDateTime(turnContext.Activity.Text, "en-us");

            foreach (var r in result)
            {
                foreach (var item in r.Resolution)
                {
                    if(item.Value is List<Dictionary<String,String>> dictionaryList)
                    {
                        foreach (var listItem in dictionaryList)
                        {
                            foreach (var kvp in listItem)
                            {
                                System.Console.WriteLine($"{kvp.Key} {kvp.Value}");
                            }
                        }
                    }
                }
            }

**Expected behavior**
It should provide a datetimerange that is accurate

timex (2019-12-04T06:00,2019-12-05T17:00,PT35H)

type datetimerange

start 2019-12-04 06:00:00

end 2019-12-05 17:00:00

This is possible if I change the format of the datetime from MM-DD-YYYY to YYYY-MM-DD

**Sample input/output**
Doesn't work: 12-04-2019 6:00 am until 12-05-2019 5:00pm
Does work: 2019-12-04 6:00am until 2019-12-05 5:00pm

 - Platform: .NET
 - Environment: Nuget Microsoft.Recognizers.Text.DateTime, LUIS
 - Version of package 1.1.4, not sure about what LUIS is using.

clintwar:
ParsePureNumberCases Incorrectly Detects Time Period When Time Period Is Part Of Hyphenated Date

When the regular expression is used to detect a time period it fails to ensure the time period detected was not actually a datetime that begins with either a month or a day.
12-04 for example looks like a time period, but the full text of it is actually 12-04-2019. 

Because of this the recognizer logic determines there is a time period and so then tries to see if it can find the date to which the time period belongs. This does find another date in the scenario outlined in 1981, and results in a completely incorrect recognizer result.

12-04-2019 6:00am until 12-05-2019 8:00pm is detected as

12:00pm until 4:00pm on the 5th of december 2019. It gets 12:00pm until 4:00pm from the 12-04 in the first date. It finds 12-05-2019 because it is the next date it finds in the query.

This code change ensures that the time period detected is not part of a date detected.

Fixes #1981.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
@@ -697,8 +697,6 @@ namespace Microsoft.Recognizers.Text.DateTime
                 // Just because we think we found a time period doesn't mean it is one, it could be the start of a hyphenated date
                 var hyphenDateMatch = this.Config.HyphenDateRegex.Match(trimmedText);
 
-                    // Regex.Match(trimmedText, "((?<year1>[0-9]{4})-?(?<month1>1[0-2]|0[1-9])-?(?<day1>3[01]|0[1-9]|[12][0-9]))|((?<month2>1[0-2]|0[1-9])-?(?<day2>3[01]|0[1-9]|[12][0-9])-?(?<year2>[0-9]{4}))|((?<day3>3[01]|0[1-9]|[12][0-9])-?(?<month3>1[0-2]|0[1-9])-?(?<year3>[0-9]{4}))");
-
                 if (hyphenDateMatch.Success && hyphenDateMatch.Index >= match.Index && (match.Index + match.Length) <= (hyphenDateMatch.Index + hyphenDateMatch.Length))
                 {
                     return ret;
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
@@ -693,6 +693,17 @@ namespace Microsoft.Recognizers.Text.DateTime
 
             if (match.Success && (match.Index == 0 || match.Index + match.Length == trimmedText.Length))
             {
+
+                // Just because we think we found a time period doesn't mean it is one, it could be the start of a hyphenated date
+                var hyphenDateMatch = this.Config.HyphenDateRegex.Match(trimmedText);
+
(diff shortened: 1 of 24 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateTimePeriodParser.cs
```diff
@@ -693,6 +693,15 @@ private DateTimeResolutionResult ParsePureNumberCases(string text, DateObject re
 
             if (match.Success && (match.Index == 0 || match.Index + match.Length == trimmedText.Length))
             {
+
+                // Just because we think we found a time period doesn't mean it is one, it could be the start of a hyphenated date
+                var hyphenDateMatch = this.Config.HyphenDateRegex.Match(trimmedText);
+
+                if (hyphenDateMatch.Success && hyphenDateMatch.Index >= match.Index && (match.Index + match.Length) <= (hyphenDateMatch.Index + hyphenDateMatch.Length))
```
tellarin: Thanks for proposing a fix, but we need to test the performance impact of this change.
Could you run some experiments?

tellarin: Could you also add some details in the PR description about the root cause you've identified.
It helps make sure the intent of the fix and as future documentation.

Grey0202: > Thanks for proposing a fix, but we need to test the performance impact of this change.
> Could you run some experiments?

This PR now causes about 0.8% performance drop according to my test. :(

clintwar: I tested the DataDrivenTests in the project and

With Fix: Run finished: 20231 tests run (0:02:34.0273429) ==========
Without Fix: Run finished: 20229 tests run (0:02:52.0896418) ==========

Seem they ran a little bit faster with the fix.

Maybe I am not validating correctly?

clintwar: The test runs vary from time to time but so far they seem to run faster with the fix.
Run 2
With Fix: Run finished: 20231 tests run (0:02:25.6527977) =========
Without Fix: Run finished: 20229 tests run (0:02:31.4956609) ==========

Grey0202: I re-checked it in the current master, it seems no performance drop now.

PR #1988, opened 2019-12-07
[EN NumberWithUnit] Revise fractional Currency

Grey0202:
- Fix fraction bigger than max fraction value
- Todo: filter invalid fraction.


.NET/Microsoft.Recognizers.Definitions.Common/BaseCurrency.cs
```diff
@@ -268,5 +268,16 @@ public static class BaseCurrency
             { @"Hào", 10 },
             { @"Ngwee", 100 }
         };
+      public static readonly Dictionary<string, long> SpecialCurrencyFractionMax = new Dictionary<string, long>
+        {
+            { @"JOD", 1000 },
```
tellarin: You can remove the CYP from this list. For all purposes it can be considered as 100 fractional, as that was it's base when abolished.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/BaseCurrencyParser.cs
```diff
@@ -175,9 +175,16 @@ private ParseResult MergeCompoundUnit(ExtractResult compoundResult)
                     // Match pure number as fraction unit.
                     if (extractResult.Type.Equals(Constants.SYS_NUM, StringComparison.Ordinal))
                     {
-                        numberValue += (double)parseResult.Value * (1.0 / 100);
-                        result.ResolutionStr += ' ' + parseResult.ResolutionStr;
-                        result.Length = parseResult.Start + parseResult.Length - result.Start;
+                        Config.SpecialCurrencyFractionMaxMap.TryGetValue(mainUnitIsoCode, out var fractionMaxValue);
+
+                        fractionMaxValue = fractionMaxValue == 0 ? 100 : fractionMaxValue;
```
tellarin: Why does the PR description mention "Todo: filter invalid fraction."? 
Doesn't this change address the issue?

tellarin: Can you declare a constant in this class for this usage of 100? Something like DEFAULT_FRACTIONAL_SUBUNIT?

.NET/Microsoft.Recognizers.Definitions.Common/BaseCurrency.cs
```diff
@@ -268,5 +268,16 @@ public static class BaseCurrency
             { @"Hào", 10 },
             { @"Ngwee", 100 }
         };
+      public static readonly Dictionary<string, long> SpecialCurrencyFractionMax = new Dictionary<string, long>
```
tellarin: Please rename to NonStandardFractionalSubunits

Specs/NumberWithUnit/English/CurrencyModel.json
```diff
@@ -2201,5 +2201,38 @@
...
+        "TypeName": "currency",
+        "Resolution": {
+          "unit": "Turkish lira",
+          "value": "10"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "deposits: 10 turkish lira 1111-22223333-4444",
```
tellarin: Can you add this case in Turkish?
"mustafa ustaya 10 bin lira 1111-22223333-4444 nolu hesaba havale yap"

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/BaseMergedUnitExtractor.cs
```diff
@@ -155,6 +156,15 @@ private void MergePureNumber(string source, List<ExtractResult> ers)
                     continue;
                 }
 
+                if (source.Length > numErs[i].Start + numErs[i].Length)
```
tellarin: Can you add a comment here explaining the purpose of this check?
Is there any perf impact?

Grey0202: I just tested it and there is no performance drop.

PR #1989, opened 2019-12-09
[EN DateTime] Make Q1/Q2 result generalized (.NET)

### LINKED ISSUE #1984: [EN DateTimeV2] Timex for "Q1/Q2" are losing generality
**Describe the bug**
The TimeX of "Q1" is "(2019-01-01,2019-04-01,P3M)" and "(2019-01-01,2019-04-01,P3M)"
While the TimeX of "Apri" is "XXXX-04"

**Expected behavior**
The TimeX is similar to "in April and Aug", which starts with "XXXX-".
(Or is there other way that we know the query does not indicate the year part?)

**Sample input/output**
Query: "in Q1 and Q2"
[
  {
    "Text": "q1",
    "Start": 3,
    "End": 4,
    "TypeName": "datetimeV2.daterange",
    "Resolution": {
      "values": [
        {
          "timex": "(2019-01-01,2019-04-01,P3M)",
          "type": "daterange",
          "start": "2019-01-01",
          "end": "2019-04-01"
        },
        {
          "timex": "(2019-01-01,2019-04-01,P3M)",
          "type": "daterange",
          "start": "2020-01-01",
          "end": "2020-04-01"
        }
      ]
    }
  },
  {
    "Text": "q2",
    "Start": 10,
    "End": 11,
    "TypeName": "datetimeV2.daterange",
    "Resolution": {
      "values": [
        {
          "timex": "(2019-04-01,2019-07-01,P3M)",
          "type": "daterange",
          "start": "2019-04-01",
          "end": "2019-07-01"
        },
        {
          "timex": "(2019-04-01,2019-07-01,P3M)",
          "type": "daterange",
          "start": "2020-04-01",
          "end": "2020-07-01"
        }
      ]
    }
  }
]

Grey0202:
This fixes #1984 (.NET only)
- Make quarter Timex result generalized when no year contains.
- Add tests

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
@@ -1873,13 +1873,15 @@ namespace Microsoft.Recognizers.Text.DateTime
                 {
                     ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDate, endDate);
                 }
+
+                ret.Timex = $"({DateTimeFormatUtil.LuisDate(-1, beginDate.Month, 1)},{DateTimeFormatUtil.LuisDate(-1, endDate.Month, 1)},P3M)";
             }
             else
             {
                 ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDate, endDate);
+                ret.Timex = $"({DateTimeFormatUtil.LuisDate(beginDate)},{DateTimeFormatUtil.LuisDate(endDate)},P3M)";
             }
 
-            ret.Timex = $"({DateTimeFormatUtil.LuisDate(beginDate)},{DateTimeFormatUtil.LuisDate(endDate)},P3M)";
             ret.Success = true;
 
             return ret;
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -14771,5 +14771,77 @@
         }
       }
     ]
+  },
+  {
+    "Input": "How about 2019 q1",
+    "Context": {
+      "ReferenceDateTime": "2019-11-07T00:00:00"
+    },
+    "Results": [
+      {
+        "Text": "2019 q1",
+        "Start": 10,
+        "End": 16,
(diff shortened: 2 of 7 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1873,13 +1873,15 @@ private DateTimeResolutionResult ParseQuarter(string text, DateObject referenceD
                 {
                     ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDate, endDate);
                 }
+
+                ret.Timex = $"({DateTimeFormatUtil.LuisDate(-1, beginDate.Month, 1)},{DateTimeFormatUtil.LuisDate(-1, endDate.Month, 1)},P3M)";
```
tellarin: Please declare a constant to represent -1

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1873,13 +1873,15 @@ private DateTimeResolutionResult ParseQuarter(string text, DateObject referenceD
...
             }
             else
             {
                 ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDate, endDate);
+                ret.Timex = $"({DateTimeFormatUtil.LuisDate(beginDate)},{DateTimeFormatUtil.LuisDate(endDate)},P3M)";
```
tellarin: Also, these string concatenations to create a timex should actually be a method that receives the dates (either as data object or as year/month/day) and return the formed timex.
Perhaps re-using/adapting GenerateDatePeriodTimex in TimexUtility.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -14719,5 +14719,57 @@
         }
       }
     ]
+  },
+  {
+    "Input": "q1 and q2 could be any year",
```
tellarin: Can you just double check that we have cases in these four formats with the year specified and they currently pass? If some is missing, please add a spec, ok?
2019 q1
q3 2019
2019-Q1
q1-2019
Please also add a spec for "2019 q3 and 2020 q1"

PR #1991, opened 2019-12-10
[.NET] Hindi DateTime TimePeriod support

LionbridgeCSII:
All test cases passing.

In DateTimeModel: 4 cases passing, 80 failing.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
@@ -717,7 +717,7 @@ namespace Microsoft.Recognizers.Text.DateTime
             pr1 = this.config.TimeParser.Parse(ers[0], referenceTime);
             pr2 = this.config.TimeParser.Parse(ers[1], referenceTime);
 
-            if (pr1.Value == null || pr2.Value == null)
+            if (pr1.Value == null || pr2.Value == null || pr1.Text == pr2.Text)
             {
                 return ret;
             }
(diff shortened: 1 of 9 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
```diff
@@ -717,7 +717,7 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
             pr1 = this.config.TimeParser.Parse(ers[0], referenceTime);
             pr2 = this.config.TimeParser.Parse(ers[1], referenceTime);
 
-            if (pr1.Value == null || pr2.Value == null)
+            if (pr1.Value == null || pr2.Value == null || pr1.Text == pr2.Text)
```
tellarin: Which situation falls into this new clause?

LionbridgeCS:
@tellarin: [SideNote] from this PR onward we will include the pass.fail from the DateTime Model with each of the subtypes. Would you kindly confirm that this ok and that the model enabling is only done when all is passing; as we can also enable the model already and with each completed subtype enable the passing test cases.

tellarin:
@LionbridgeCS, this is OK. Actual model enabling should only happen when all is passing.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseTimePeriodParser.cs
LionbridgeCSII: in Hindi, there are expressions like "morning-morning" meaning "early-morning" that are otherwise parsed as "from morning to morning"

LionbridgeCSII: the change does not seem to affect other languages.
Is it better that we revert it and set the corresponding case as not supported?

tellarin: I'm Ok with this, just add a comment in the code explaining what it addresses.

tellarin: I'm Ok with this, just add a comment in the code explaining what it addresses.

tellarin: Please add the comment in the next PR, ok?

PR #1992, opened 2019-12-11
[EN Datetime] Revise DatePeriod Parser

Grey0202:
- Avoid single years being extracted. 
- Revise Timex generate (.NET)

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
@@ -1263,7 +1263,7 @@ namespace Microsoft.Recognizers.Text.DateTime
                         DateObject.MinValue.SafeCreateFromValue(endYear, 1, 1).AddDays(-1) :
                         DateObject.MinValue.SafeCreateFromValue(endYear, 1, 1);
 
-                    ret.Timex = $"({DateTimeFormatUtil.LuisDate(beginDay)},{DateTimeFormatUtil.LuisDate(endDay)},P{endYear - beginYear}Y)";
+                    ret.Timex = TimexUtility.GenerateDatePeriodTimex(beginDay, endDay, DatePeriodTimexType.ByYear);
                     ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDay, endDay);
                     ret.Success = true;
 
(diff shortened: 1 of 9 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1873,15 +1873,13 @@ private DateTimeResolutionResult ParseQuarter(string text, DateObject referenceD
...
-
-                ret.Timex = $"({DateTimeFormatUtil.LuisDate(-1, beginDate.Month, 1)},{DateTimeFormatUtil.LuisDate(-1, endDate.Month, 1)},P3M)";
             }
             else
             {
                 ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDate, endDate);
-                ret.Timex = $"({DateTimeFormatUtil.LuisDate(beginDate)},{DateTimeFormatUtil.LuisDate(endDate)},P3M)";
             }
 
+            ret.Timex = TimexUtility.GenerateDatePeriodTimex(beginDate, endDate, DatePeriodTimexType.ByMonth, noSpecificYear);
```
tellarin: The last parameter should not be a bool. It needs to be able to represent any part of the missing context, not only year.
Maybe a flags enum?

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -85,6 +85,53 @@ public static string GenerateDatePeriodTimex(DateObject begin, DateObject end, D
             return $"({DateTimeFormatUtil.LuisDate(begin, alternativeBegin)},{DateTimeFormatUtil.LuisDate(end, alternativeEnd)},{datePeriodTimex})";
         }
 
+        public static string GenerateDatePeriodTimex(DateObject begin, DateObject end, DatePeriodTimexType timexType, bool nonspecificYear, bool nonspecificMonth = false, bool nonspecificDay = false)
```
tellarin: Why this new method and not re-use a previous GenerateDatePeriodTimex?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
tellarin: What about UnspecificDateTimeTerm?

PR #1993, opened 2019-12-12
[.NET] Hindi DateTime Duration support

LionbridgeCSII:
All test cases pass.
DateTimeModel: 34 passing, 50 failing.

Comments:
The case "मैं साढ़े 3 सेकंड के लिए बाहर रहुंगा" (I'll leave for 3 and a half seconds)
required a change to the base Extractor and Parser analogous to the ones implemented for Turkish because the modifier "and a half" precedes the number. The case has been added to DurationParser.json (originally, it was only in DurationExtractor.json)

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Hindi
 
         public IImmutableDictionary<string, long> UnitValueMap { get; }
 
-        bool IDurationExtractorConfiguration.CheckBothBeforeAfter => DateTimeDefinitions.CheckBothBeforeAfter;
+        bool IDurationExtractorConfiguration.CheckBothBeforeAfter => true;
 
         Regex IDurationExtractorConfiguration.FollowedUnit => DurationFollowedUnit;
 
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Hindi
 
         public IImmutableDictionary<string, long> UnitValueMap { get; }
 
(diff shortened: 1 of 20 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
```diff
@@ -70,7 +70,7 @@ public HindiDurationExtractorConfiguration(IDateTimeOptionsConfiguration config)
 
         public IImmutableDictionary<string, long> UnitValueMap { get; }
 
-        bool IDurationExtractorConfiguration.CheckBothBeforeAfter => DateTimeDefinitions.CheckBothBeforeAfter;
+        bool IDurationExtractorConfiguration.CheckBothBeforeAfter => true;
```
tellarin: This should be read from the YAML file.

tellarin: If this flag only applies to Durations, not in general, please declare different variables in the YAML.

.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Parsers/HindiDurationParserConfiguration.cs
```diff
@@ -8,8 +11,8 @@ public class HindiDurationParserConfiguration : BaseDateTimeOptionsConfiguration
         public HindiDurationParserConfiguration(ICommonDateTimeParserConfiguration config)
            : base(config)
         {
-            CardinalExtractor = config.CardinalExtractor;
-            NumberParser = config.NumberParser;
+            CardinalExtractor = Number.Hindi.CardinalExtractor.GetInstance();
```
tellarin: Why not get these two from the parser config?

LionbridgeCSII: right, moved modifications to HindiCommonDateTimeParserConfiguration

.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
LionbridgeCSII: ok, modified to use the global flag in the yaml. If necessary, we will introduce subtype specific flags in future PRs (in previous tests, the use of the global flag broke other subtypes, but at the moment everything works)

PR #2005, opened 2019-12-22
[.NET] Hindi DateTime DateTime support

LionbridgeCSII:
All test cases pass.
DateTimeModel: 45|39

### DIFF
--- a/Specs/DateTime/Hindi/TimeExtractor.json
+++ b/Specs/DateTime/Hindi/TimeExtractor.json
@@ -324,23 +324,35 @@
     ]
   },
   {
-    "Input": "मैं दिन में 7 बजे वापस जाउंगा",
+    "Input": "मैं अपराहन में 7 बजे वापस जाउंगा",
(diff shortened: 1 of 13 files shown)


Specs/DateTime/Hindi/TimeExtractor.json
```diff
@@ -324,47 +324,47 @@
     ]
   },
   {
-    "Input": "मैं दिन में 7 बजे वापस जाउंगा",
+    "Input": "मैं अपराहन में 7 बजे वापस जाउंगा",
```
tellarin: What's the reason for these multiple sentence changes and previously supported types?
Whenever a PR includes such changes, can you add a comment in the PR description?

LionbridgeCSII: sorry, forgot to add comment. 
These cases were changed because the original translation was not accurate. In this specific case, it meant "I'll be back at 7 in the day" instead of "I'll be back at 7 pm"

PR #2006, opened 2019-12-22
[.NET] Hindi DateTime DateTimePeriod support

LionbridgeCSII:
Extractor: 52 passing, 16 failing;
Parser: 35 passing, 33 failing.
DateTimeModel: 46|38

### DIFF
--- a/Specs/DateTime/Hindi/DateTimePeriodParser.json
+++ b/Specs/DateTime/Hindi/DateTimePeriodParser.json
@@ -4,7 +4,6 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T16:12:00"
     },
-    "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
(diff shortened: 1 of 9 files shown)


Specs/DateTime/Hindi/TimeExtractor.json
```diff
@@ -324,47 +324,47 @@
     ]
   },
   {
-    "Input": "मैं दिन में 7 बजे वापस जाउंगा",
+    "Input": "मैं अपराहन में 7 बजे वापस जाउंगा",
```
tellarin: Similar comments to previous PR.

Specs/DateTime/Hindi/DateTimePeriodParser.json
```diff
@@ -105,18 +101,18 @@
         "Start": 4,
         "Length": 31
       }
-    ]
+    ],
+    "NotSupported": "dotnet"
```
tellarin: Any specific reason not to support this case?
It's always helpful to add a "Comment" attribute to the spec case for documentation.

Specs/DateTime/Hindi/TimeExtractor.json
LionbridgeCSII: Same as before, the translation of these cases was not accurate. 
For example the case "I'll be back afternoon 7:00:05" in TimeParser.json was originally translated to "मैं दिन में 7:00:05 पर वापस आउंगी" where "दिन", meaning "day", does not give an indication of "am/pm". It has therefore been replaced with "अपराहन" which means "afternoon/pm"

Specs/DateTime/Hindi/DateTimePeriodParser.json
LionbridgeCSII: no specific reason, this is still a work in progress and we have not finished solving all the cases yet.

tellarin:
Could you resolve the conflicts and remove the 'Draft' mark from the next PRs? Thanks.

LionbridgeCSII:
@tellarin, rebased and pushed update with full support (all cases pass now).

PR #2007, opened 2019-12-22
[.NET] Hindi DateTime Merged partial support

LionbridgeCSII:
Extractor: all passing;
Parser: 35 passing, 21 failing.
DateTimeModel: 46|38


LionbridgeCSII:
Parser: 47 passing, 9 failing.

1 case removed from Extractor:
the case "What's the day look like?" was translated to "ये दिन कैसा लग रहा हैं " meaning "What's these days look like?" but a more accurate translation is "यह दिन कैसा लग रहा हैं" (What's this day look like?)
which however was already present in the json.

PR #2008, opened 2019-12-22
[.NET] Hindi DateTime DateTimeModel complete support

LionbridgeCSII:
DateTimeModel: 67 passing, 17 failing.


LionbridgeCSII:
83 cases passing, 1 case not supported.

Failing case:
"मैं हर तीन हफ्ते में बाहर जाउंगा" (I'll leave every three week). The problem is the same already encountered in Set: the case is literally "I'll leave every in three week" where "in three week" is extracted as Date instead of Duration and the final result is different.

PR #2013, opened 2019-12-30
[.NET] Dutch DateTime Date support

LionbridgeCSII:
All test cases passes.

Notes:
Some test cases were updated due to ambiguous translations or when the corresponding English test case made use of some particular arrangement of words.
BaseDurationParser has been modified to take into account the case when "business day" is just one word ("werkdag" in Dutch).

### DIFF
--- a/Patterns/Dutch/Dutch-DateTime.yaml
+++ b/Patterns/Dutch/Dutch-DateTime.yaml
@@ -11,7 +11,7 @@ RangeConnectorRegex: !nestedRegex
 ArticleRegex: !simpleRegex
   def: \b(de|het|een)\b
 RelativeRegex: !simpleRegex
-  def: (?<order>dit|deze|volgende?|komende?|aankomende?|aanstaande?|huidige?|vorige?|verleden|voorgaande?|vorige?|laatste|afgelopen)
+  def: \b(?<order>dit|deze|volgende?|komende?|aankomende?|aanstaande?|huidige?|vorige?|verleden|voorgaande?|vorige?|laatste|afgelopen|de)\b
 StrictRelativeRegex: !simpleRegex
   def: \b(?<order>dit|deze|volgende?|komende?|aankomende?|aanstaande?|huidige?|vorige?|verleden|voorgaande?|vorige?|laatste|afgelopen)\b
 UpcomingPrefixRegex: !simpleRegex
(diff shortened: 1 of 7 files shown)


LionbridgeCSII:
@iMicknl, PR ready for review.

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -738,10 +751,19 @@ DayOfWeek: !dictionary
     vrijdag: 5
     zaterdag: 6
     zondag: 0
+    maandagen: 1
```
iMicknl: @tellarin is this necessary? Or can this be normalized in code to use the day without the `en` suffix?

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -762,6 +784,25 @@ DayOfWeek: !dictionary
     zo.: 0
     zon: 0
     zon.: 0
+    monday: 1
```
iMicknl: @LionbridgeCSII why is this necessary? Are there situations in Dutch which require mixed English + Dutch recognition?

Patterns/Dutch/Dutch-Numbers.yaml
```diff
@@ -205,7 +205,7 @@ HalfADozenText: zes
 WordSeparatorToken: en
 WrittenDecimalSeparatorTexts: [komma]
 WrittenGroupSeparatorTexts: [punt]
-WrittenIntegerSeparatorTexts: [en]
+WrittenIntegerSeparatorTexts: [en, ën]
```
iMicknl: Are there tests in the DutchNumber parser for this?

Specs/DateTime/Dutch/DateExtractor.json
```diff
@@ -1085,40 +1059,37 @@
   },
   {
     "Input": "Ik ga de zesde zondag terug",
-    "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
-        "Text": "zesde zondag",
+        "Text": "zondag",
```
iMicknl: Isn't this supported in the English version? If so; I think this case should stay and be marked `NotSupported`.

Specs/DateTime/Dutch/DateExtractor.json
```diff
@@ -1467,15 +1435,14 @@
     ]
   },
   {
-    "Input": "Ik ga 21-04-'16 terug",
-    "NotSupported": "dotnet",
+    "Input": "Ik ga 21/04/16 terug",
```
iMicknl: Is 21-04-16 covered somewhere else? 14-04-2011 and 14.04.2011 are very common.

https://taaladvies.net/taal/advies/vraag/1489/datumnotatie/

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: this is necessary because with certain patterns the extractor extracts the English version of the weekday and then looks for it in DayOfWeek dictionary. So, we included it because it would fail otherwise. (Note that the same has been done for Italian and Turkish)

Patterns/Dutch/Dutch-Numbers.yaml
LionbridgeCSII: yes, for example "tweeënvijftig" and "tweeenvijftig" (fifty two) are both included in NumberModel.json

Specs/DateTime/Dutch/DateExtractor.json
LionbridgeCSII: it is supported in the English version, but also there only "sunday" is extracted from "i'll go back sixth sunday"

Specs/DateTime/Dutch/DateExtractor.json
LionbridgeCSII: I think this has been changed to match the English case "i'll go back 21/04/16". There are other test cases covering the patterns with "-" and "." separators.

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: @iMicknl, the regexes WeekDayRegex and SingleWeekDayRegex can be refactored so that the suffixed variations become unnecessary. (we proceeded in this way following the work done in Turkish where DayOfWeek also contains various suffixed entries). 
Should we update the file?

Specs/DateTime/Dutch/DateExtractor.json
tellarin: I'll add support for ""sixth Sunday" as an open issue across languages. In the context of this PR, I think the current behaviour is appropriate.

Patterns/Dutch/Dutch-DateTime.yaml
tellarin: This is probably something that should be refactored in the future. But it was marked as low priority to change.

Patterns/Dutch/Dutch-DateTime.yaml
tellarin: @LionbridgeCSII, I'll merge this PR as-is, but please make the suggested change in a follow up PR.

PR #2019, opened 2020-01-05
[CN DatetimeV2 .NET] Support cases like "从2015年起" and "2019年2月1日或之前"

Grey0202:
- Currently .NET only.
- Revise Mod part in Chinese DateTime.
- Add MergedTest.
- Revise some Regex to support the new features


Specs/DateTime/Chinese/TimeExtractor.json
```diff
@@ -58,10 +58,10 @@
     "Input": "今天大约晚上十点以后,微软大厦门口见",
     "Results": [
       {
-        "Text": "大约晚上十点以后",
+        "Text": "大约晚上十点",
```
Grey0202: This change is because "大约晚上十点以后" now belongs to the merge part.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -70,6 +70,9 @@ public static class Constants
         // "until" -> Same as "before", but including the extracted datetime entity. Equivalent to "<="
         public const string UNTIL_MOD = "until";
 
+        // Include the date metioned, to make "before" -> "until" or "after" -> "since". Such as "on or earlier than 1/1/2016".
+        public const string INCLUDE_MOD = "include";
```
tellarin: This is not a modifier, this is a named group. It should be properly named and added in the 
// Groups' names for named groups in regexes
section.

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -3894,13 +3894,12 @@
...
     },
-    "Comment": "This is notsupported for now because of the FromToRegex",
-    "NotSupported": "dotnet, java, javascript, python",
+    "NotSupported": "java, javascript, python",
     "Results": [
       {
-        "Text": "从2015年",
-        "Start": 0,
-        "End": 5,
+        "Text": "2015年起",
```
tellarin: Why the change? What did the previous text mean?

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -3921,13 +3920,12 @@
...
     },
-    "Comment": "This is notsupported for now because of the FromToRegex",
-    "NotSupported": "dotnet, java, javascript, python",
+    "NotSupported": "java, javascript, python",
     "Results": [
       {
-        "Text": "自2014年6月14日",
-        "Start": 5,
-        "End": 15,
+        "Text": "2014年6月14日起",
```
tellarin: Same here. Can you explain the difference? I don't speak Chinese. :-(

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
```diff
@@ -219,5 +233,35 @@ private void AddTo(List<ExtractResult> dst, List<ExtractResult> src)
                 }
             }
         }
+
+        // Avoid adding mod for ambiguity cases, such as "从" in "从 ... 到 ..." should not add mod
+        // TODO: Revise PotentialAmbiguousRangeRegex to support cases like "从2015年起，哪所大学需要的分数在80到90之间"
```
tellarin: How big do you estimate is the effort to cover this case?

Specs/DateTime/Chinese/DateTimeModel.json
Grey0202: “自2014年6月14日起” is like "from 2014-6-14 begin", the ideal result is the whole sentence. But because of the issue related to PotentialAmbiguousRangeRegex, I did a workaround to get part of the result. “自2014年6月14日起” and  "2014年6月14日起" mean the same in chinese.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseMergedExtractorConfiguration.cs
Grey0202: Currently, in En "**from** 2015, I have wanted **to** buy a new car" it will only get "2015", not from 2015.
I'm not sure how to fix this bug.

PR #2020, opened 2020-01-07
Fixing bug with "weekend" flag on Week in year, e.g. "2020-W02-WE"

HiltonGiesenow:
This was actually raised in a StackOverflow question - see https://stackoverflow.com/questions/59615953/luis-test-tool-datetime-different-than-locally-resolved-datetime

Basically, the TimexResolver is ignoring if the timex has a "weekend" attached to it, to get the start and end of the period.


HiltonGiesenow:
wow, that was quick! thanks!

HowdySys:
Big thanks to all of you, and especially to @HiltonGiesenow !
- Roy (Stack Overflow Poster)

HiltonGiesenow:
glad to help!

PR #2022, opened 2020-01-07
[CN NumberRange .NET] Revise regex to recognize additional range cases

Grey0202:
- Add support for cases like "至少是", “至少为”, "最多有".
- Add support for "大于0并小于2"
- Add support for "≤" and "≥"
- Contains some missing regex changes generated by build.


Python/libraries/recognizers-date-time/recognizers_date_time/resources/english_time_zone.py
```diff
@@ -9,7 +9,6 @@
 # Licensed under the MIT License.
 # ------------------------------------------------------------------------------
 
-from .base_date_time import BaseDateTime
```
tellarin: Why remove this?

Grey0202: This line is removed by the build. I think someone removed it in other commits and forget to re-generate this file.

PR #2025, opened 2020-01-15
[.NET NumberRange] Filter wrong number range such as "more than 5 and less than 0"

Grey0202:
- Fix the range bug in all languages.
- Break wong number range into two parts.
- Fix a number range regex mistake in Spanish
- Improve the way label entity to support "大于10小于0"
- Add some cases.


.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -50,15 +50,53 @@ public virtual List<ExtractResult> Extract(string source)
...
                     if (start >= 0 && length > 0)
                     {
-                        for (var j = 0; j < length; j++)
+                        // Keep Source Data for extra information
+                        matchSource.Add(new Tuple<int, int>(start, length), collection.Value);
+                    }
+                }
+            }
+
+            foreach (var match in matchSource)
```
tellarin: Can you add some code comments here to explain the purpose of the change?

.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberRangeExtractor.cs
```diff
@@ -69,7 +107,7 @@ public virtual List<ExtractResult> Extract(string source)
                     if (i + 1 == source.Length || !matched[i + 1])
                     {
                         var start = last + 1;
-                        var length = i - last;
+                        var length = i - last + 1;
```
tellarin: Why this change? There is no spec change to show the effect of this.

Specs/Number/Chinese/NumberRangeModel.json
```diff
@@ -721,12 +721,21 @@
...
+        }
+      },
+      {
+        "Text": "小于12",
+        "Start": 4,
         "End": 7,
         "TypeName": "numberrange",
         "Resolution": {
-          "value": "(12,50)"
+          "value": "(,12)"
```
tellarin: Can you add a complex case here, similar to the last one you added in English that results in three ranges (two separate, and one "merged")?

PR #2027, opened 2020-01-19
[FR DateTimeV2] Fix French datetime matches numbers

### LINKED ISSUE #1523: [FR DateTimeV2] French datetime matches numbers
For some reason DateTime Model in French matches numbers from 0 to 24
Example: "Je veux le numéro 3", `3` is matched as datetime when it should not.

Grey0202:
- This fixes #1523
- The single number now will not be recognized as Time
- Add some tests.


Specs/DateTime/French/DateTimeModel.json
```diff
@@ -1781,5 +1781,341 @@
...
+  {
+    "Input": "Dix heures mercredi matin",
+    "Context": {
+      "ReferenceDateTime": "2019-08-12T00:00:00"
+    },
+    "Comment": "Not supported because of an extra result.",
+    "NotSupported": "dotnet, javascript, java, python",
+    "Results": [
+      {
+        "Text": "dix heures mercredi mati",
```
tellarin: It should end in "matin".

Specs/DateTime/French/DateTimeModel.json
```diff
@@ -1781,5 +1781,341 @@
...
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Dix heures mercredi matin",
+    "Context": {
+      "ReferenceDateTime": "2019-08-12T00:00:00"
+    },
+    "Comment": "Not supported because of an extra result.",
```
tellarin: What do you mean by "extra result"?

Specs/DateTime/French/SetExtractor.json
```diff
@@ -33,13 +33,13 @@
     ]
   },
   {
-    "Input": "Je vais partir 9 lundis",
```
tellarin: Does "Je vais partir à 9 lundis" work?

Grey0202: "9 lundis" can be extracted but without parse result the same as the previous version.

tellarin: OK. Please add this case in the spec.

PR #2028, opened 2020-01-22
[ES DateTime] Revise PrepositionRegex to avoid mistake dateRange

### LINKED ISSUE #1958: [ES DateTimeV2] DateRange incorrectly detected instead of single DateTime
When I write my sentence with a datetimeV2 and the word **_una_** it recognizes that there is a datetimeV2 but detect the entity as a **datetimerange** taking the entity from the word una

- organizame **una** reunion con Francesco Sparacio **mañana a las 12** 
- organizame **una** reunion **el 27/11 a las 24**

I think this might be related to the fact that the article “una” is also meaning an hour in Spanish (1 o clock) but the resulting detected entity is wrong making inusable the DateTimeV2 recognition

Grey0202:
- This fixes #1958 
- Cases like "una reunión el viernes en londresz" will no longer merge.
- Disable some cases because they could be ambiguous (also not supported in EN).
- TODO: "programa una reunión" will get "una ", contain a space.


iMicknl:
@lomarmo since you identified the issue & are native Spanish. Will this PR fix the issue with `una` we noticed? (have a look at [DateTimeModel.json](https://github.com/microsoft/Recognizers-Text/pull/2028/files#diff-9461b31abbb322828e022f134160620d)).

Is `una` always one? Or only in special cases. I think it is not possible to ignore una when it is placed in front of another noun.

lomarmo:
> @lomarmo since you identified the issue & are native Spanish. Will this PR fix the issue with `una` we noticed? (have a look at [DateTimeModel.json](https://github.com/microsoft/Recognizers-Text/pull/2028/files#diff-9461b31abbb322828e022f134160620d)).
> 
> Is `una` always one? Or only in special cases. I think it is not possible to ignore una when it is placed in front of another noun.

Actually, "una" should only be considered a DateTime element when it is preceeded by "la" ("la una"). In all other cases, it is a femenine quantifier (it is used to indicate one single element).

I'll give you some examples. In all cases, **una** means "one instance of" and _la una_ refers to a specific time:

(1) Tengo **una** reunión a _la una_ -> I have **a** meeting at _one_
(2) Planifica **una** visita el próximo lunes a las doce -> Schedule **a** visit next monday at twelve
(3) Tengo que ver a **una** persona el sábado -> I have to meet **one** person next saturday
(4) Tengo **una** reunión desde _la una_ hasta las tres -> I have **a** meeting from _one_ to three
(5) Tengo **una** cita desde las doce hasta la _una_ -> I have **an** appointment from twelve to _one_

I hope this clarifies the question.

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -597,5 +597,161 @@
...
+  },
+  {
+    "Input": "organizame una reunion con Francesco Sparacio mañana a las 12",
+    "Context": {
+      "ReferenceDateTime": "2020-01-23T01:00:00"
+    },
+    "NotSupported": "javascript, python, java",
+    "Results": [
+      {
+        "Text": "una ",
```
tellarin: This is incorrect. Can you take a look at the examples provided in the issue for the cases where "una" should be time?
Isn't this similar to what you did in French?

tellarin:
Closing this old PR as the issue has been solved elsewhere.

PR #2051, opened 2020-03-16
[SV PhoneNumbers] Add support for swedish phone numbers

pete1854:
Adding support for phone numbers. All test pass.


Patterns/Base-PhoneNumbers.yaml
```diff
@@ -35,6 +35,9 @@ ITPhoneNumberRegex: !paramsRegex
 NLPhoneNumberRegex: !paramsRegex
   def: ((((\(\s?(\+\s?|00)31\s?\)\s?)|(((?<!\d)\+\s?|{WordBoundariesRegex}00)31\s?))?((({WordBoundariesRegex}|(?<=31))0?\d{1,3}|\(\s?0?\d{1,3}\s?\)|\(0\)[-\s]?\d{1,3})\s?-?[\d\s]{5,11}\d))|\b\d{10,12})(?!-){EndWordBoundariesRegex}
   params: [ WordBoundariesRegex, EndWordBoundariesRegex  ]
+SEPhoneNumberRegex: !paramsRegex
```
tellarin: Can you merge the NL and SE regexes? Just an or between the international codes seems to be sufficient.
This would have better performance. Please add a comment in the YAML file saying this regex combines the two. Also, you can rename the regex too.
Thanks!

pete1854: Do you mean that I should replace NLPhoneNumberRegex and SVPhoneNumberRegex with something like "NLSVPhoneNumberRegex", or what? I'm not quite sure I understand what you mean. What if a future requirement need the Dutch or Swedish contributor to make changes to their Regex? Please clarify.

tellarin: Maybe EUPhoneNumberRegex? The idea is that this one should be the general one. Only specific cases that diverge from it should have their own regex. Also, these numbers should be recognized in any language.
Only if a number format is only used locally and causes too many false positives in other languages, then it should have it's own regex and used only in the language-specific recognizer. Which is not the case for the ones caught in SEPhoneNumberRegex.
Does this make sense?

pete1854: Thanks for clarifying. I will try and update the PR in a few days.

pete1854:
I seem to have introduced a regression in the English phone number recognizer, though I can’t figure out what causes it. I haven’t made any changes other than to the swedish/dutch code... Any clues?

Patterns/Base-PhoneNumbers.yaml
pete1854: I seem to have introduced a regression in the English phone number recognizer, though I can’t figure out what causes it. I haven’t made any changes other than to the swedish/dutch code... Any clues? If you look at the build log you'll find two cases in the English test were it fails.

tellarin: Thanks for the updates! We'll try to take a closer look ASAP.

pete1854: Hi @tellarin,
I would like to cancel this PR. I have revised the pattern to fully comply w/ the SIS standard for phone numbers in sweden. The pattern now differs substantially from the dutch and I would therefore like to ADD a pattern for being able to create a swedish recognizer specifically. Ok?

tellarin:
Closing this as it's now superseded by #2192.

PR #2062, opened 2020-03-30
[.NET Number] Add false positive filter for Chinese

visionshao:
I added the false positive filter for Chinese NumberExractor according to English NumberExtractor. For this, I added a AmbiguityFilterDict in Chinese NumbersDefinitions of Recognizers.Definitions.


.NET/Microsoft.Recognizers.Definitions.Common/Chinese/NumbersDefinitions.cs
```diff
@@ -255,5 +255,9 @@ public static class NumbersDefinitions
         {
             { @"", @"" }
         };
+      public static readonly Dictionary<string, string> AmbiguityFiltersDict = new Dictionary<string, string>
```
tellarin: Can you add the modified YAML file that auto generates this too?

PR #2073, opened 2020-04-24
Extend date range to cover "end/beginning/middle of $year" in English

visionshao:
- Modified LaterEarlyPeriodRegex in English to also match years; 
- Added clauses to BaseDatePeriodParser to resolve correctly; 
- New test specs.


Specs/DateTime/English/DateTimeModelComplexCalendar.json
```diff
@@ -12667,5 +12667,83 @@
...
+        "End": 49,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "2000",
+              "Mod": "mid",
+              "type": "daterange",
+              "start": "2000-05-01",
+              "end": "2000-08-31"
```
tellarin, line 12716: This should be 2000-09-01, as the end date is not included, right? End and begin behave this way.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1098,6 +1098,43 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
...
+                        else if (midPrefix)
+                        {
+                            beginDate = DateObject.MinValue.SafeCreateFromValue(year, 5, 1);
+                            endDate = inclusiveEndPeriod ?
+                                DateObject.MinValue.SafeCreateFromValue(year, 8, 30) :
+                                DateObject.MinValue.SafeCreateFromValue(year, 8, 30).AddDays(1);
+                        }
+                        else if (latePrefix)
+                        {
+                            beginDate = DateObject.MinValue.SafeCreateFromValue(year, 9, 1);
```
tellarin, line 1133: Why no endDate here?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1098,6 +1098,43 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
...
+
+                        ret.Success = true;
+                        return ret;
+                    }
+                    else if (!string.IsNullOrEmpty(match.Groups["FourDigitYear"].Value))
+                    {
+                        var date = referenceDate.AddYears(swift);
+                        year = int.Parse(match.Groups["FourDigitYear"].Value);
+
+                        var beginDate = DateObject.MinValue.SafeCreateFromValue(year, 1, 1);
```
tellarin, line 1113: Please add a code comment here describing the used resolution definition:
// Early/mid/late are resolved in this policy to 4 month ranges at the start/middle/end of the year.

Specs/DateTime/English/DateTimeModelComplexCalendar.json
visionshao: get it!

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
visionshao: there is a endDate at the beginning of this function and latePrefix part will use this default endDate.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
visionshao: ok!

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
tellarin: ok

visionshao:
PS: in this PR, model don't support "since/from the end of $Year"

PR #2075, opened 2020-04-29
Arabic Number Model and Percent Model Pull Request

neudurgeshp:
**Dev Changes**
- NumberModel (Covered 98 Cases / In progress 13 Cases)
	Integer
	Decimal
	Fraction
	Power
-Percent Model؜ (Covered 100%)
**Test Specs**	
-NumberModel Test Specs: 111
-PercentModel Specs: 100%


.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
```diff
@@ -81,7 +81,7 @@ public NumberRangeModel GetNumberRangeModel(string culture = null, bool fallback
 
         protected override void InitializeConfiguration()
         {
-            /*
+
             RegisterModel<NumberModel>(
```
tellarin, line 85: Is the build failure because tests are enabled and they are trying to validate resolution implementation? Or some other reason?

neudurgeshp: I don't think so. I think this is because of duplicate test case and that will be resolved in next PR Today. You can see in attached in-line image. 
MicrosoftTeams-image

Specs/Number/Arabic/PercentModel.json
```diff
@@ -1,180 +1,222 @@
...
         },
-        "Start": 0,
-        "End": 3
+        "Start": 25,
+        "End": 28
       }
     ]
   },
   {
-    "Input": " 100% ",
```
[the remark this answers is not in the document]
sandeep-sankaran: We have planned a different PR for colloquial usage, negative cases and few variations. We will have a discussion on this before including these cases into the test specs.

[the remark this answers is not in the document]
tellarin: Adding cases in separate PRs is fine. But please don't delete cases that already exist in the specs unless they are invalid.
If certain cases have not yet been validated, add a "Comment" attribute stating that.
Otherwise it's very hard to track why different cases were editted/added/removed.

.NET/Microsoft.Recognizers.Text.Number/NumberRecognizer.cs
tellarin: Please don't create a new PR. You can update this one.
The repository also has a script to check for duplicate cases. I suggest running it before any PRs.

neudurgeshp: Understood. I just updated this one only and built successfully. Moving forward we'll be validating for duplicate cases before raising the PRs.

Specs/Number/Arabic/PercentModel.json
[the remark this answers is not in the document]
sandeep-sankaran: Added the existing test cases with comments Pending Validation in both Number Model and Percent Model.

Specs/Number/Arabic/NumberModel.json
```diff
@@ -1738,8 +4345,9 @@
   },
   {
     "Input": "1 في واحد و عشرون",
-    "NotSupported": "javascript, python",
+    "Comment": "PendingValidation",
     "NotSupportedByDesign": "javascript,python,java",
+    "NotSupported": "javascript, python, dotnet",
```
tellarin, line 4350: Minor issue: should be only dotnet, as other are already in NotSupportedByDesign

Specs/Number/Arabic/NumberModel.json
```diff
@@ -1808,8 +4422,9 @@
   },
   {
     "Input": "سيتم التعرف على 423 0000 كرقمين.",
-    "NotSupported": "javascript, python",
+    "Comment": "PendingValidation",
     "NotSupportedByDesign": "javascript,python,java",
+    "NotSupported": "javascript, python, dotnet",
```
tellarin, line 4427: Same.

neudurgeshp:
> Approved with comments.
> Please make sure the usage of NotSupported/NotSupportedByDesign is consistent. A few cases are not.
> Do you have a tentative ETA by when the previous spec cases can be validated?
> Lastly, when can we expect the PRs for the other Arabic number types?
> Thanks!

- We will be raising PRs for other Arabic number types Today  by EOD.
- We will send separate mail on ETA for previous spec validation
Thank-you!

Specs/Number/Arabic/NumberModel.json
sandeep-sankaran: These were existing test cases from the older spec. We have added dotnet to NotSupported as part of the PR.  A total of 9 such cases were found in NumberModel and we have fixed it. A new pull request 2083 is raised where these changes are included.

Issue #2047, opened 2020-03-11
[EN Number] Entities recognized incorrectly in input cases like "5k-20k"


perkins4j2:
**Describe the bug**
Don't recognize 5k-20k

**To Reproduce**
1. code
```java
        Gson gson = new GsonBuilder().create();
        System.out.println("====A");
        String input = "5k-20k";
        System.out.println("input:" + input);

        System.out.println("====B");
        List<ModelResult> results = NumberRecognizer.recognizeNumber(input, Culture.Chinese);
        results.forEach(e -> {
            System.out.println(gson.toJson(e));
        });

        System.out.println("====C");
        results = NumberRecognizer.recognizeNumberRange(input, Culture.Chinese);
        results.forEach(e -> {
            System.out.println(gson.toJson(e));
        });
```

2.result
====A
input:5k-20k
====B
WARN: Look-behind groups with no maximum length not supported. Java version <= 8.
====C
{"text":"5k-20k","start":0,"end":6,"typeName":"numberrange","resolution":{}}

3.error
- recognizeNumber don't recognize
- recognizeNumberRange recognized,but `resolution is null`

**Expected behavior**
- recognizeNumber:5000,20000
- recognizeNumberRange:5000 - 20000

**Platform (please complete the following information):**
 - Platform:  Java
 - Environment: package & console
 - Version of package : master

perkins4j2:
debug info:
class:BaseNumberExtractor
method:extract(String source)
line:Match[] matches = RegExpUtility.getMatches(k, source);

k=((?<!(\d{1,10}\s{0,10})|[-－])[-－])?[\d１２３４５６７８９０]+\s*(K|k|M|G|T|B|b)
source=5k-20k

result:
5k
-20k

so modify code:
class:ChineseNumeric
filed:NegativeNumberTermsRegexNum
`public static final String NegativeNumberTermsRegexNum = "((?<!(\\d+\\s*)|[-－])[-－])";`
should be:
`public static final String NegativeNumberTermsRegexNum = "((?<!(\\d+\\s*)|[-－]))";`

test again :
I found the following entities (2):
{
  "text" : "5k",
  "start" : 0,
  "end" : 2,
  "typeName" : "number",
  "resolution" : {
    "value" : "5000"
  }
}
{
  "text" : "20k",
  "start" : 3,
  "end" : 6,
  "typeName" : "number",
  "resolution" : {
    "value" : "20000"
  }
}

[...]

PR #2077, opened 2020-04-30
[.NET] English NumberRange bug #2047 fix

LionbridgeCSII:
fix to bug #2047 
added relevant test case to NumberRangeModel
(we noticed that the test DataTypes_Resolver_DateTime_TuesAt12PM_UtcInput fails (regardless of the above changes))


tellarin:
@LionbridgeCSII, we'll review the PR soon.
Regarding the failure you mention, it didn't happen in this build. Perhaps it's due to some difference in your local environment?
Please open a lower priority issue to document it, so we can analyse later.
Thanks.

tellarin:
@LionbridgeCSII, the original issue also mentions Chinese. Can you fix that as well?

PR #2083, opened 2020-05-04
Arabic Ordinal and NumberRange & Spec updates

neudurgeshp:
- Spec modified as per change request 
   https://github.com/microsoft/Recognizers-Text/pull/2075#issuecomment-623312089
- Ordinal updates
- Number Range updates


.NET/Microsoft.Recognizers.Text.Number/Arabic/Extractors/NumberRangeExtractor.cs
```diff
@@ -88,5 +100,200 @@ public NumberRangeExtractor(INumberOptionsConfiguration config)
         internal sealed override Regex AmbiguousFractionConnectorsRegex { get; }
 
         protected sealed override string ExtractType { get; } = Constants.SYS_NUMRANGE;
+
+        public override List<ExtractResult> Extract(string source)
```
tellarin, line 104: This looks like the same base NumberRangeExtractor in other languages. Can't the code be changed to re-use?
Ideally we should introduce specific language-specific customization points, but avoid duplicating code. Otherwise code will diverge and issues would need to be fixed in multiple places.

neudurgeshp: Agreed, will be fixing that  and will update the same PR.

tellarin:
@neudurgeshp the build/test pipeline now supports a new attribute "IgnoreResolution" to be used in the specs to skip resolution, but still validate type/extraction/properties.
Can you do a separate PR removing the previous "NotSupported: dotnet", and adding the appropriate "IgnoreResolution"?

neudurgeshp:
> @neudurgeshp, the changes look good.
> But before merging, we should do two things:
> 
> * include the re-generated Arabic resources after the changes in the Arabic YAML (part of this PR);
> * enable the test execution for Ordinal and NumberRange (even if all tests are currently skipped until the IgnoreResolution is in use).

@tellarin 
Sure,  we will be including the regenerated resources going forward into the PR.
It's good to have IgnoreResolution now. So for test execution we will be raising a new PR as you asked.
Thanks!

neudurgeshp:
> @neudurgeshp the build/test pipeline now supports a new attribute "IgnoreResolution" to be used in the specs to skip resolution, but still validate type/extraction/properties.
> Can you do a separate PR removing the previous "NotSupported: dotnet", and adding the appropriate "IgnoreResolution"?

Sure, we'll be doing separate PR for the same.

tellarin:
It seems I can't push to this PR, so I'll merge and make the pending change.
When creating PRs, please allow repo owners to push changes.

neudurgeshp:
> It seems I can't push to this PR, so I'll merge and make the pending change.
> When creating PRs, please allow repo owners to push changes.

Is that for me?

tellarin:
> > It seems I can't push to this PR, so I'll merge and make the pending change.
> > When creating PRs, please allow repo owners to push changes.
> 
> Is that for me?

Yes. I tried re-enabling the tests here and pushing to the PR, but was denied. So I had to create PR #2086.

PR #2092, opened 2020-05-06
[.NET] Dutch Number NumberRangeModel support

LionbridgeCSII:
all cases are passing


iMicknl:
> Looks OK to me. @iMicknl, could you take a quick look at the pattern/spec changes?

Looks good. The only changed testcase is for 'tussen' (between), where it also includes the first number. Not sure how `between` is handled in English, but this makes sense for me.

PR #2100, opened 2020-05-08
[JA Ordinal] Improvements to coverage

neudurgeshp:
- Spec (96) including Existing-12 and New-84
- Not supported: 8
- Ordinal Extractor improvement


neudurgeshp:
@tellarin one test case failed for ordinal for input "第一位第二位第三位第四位" for javascript support.
What your suggestion on that?
Do you want us to mark that to not supported for JavaScript or something else?

tellarin:
> @tellarin one test case failed for ordinal for input "第一位第二位第三位第四位" for javascript support.
> What your suggestion on that?
> Do you want us to mark that to not supported for JavaScript or something else?

Please mark it as NotSupported in JavaScript.

PR #2102, opened 2020-05-09
[.NET] Dutch DateTime Time support

LionbridgeCSII:
all test cases pass


LionbridgeCSII:
@tellarin, then should we enable the Dutch DateTimeModel in DateTimeRecognizer.cs in the upcoming PRs? 
I remember for Hindi, we used to only report the number of pass/fail in the DateTimeModel without actually enabling it until the end.

tellarin:
It shouldn't be enabled until the end. But, if possible, adding the pass/fail numbers in the PR description also for DateTimeModel, helps having a bigger picture than just the extractor/parser for each sub-type.

Issue #945, opened 2018-10-31
[EN DatetimeV2] Fix range cases like "10/1-11/2/2017"


guom08:
**Describe the bug**
Currently, this can't be recognized as a range.
The reason is because the Regex will first match the "10/1-11", after that "11/2/2017" will not be matched because the "11" is overlapped.

**Expected behavior**
The output should be "10/1/2017" - "11/2/2017" with "2017" as the context year

PR #2104, opened 2020-05-09
[.NET] English DatePeriod bug #945 fix

LionbridgeCSII:
Fixes bug #945

### DIFF
--- a/Patterns/English/English-DateTime.yaml
+++ b/Patterns/English/English-DateTime.yaml
@@ -250,6 +250,7 @@ DateExtractor3: !nestedRegex
 DateExtractor4: !nestedRegex
   def: \b{MonthNumRegex}\s*[/\\\-]\s*{DayRegex}[\.]?\s*[/\\\-]\s*{DateYearRegex}
   references: [ MonthNumRegex, DayRegex, DateYearRegex ]
+# The final lookahead in DateExtractor5 avoids extracting as date "10/1-11" from an input like "10/1-11/2/2017"
 DateExtractor5: !nestedRegex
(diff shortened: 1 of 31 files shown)


Patterns/English/English-DateTime.yaml
```diff
@@ -251,7 +251,7 @@ DateExtractor4: !nestedRegex
   def: \b{MonthNumRegex}\s*[/\\\-]\s*{DayRegex}[\.]?\s*[/\\\-]\s*{DateYearRegex}
   references: [ MonthNumRegex, DayRegex, DateYearRegex ]
 DateExtractor5: !nestedRegex
-  def: \b{DayRegex}\s*[/\\\-\.]\s*({MonthNumRegex}|{MonthRegex})\s*[/\\\-\.]\s*{DateYearRegex}
+  def: \b{DayRegex}\s*[/\\\-\.]\s*({MonthNumRegex}|{MonthRegex})\s*[/\\\-\.]\s*{DateYearRegex}(?!\s*[/\\\-\.]\s*\d+)
```
tellarin, line 254: My understanding is that this is making the regex not match if some suffix exists, correct? If so, this seems a reasonable solution. 
However, I think it would be important to add a comment in the YAML file explaining the reason for this component in the regex. This will greatly serve future maintainability and extension to new languages.
Also, this fix seems to target a general problem, so it should be made across languages too. Not only English (at least in languages were such entity mention form is plausible).

LionbridgeCSII: yes, correct. 
Ok, we will add the comment and check the behaviour in the other languages.

tellarin: Do you have an estimate on when you can update the PR? It's getting little stale and now has a conflict with /master.

LionbridgeCSII: @tellarin, just pushed the update

PR #2106, opened 2020-05-12
[JA Fraction] Spec Update and Extractor - Negative support for both Kanji and HW

neudurgeshp:
**PR Includes:**
- **Extractor:** Negative support for both Kanji and HW
- **Spec Update:** Removed NotSupported for fraction
- Included Number Definitions

### DIFF
--- a/Specs/Number/Japanese/NumberModel.json
+++ b/Specs/Number/Japanese/NumberModel.json
@@ -1913,7 +1913,6 @@
     "Input": "出生率は−５分の４です。",
     "IgnoreResolution": true,
     "NotSupportedByDesign": "javascript, java, python",
-    "NotSupported": "dotnet",
     "Results": [
       {
         "Text": "-5分の4",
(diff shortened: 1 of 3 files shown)


Specs/Number/Japanese/NumberModel.json
```diff
@@ -3153,7 +3148,7 @@
     "Input": "答えは−３５分の１００です。",
     "IgnoreResolution": true,
     "NotSupportedByDesign": "javascript, java, python",
-    "NotSupported": "dotnet",
+
```
tellarin, line 3151: Nitpick: Please remove blank lines from the specs.

neudurgeshp: Sure @tellarin . That will be fixed in coming PR.

neudurgeshp: Fixed in https://github.com/microsoft/Recognizers-Text/pull/2109

PR #2109, opened 2020-05-14
[JA Number] Decimal Coverage Improvements

neudurgeshp:
**PR Includes:**

- Double Extractor support
- Spec modifications
   - Removed NotSupported for decimal,
   - Removed blankspace (https://github.com/Microsoft/Recognizers-Text/pull/2106#discussion_r424116178)
- Included NumberDefinitions


Patterns/Japanese/Japanese-Numbers.yaml
```diff
@@ -194,6 +197,12 @@ DoubleAllFloatRegex: !nestedRegex
 DoubleExponentialNotationRegex: !nestedRegex
   def: (?<!{ZeroToNineFullHalfRegex}+[\.．])({NegativeNumberTermsRegexNum}\s*)?{ZeroToNineFullHalfRegex}+([\.．]{ZeroToNineFullHalfRegex}+)?e(([-－+＋]*[1-9１２３４５６７８９]{ZeroToNineFullHalfRegex}*)|[0０](?!{ZeroToNineFullHalfRegex}+))
   references: [ZeroToNineFullHalfRegex, NegativeNumberTermsRegexNum]
+DoubleExponentialNotationRegex1: !nestedRegex
+  def: (?<!{ZeroToNineFullHalfRegex}+[\.．])(({NegativeNumberTermsRegexNum}|{NegativeNumberTermsRegex})\s*)?{ZeroToNineFullHalfRegex}+([\.．]{ZeroToNineFullHalfRegex}+)?(×)?(10)?(e|の)(([-－+＋]*[1-9１２３４５６７８９]{ZeroToNineFullHalfRegex}*)[乗]|[0０](?!{ZeroToNineFullHalfRegex}+))
```
tellarin, line 201: Why do we need a separate regex from DoubleExponentialNotationRegex?
Also, aren't the full-width digits in both redundant (due to pre-processing)? Removing them would bring some performance benefit.

Specs/Number/Japanese/NumberModel.json
```diff
@@ -8583,6 +8571,7 @@
     "Input": "毎日この薬を〇・二五ミリリットル飲んでください。",
     "IgnoreResolution": true,
     "NotSupportedByDesign": "javascript, java, python",
+    "NotSupported": "dotnet",
```
tellarin, line 8574: Why is this case now not supported? If this is a temporary side effect, please add a "Comment" attribute in the test spec describing the reason.

Patterns/Japanese/Japanese-Numbers.yaml
neudurgeshp: @tellarin 
That makes sense. We did that for the exponential new cases. For now we are going to revert that particular regex DoubleExponentialNotationRegex1 from the PR. We'll revisit those cases and will fix that.

Specs/Number/Japanese/NumberModel.json
neudurgeshp: Sure, we'll update that.

PR #2111, opened 2020-05-14
[JA NumberRange] Extractor Support and Added New Cases

neudurgeshp:
**PR Includes:**

- Extractor
- Spec (Cases-134)
    - Existing: 21, 
    - New : 113 (102 Passed, 11 Skipped [NotSupported])
- Number Definitions
- Enabled Test for NumberRangeModel


Specs/Number/Japanese/NumberRangeModel.json
```diff
@@ -322,5 +342,1785 @@
...
+  {
+    "Input": "この数は１００より大きく３００より小さいです。",
+    "IgnoreResolution": true,
+    "NotSupportedByDesign": "java, javascript, python",
+    "Results": [
+      {
+        "Text": "100より大きく300より小さい",
+        "TypeName": "numberrange",
+        "Resolution": {
+          "value": "-100,300"
```
tellarin, line 419: It seems some of the specs are not fully complete. Please double check values.
By default, every interval is closed at the start (includes the starting point) and open at the end (does not include the end point). Modifiers will change this behaviour. Please look into English and Chinese for examples.

neudurgeshp: Re-validated and made changes

neudurgeshp:
> Mostly looks good. The new NotSupported cases for extraction will be refined in a follow up PR, correct?
> Thanks.

Yes, as we planned we would be revisiting the Pending/NotSupported for Arabic and Japanese after KO iterations.

PR #2113, opened 2020-05-17
[.NET] Hindi Number NumberRangeModel support

LionbridgeCSII:
74 cases passing
2 cases failing

Failing cases:

- "x 10 से बड़ी और 20 से छोटी है. y 50 से अधिक नहीं है और 20 से कम नहीं है." (x is larger than 10 and smaller than 20. y is no more than 50 and no less than 20.)
- "उनका स्कोर 5000 के बराबर या 5000 से कम है" (His score is equal to 5000 or less than 5000)

The problem is the same observed in Turkish (https://github.com/microsoft/Recognizers-Text/pull/1947): since the modifier "less than" follows the number, the method ValidateMatchAndGetStartAndLength in BaseNumberRangeExtractor always returns as start index the index of the first occurrence of the number.

Some cases where updated due to incorrect translation. We left comments directly in the json file, is it ok? Or is it better to move the comments here?

Modified cases:

- "यह संख्या तीन हजार नौ सौ पैंसठ से बड़ी है।" (This number is bigger or equal than three thousand nine hundred and sixty five.), but the hindi case translates to "This number is larger than three thousand nine hundred and sixty five.". So, it has been updated.
- "इन उत्पादों में पांच सौ से ज़्यादा हैं." (There're about five hundred and more in these products.), but the hindi case translates to "There are more than 500 of these products.", and "more than x" resolves to '(x,)' while "x and more" resolves to '[x,)'.
- "5 में 2 या ज़्यादा कैसा रहेगा" (How about 2 or more in 5)
- "यह संख्या 1998 में 20000 के बराबर है" (This number is equal to 20000 in 1998)
- "यह संख्या 2008 में 200 से 300 है." (This number is 200 to 300 in 2008.)
- "यह संख्या 2008 में 200 से 3000000 है" (This number is 200 to 3000000 in 2008)

For the last 4 cases, they all involve wrong fractions like "5 में 2" ("5 in 2" instead of "2 in 5"), so they have been updated to "5 में से 2" which translates to "2 in 5"


.NET/Microsoft.Recognizers.Text.Number/Hindi/Extractors/NumberRangeExtractor.cs
```diff
@@ -13,7 +13,7 @@ public NumberRangeExtractor(INumberOptionsConfiguration config)
             : base(
                   NumberExtractor.GetInstance(),
                   OrdinalExtractor.GetInstance(),
-                  new BaseNumberParser(new HindiNumberParserConfiguration(config)),
+                  new BaseIndianNumberParser(new HindiNumberParserConfiguration(config)),
```
tellarin, line 16: Question: Could BaseIndianNumberParser be unified with BaseNumberParser and language specific behaviour be contained in it's parser configuration object? That would be preferable, if possible.

Specs/Number/Hindi/NumberRangeModel.json
```diff
@@ -211,43 +198,41 @@
...
         "Start": 7,
-        "End": 34
+        "End": 35
       }
     ]
   },
   {
-    "Input": "यह संख्या तीन हजार नौ सौ पैंसठ से बड़ी है।",
+    "Input": "यह संख्या तीन हजार नौ सौ पैंसठ से बड़ी या उसके समान है।",
+    "Comment": "Retranslated due to incorrect translation.",
```
tellarin, line 218: Description of changes should happen at the PR level. The "Comment" field in the specs is specific for comments related to the spec itself, for example, some case not currently supported and the reason.

Specs/Number/Hindi/NumberRangeModel.json
```diff
@@ -281,45 +265,43 @@
...
         "Start": 10,
-        "End": 18
+        "End": 23
       }
     ]
   },
   {
-    "Input": "इन उत्पादों में पांच सौ से ज़्यादा हैं.",
+    "Input": "इन उत्पादों में पांच सौ और ज़्यादा हैं.",
+    "Comment": "Retranslated due to incorrect translation.",
```
tellarin, line 299: Same.

Specs/Number/Hindi/NumberRangeModel.json
```diff
@@ -527,8 +498,7 @@
   },
   {
     "Input": "x 10 से बड़ी और 20 से छोटी है. y 50 से अधिक नहीं है और 20 से कम नहीं है.",
-    "NotSupported": "javascript, python",
-    "NotSupported": "dotnet",
+    "NotSupported": "javascript, python, dotnet",
```
tellarin, line 501: Cases like this one are the ones that should have "Comment" to document the reason for the temporary lack of support.
Also, javascript and python are already listed in NotSupportedByDesign, so they should not be tagged as NotSupported too.

Specs/Number/Hindi/NumberRangeModel.json
```diff
@@ -949,44 +896,42 @@
     ]
   },
   {
-    "Input": "5 में 2 या ज़्यादा कैसा रहेगा",
+    "Input": "हरेक 5 में से 2 या ज़्यादा",
+    "Comment": "Retranslated due to incorrect translation.",
```
tellarin, line 900: Same and so on.

tellarin:
> We left comments directly in the json file, is it ok?

@LionbridgeCSII, comments in the spec should be used for description of the speced behaviour or the reasons why support is partial or not implemented (temporarily or by design). 
Comments such as the reason behind edits in previously existing spec target helping the review of PRs and don't need to exist in the repo later, so they should appear at the PR level and not in files that will go into the repo.
Having said that, it would be valuable to register what was wrong in the previous spec, not just that it was a translation error. But it's understandable that there is not always too much info.

tellarin:
>  ...since the modifier "less than" follows the number...

@LionbridgeCSII, this could be addressed by adding a customization point in the range extractor to change behaviour in Turkish/Hindi, or by adding some language-specific attribute in the YAML file for each language that guides the behaviour of the appropriate code (similar to "CheckBothBeforeAfter"). Could you try to come up with a design for that? 

This can be its own PR after the current one merges. 

Thanks!

.NET/Microsoft.Recognizers.Text.Number/Hindi/Extractors/NumberRangeExtractor.cs
LionbridgeCSII: will investigate this

LionbridgeCSII: will check this

Specs/Number/Hindi/NumberRangeModel.json
LionbridgeCSII: ok, removed comments

Specs/Number/Hindi/NumberRangeModel.json
LionbridgeCSII: removed

Specs/Number/Hindi/NumberRangeModel.json
LionbridgeCSII: added comments, removed "NotSupported": "javascript, python",

Specs/Number/Hindi/NumberRangeModel.json
LionbridgeCSII: removed

LionbridgeCSII:
@tellarin, added comments about modified cases to the PR description (and removed from spec)

Issue #2112, opened 2020-05-16
[EN Number] Add support for Indian numbering system (lakh, crore,...)


LionbridgeCSII:
Indian numbering system is not currently supported in English, so the following examples are not correctly handled by the Text Recognizers:

"There are 1 crore boxes in this container"
"The submarine mein weigh 25 lakh tonnes"
"Am I eligible for a lone of 35 lakh rupees"

**Expected behavior**
numbers containing "lakh", "crore",... should be correctly extracted and parsed in English.

tellarin:
I'm not very familiar with usage in written form, but would people also write in the digit form like "5,00,000"? If so, the implementation of this functionality should also cover support for such cases.

PR #2115, opened 2020-05-17
[EN Number] Add support for Indian numbering system entities (#2112)

LionbridgeCSII:
fix to issue #2112 

Added support for lakh and crore in English Number (and relevant test cases in Number and NumberWithUnit).
There are other larger numbers in the Indian numbering system (arab, kharab, nil,...), but we did not implement them for the following reasons: 
- "these are not commonly used and are unfamiliar to most speakers" (wikipedia)
- they could give ambiguity issues (arab, nil)

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Recognizers.Definitions.English
       public const string LangMarker = @"Eng";
       public const bool CompoundNumberLanguage = false;
       public const bool MultiDecimalSeparatorCulture = true;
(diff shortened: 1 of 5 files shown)


LionbridgeCSII:
@tellarin, 
- removed "NotSupportedByDesign": "javascript,python,java" from new test cases (a part for "I count 6 crore and two hundred", since all cases like "3 hundred and one" are also unsupported)

- Added test case for "lakh crore"

Note that while "three lakh crore" works correctly, "3 lakh crore" only returns "3 lakh".
The same happens with the other RoundNumbers, 
e.g. "three hundred thousand" works fine, but "3 hundred thousand" returns "3 hundred".

tellarin:
> ...3 hundred thousand...

Can you take a look at potentially fixing this in a new PR (estimate the effort/impact)?
Thanks.

PR #2124, opened 2020-05-20
[KO Number] Integer Extractor and Spec Cases

neudurgeshp:
**PR Includes:**
- Integer Extractor
- Number Definitions
- Spec (Cases 234)
   - Existing: 108
   - New: 126
     **New Cases Details**
     Integer (47): Passed-36 Skipped-11
     Negative (10): Passed-4 Skipped-6

       **Pending Implementation**
     Decimal: 9
     Power: 2
     Fraction: 58


Specs/Number/Korean/NumberModel.json
```diff
@@ -1507,5 +1615,2287 @@
...
+        },
+        "Start": 0,
+        "End": 3
+      }
+    ]
+  },
+  {
+    "Input": "322백은 유효하지 않은 수이다",
+    "IgnoreResolution": true,
+    "NotSupportedByDesign": "java, javascript, python, dotnet",
```
tellarin, line 2063: Why are some cases marked as NotSupportedByDesign for .NET?
Cases in .NET should be supported to the extent possible at least at the extraction level. If only temporarily not passing, please add to NotSupported.
In either case, any NotSupported/NotSupportedByDesign should have a Comment with the reason, to help reviews and future implementation planning.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -112,7 +112,7 @@ public override ParseResult Parse(ExtractResult extResult)
             }
 
             // Japanese determine the subtype
-            if (Config.CultureInfo.Name == "ja-JP" && ret != null)
+            if ((Config.CultureInfo.Name == "ja-JP" || Config.CultureInfo.Name == "ko-KR") && ret != null)
```
tellarin, line 115: Please add a `@TODO` comment here to refactor this check. We'll plan for it to happen soon.

neudurgeshp: Sure

Specs/Number/Korean/NumberModel.json
neudurgeshp: Sure. Going to update those cases

Specs/Number/Korean/NumberModel.json
```diff
@@ -1507,5 +1615,2459 @@
...
+        "End": 12
+      }
+    ]
+  },
+  {
+    "Input": "십만 달러",
+    "IgnoreResolution": true,
+    "NotSupportedByDesign": "java, javascript, python",
+    "NotSupported": "dotnet",
+    "Comment": "PendingValidation",
```
tellarin, line 1756: @neudurgeshp, why are these new cases now marked as pending validation? Aren't they exactly the newly created cases?
If not supported yet in the current PR, please come on the PR description for the reasons, ok? But we shouldn't mix status definitions of spec cases.

tellarin: Please update the PR description with a breakdown of the new cases and how many pass. Thanks.

neudurgeshp: Sure, FYI these cases were added by mistake and  need more research by Language expert hence marking them not supported for now and will be updating appropriate comment rather than "**PendingValidation**".

PR #2126, opened 2020-05-21
[.NET] Hindi/Turkish Number NumberRange fix

LionbridgeCSII:
Fix for Hindi (#2113) and Turkish (#1947) unsupported cases in NumberRangeModel.
The solution turned out to be simpler than expected.


LionbridgeCSII:
@tellarin, looking into it. 
It is weird because the test passes when running the test, but it fails when debugging or using the build script

PR #2129, opened 2020-05-22
[KO Number] Number Extractor For Coverage Improvement

neudurgeshp:
**PR Includes**

- Number Extractor
- Number Definitions
- Spec (Cases 234)
- Existing: 108
- New: 126

   **New Cases Details**
- Integer (47): Passed-40, Skipped-7
- Negative (10): Passed-4, Skipped-6
- Decimal (9): Passed-9
- Power (2): Passed-2
- Fraction (58): Passed-32, Skipped-26


Specs/Number/Korean/NumberModel.json
```diff
@@ -2663,7 +2655,7 @@
     "IgnoreResolution": true,
     "NotSupportedByDesign": "java, javascript, python",
     "NotSupported": "dotnet",
-    "Comment": "PendingImplementation",
+    "Comment": "Point value parser need to be refactored",
```
tellarin, line 2658: Why? And does it current fail extraction?

Specs/Number/Korean/NumberModel.json
```diff
@@ -2911,7 +2883,7 @@
     "IgnoreResolution": true,
     "NotSupportedByDesign": "java, javascript, python",
     "NotSupported": "dotnet",
-    "Comment": "PendingImplementation",
+    "Comment": "As per language expert, input needs to be corrected",
```
tellarin, line 2886: Can you explain why in the PR description?

Specs/Number/Korean/NumberModel.json
```diff
@@ -3044,7 +3008,7 @@
...
```
tellarin, line 3011: Same, and so on.

Specs/Number/Korean/NumberModel.json
neudurgeshp: There is no issue in extraction but during parsing it throws exception for value 2.12**백** from **GetPointValue** method, where we may need to add RoundNumberMap support as well. 
Here is the line..
**pointValue += scale * Config.**ZeroToNineMap**[c];**

**백 ->100**  is round number that is not part of ZeroToNineMap dictionary.

Specs/Number/Korean/NumberModel.json
neudurgeshp: Sure.

Specs/Number/Korean/NumberModel.json
neudurgeshp: Sure.

PR #2130, opened 2020-05-22
[EN Number] Fix to issue pointed out in #2115

LionbridgeCSII:
Added support for "3 hundred thousand".
The fix does not seems to affect other entities.
Added test cases to NumberModel.


tellarin (changes requested):
This looks reasonable, but there will probably be a significant performance impact.
Marking this as pending changes until we can collect some perf metrics.

PR #2133, opened 2020-05-23
Fix code sample in the .NET platform page

lindexi:


lindexi:
@tellarin Thank you

tellarin:
@lindexi, thank you for spotting the error! 👍

PR #2134, opened 2020-05-23
[.NET] Hindi DateTime Time enhancements 

LionbridgeCSII:
Support for additional cases in Time (7 in TimeExtractor, 6 in TimeParser).
All new test cases pass.
One old case skipped (due to different sentence structure as pointed out in  #1983)

### DIFF
--- a/Specs/DateTime/Hindi/TimeParser.json
+++ b/Specs/DateTime/Hindi/TimeParser.json
@@ -1159,7 +1159,7 @@
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
-        "Text": "दोपहर",
+        "Text": "दोपहर में",
         "Type": "time",
         "Value": {
           "Timex": "T12",
@@ -1171,7 +1171,7 @@
           }
         },
         "Start": 0,
-        "Length": 5
+        "Length": 9
       }
     ]
   },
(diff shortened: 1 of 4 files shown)


Specs/DateTime/Hindi/TimeParser.json
```diff
@@ -1616,5 +1616,134 @@
         "Length": 28
       }
     ]
+  },
+  {
+    "Input": "मैं 7 बजकर 1 मिनट पर वापस आ जाउंगा",
+    "NotSupported": "python",
```
tellarin, line 1622: Please be careful to not add unnecessary attributes. Python is already listed as NotSupportedByDesign

LionbridgeCSII: removed

LionbridgeCSII:
> Which previous spec was disabled? I didn't see any Time case disabled for dotnet here. Did I miss it?

This one "मैं साढ़े पाँच बजने में तीन मिनट बाकी होने पर वापस जाउंगा" that corresponds to "I'll go back three mins to five thirty" but in Hindi has a different structure (half past five three mins to).

tellarin:
> > Which previous spec was disabled? I didn't see any Time case disabled for dotnet here. Did I miss it?
> 
> This one "मैं साढ़े पाँच बजने में तीन मिनट बाकी होने पर वापस जाउंगा" that corresponds to "I'll go back three mins to five thirty" but in Hindi has a different structure (half past five three mins to).

I don't see this case being disabled in this PR. If it was already previously disabled, that's OK.
Can you add a Comment on the spec case describing why not support for documentation (if not already commented)?

Issue #2068, opened 2020-04-22
[* PhoneNumber] (914)349-8600 not recognized as PhoneNumber


sunnysharma03:
**Describe the bug**
PhoneNumber Prebuilt Entity is not able to recognize PhoneNumbers correctly from utterances.

Examples:

**OFFICE-1(516)733-3989**
**Purchasing Phone (914)349-8600 04/21/2020 12:37:30**
--

**Expected behavior**

from OFFICE-1(516)733-3989 -> 516-733-3989 should be extracted as phoneNumber
from Purchasing Phone (914)349-8600 04/21/2020 12:37:30 -> 914-349-8600 should be extracted as phoneNumber

**Additional context**

PR #2136, opened 2020-05-25
Fix issues with phone number not being extracted (#2068)

visionshao:
Fix the issue #2068 
Change the regexes of NLPhoneNumber and ITPhoneNumber;
Add codes for checking if extracted string has a non-digit string before "-" ;
Add Specs to Specs/Sequence/English/PhoneNumberModel.json;


.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BasePhoneNumberExtractor.cs
```diff
@@ -130,8 +130,11 @@ public override List<ExtractResult> Extract(string text)
                             var charGap = text[(int)(er.Start - 2)];
                             if (!char.IsNumber(charGap) && !char.IsWhiteSpace(charGap))
                             {
+                                // check if the extracted string has a non-digit string before "-".
+                                var flag = Regex.IsMatch(text.Substring(0, (int)(er.Start - 2)), @"^[^0-9]+$");
```
tellarin, line 134: Checking some kind of prefix for the tagged potential phonenumber only if fortmatted? Is that it?

Specs/Sequence/English/PhoneNumberModel.json
```diff
@@ -4776,5 +4776,42 @@
   {
     "Input": "My card number is 12345678",
     "Results": []
+  },
+  {
+    "Input": "OFFICE-1(516)733-3989",
```
tellarin, line 4781: Can you add a few more cases to show that your changes handle more general cases?
Ex: "OFFICE - 1(516)733-3989" or "Office:(516)733-3989"

.NET/Microsoft.Recognizers.Text.Sequence/Extractors/BasePhoneNumberExtractor.cs
visionshao: yes

Specs/Sequence/English/PhoneNumberModel.json
visionshao: ok

PR #2138, opened 2020-05-25
[.NET] Hindi DateTime DateTime enhancements

LionbridgeCSII:
Support for additional cases in DateTime (20 in Extractor and 8 in Parser).
All new cases pass.
No changes in DataTimeModel.

### DIFF
--- a/Specs/DateTime/Hindi/DateTimeExtractor.json
+++ b/Specs/DateTime/Hindi/DateTimeExtractor.json
@@ -562,5 +562,246 @@
         "Length": 21
       }
     ]
+  },
+  {
+    "Input": "मैं कल सुबह 9 बजे वापस जाऊंगा",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "कल सुबह 9 बजे",
+        "Type": "datetime",
+        "Start": 4,
+        "Length": 13
+      }
+    ]
+  },
+  {
+    "Input": "मैं कल सुबह 9 बजे वापस जाऊँगा",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "कल सुबह 9 बजे",
+        "Type": "datetime",
+        "Start": 4,
+        "Length": 13
+      }
+    ]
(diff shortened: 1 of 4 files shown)


Specs/DateTime/Hindi/DateTimeExtractor.json
```diff
@@ -562,5 +562,246 @@
...
+        "Text": "इस शुक्रवार को दोपहर एक बजे",
+        "Type": "datetime",
+        "Start": 0,
+        "Length": 27
+      }
+    ]
+  },
+  {
+    "Input": "दोपहर 12:30 बजे शुक्रवार के लिए लंच जोड़ें",
+    "Comment": "Disable this for now because of new features in .NET",
```
tellarin, line 616: This is not disabled. Does this comment apply? And which feature was it?

LionbridgeCSII: this comes from the English spec (where the test case is not disabled as well). 
I will remove both.

Specs/DateTime/Hindi/TimePeriodExtractor.json
```diff
@@ -539,18 +539,6 @@
       }
     ]
   },
-  {
-    "Input": "शाम 1:30 से 3:30 तक मुलाकात रखो ",
```
tellarin, line 543: Why has this case been removed?

LionbridgeCSII: it was duplicate

tellarin: I don't see another case for "शाम 1:30 से 3:30 तक मुलाकात रखो" in this spec.
There is a case for "1:30 से 3:30 तक मुलाकात रखो", but that's not the same. In my understanding one of them is specific because of  शाम while the other could be interpreted as two possible time periods. 
I suggest keeping both in this case.

LionbridgeCSII: but it seems after rebasing it has been removed again, I will reintroduce it.

Issue #1792, opened 2019-08-09
[CN Number] Corner cases for formal numbers


wgx998877:
some corner case in chinese for formal numbers:
“收拾”， “朝花夕拾”，“队伍”, “放肆”, “大陆”, "海陆" ...

PR #2141, opened 2020-05-26
[CN Number] Filter false positives in Chinese Number (#1792)

visionshao:
Fix the issue #1792
Add some corner cases of Chinese Number to ambiguity filter;
Add test cases to Chinese NumberModel.json

### DIFF
--- a/Specs/Number/Chinese/NumberModel.json
+++ b/Specs/Number/Chinese/NumberModel.json
@@ -4788,5 +4788,65 @@
     "Input": "这是我的队伍",
     "NotSupported": "python, javascript, java",
     "Results": []
+  },
+  {
+    "Input": "小明每天收拾房间",
+    "NotSupported": "python, javascript, java",
+    "Results": []
+  },
+  {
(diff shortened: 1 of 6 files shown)


Specs/Number/Chinese/NumberModel.json
```diff
@@ -4788,5 +4788,65 @@
     "Input": "这是我的队伍",
     "NotSupported": "python, javascript, java",
     "Results": []
+  },
+  {
+    "Input": "小明每天收拾房间",
```
tellarin, line 4793: Aren't these supported in the other platforms once the resources are generated? If so, please include the updates resource files and enable the tests, ok?
Thanks.

tellarin, line 4793: Aren't these supported in the other platforms once the resources are generated? If so, please include the updates resource files and enable the tests, ok?
Thanks.

visionshao: ok

Issue #1789, opened 2019-08-08
[EN DateTimeV2] Support "night ranges"


tellarin:
Support for "night ranges", like typically used in hotel scenarios:
- "Book a room for two nights"

Issue linked to #1717.

tellarin:
@LionbridgeCSII, can you take a look at this one? Resolution would be same as similar cases as "two days".

PR #2142, opened 2020-05-26
[EN DateTimeV2] Support for ranges with "nights" (#1789)

LionbridgeCSII:
Fix to issue #1789.
Added relevant test cases in DateTimeModel.
The result of the existing test case "Can I do a booking for the 09th of May for 2 nights?" has been modified to match the one from "Can I do a booking for the 09th of May for 2 days?"

### DIFF
--- a/Patterns/English/English-DateTime.yaml
+++ b/Patterns/English/English-DateTime.yaml
@@ -346,7 +346,7 @@ IshRegex: !nestedRegex
   def: '\b({BaseDateTime.HourRegex}(-|——)?ish|noon(ish)?)\b'
   references: [ BaseDateTime.HourRegex ]
 TimeUnitRegex: !simpleRegex
-  def: ([^A-Za-z]{1,}|\b)(?<unit>h(ou)?rs?|h|min(ute)?s?|sec(ond)?s?)\b
+  def: ([^A-Za-z]{1,}|\b)(?<unit>h(ou)?rs?|h|min(ute)?s?|sec(ond)?s?|nights?)\b
 RestrictedTimeUnitRegex: !simpleRegex
   def: (?<unit>hour|minute)\b
 FivesRegex: !simpleRegex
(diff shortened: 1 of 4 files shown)


Patterns/English/English-DateTime.yaml
```diff
@@ -346,7 +346,7 @@ IshRegex: !nestedRegex
   def: '\b({BaseDateTime.HourRegex}(-|——)?ish|noon(ish)?)\b'
   references: [ BaseDateTime.HourRegex ]
 TimeUnitRegex: !simpleRegex
-  def: ([^A-Za-z]{1,}|\b)(?<unit>h(ou)?rs?|h|min(ute)?s?|sec(ond)?s?)\b
+  def: ([^A-Za-z]{1,}|\b)(?<unit>h(ou)?rs?|h|min(ute)?s?|sec(ond)?s?|nights?)\b
```
tellarin, line 349: Why is this change necessary? "nights" don't specify time.

Patterns/English/English-DateTime.yaml
```diff
@@ -477,7 +477,7 @@ LessThanRegex: !simpleRegex
 MoreThanRegex: !simpleRegex
   def: \b(more\s+than)\b
 DurationUnitRegex: !nestedRegex
-  def: (?<unit>{DateUnitRegex}|h(ou)?rs?|h|min(ute)?s?|sec(ond)?s?)\b
+  def: (?<unit>{DateUnitRegex}|h(ou)?rs?|h|min(ute)?s?|sec(ond)?s?|nights?)\b
```
tellarin, line 480: Why not make this change in DateUnitRegex? It's conceptually right and where similar cases for "days" are captured.

LionbridgeCSII: Because then expressions like "last night" will be resolved as "daterange" instead of "datetimerange"

Patterns/English/English-DateTime.yaml
LionbridgeCSII: yes right, it is not necessary. Removed

Patterns/English/English-DateTime.yaml
tellarin: Seems reasonable. But ideally the term should then be added closes to {DateUnitRegex} in the regex. Not a merge blocker though.
Thanks.

PR #2144, opened 2020-05-26
[EN Number] Add support for Indian numbering system in digits (#2112)

LionbridgeCSII:
Support for Indian numbering system in digits as requested in #2112.
Added test cases to NumberModel and CurrencyModel.

The digit notation "5,00,000" is not as common as the written form "5 lakh", but it is a government approved numbering system in India and is used, for example, in Indian transactions.

### DIFF
--- a/Specs/Number/English/NumberModel.json
+++ b/Specs/Number/English/NumberModel.json
@@ -2700,5 +2700,37 @@
         "End": 3
       }
     ]
+  },
+  {
+    "Input": "12,34,567.89",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "12,34,567.89",
+        "TypeName": "number",
+        "Resolution": {
+          "value": "1234567.89",
+          "subtype": "decimal"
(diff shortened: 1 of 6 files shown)


Specs/Number/English/NumberModel.json
```diff
@@ -2700,5 +2700,37 @@
         "End": 3
       }
     ]
+  },
+  {
+    "Input": "12,34,567.89",
```
tellarin, line 2705: Can you add a proper sentence, so there's a little more context for the spec case?
Also, please add a "Comment" in the spec to document that this is because of the Indian numbering system and the format may not need to be supported in other languages.

LionbridgeCSII:
> @LionbridgeCSII, does the current Hindi implementation already support this number format too? If not, please create a PR for it as well.

yes it does, the new regex comes from the Hindi yaml file.

Specs/Number/English/NumberModel.json
LionbridgeCSII: done

PR #2146, opened 2020-05-28
[KO NumberRange] Extractor implementation and adding test specs

neudurgeshp:
**PR Includes**
- Number-Range Extractor
- Number Definitions
- Test Spec: Cases 122 (Extractor Passed - 71, Full Passed - 54, Skipped - 51)

**Existing Cases: 53**
- Extractor Passed: 23
- Full Passed: 19
- Skipped: 30

**New Cases: 69**
- Extractor Passed: 48
- Full Passed: 35
- Skipped: 21

### DIFF
--- a/Patterns/Korean/Korean-Numbers.yaml
+++ b/Patterns/Korean/Korean-Numbers.yaml
@@ -304,7 +304,7 @@ SimpleIntegerPercentageRegex: !nestedRegex
 TillRegex: !simpleRegex
   def: (부터|에서|--|-|—|–|——|~)
 MoreRegex: !simpleRegex
-  def: (초과|많|높|크|더많|더높|더크|넘는|초과이다|크고)
+  def: (초과|많|높|크|더많|더높|더크|>|넘는|초과이다|크고)
 LessRegex: !simpleRegex
   def: (미만|적|낮|작|더적|더낮|더적|이하|이하이다|<|아래|작다)
 EqualRegex: !simpleRegex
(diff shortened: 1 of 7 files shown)


Patterns/Korean/Korean-Numbers.yaml
```diff
@@ -302,23 +302,69 @@ SimpleIntegerPercentageRegex: !nestedRegex
   references: [NegativeNumberTermsRegexNum, ZeroToNineFullHalfRegex, AllIntRegex, RoundNumberIntegerRegex, NegativeNumberTermsRegex]
 #NumberRangeExtractor
 TillRegex: !simpleRegex
-  def: (부터|까지|--|-|—|——|~)
+  def: (부터|에서|--|-|—|–|——|~)
 MoreRegex: !simpleRegex
-  def: (초과|많|높|크|더많|더높|더크|>)
+  def: (초과|많|높|크|더많|더높|더크|넘는|초과이다|크고)
```
tellarin, line 307: Shouldn't ">" be kept in the regex?
"<" is kept in the corresponding regex for less.

Patterns/Korean/Korean-Numbers.yaml
```diff
@@ -302,23 +302,69 @@ SimpleIntegerPercentageRegex: !nestedRegex
...
   def: \s*(이상)
 LessOrEqual: !nestedRegex
-  def: (({LessRegex}\s*(거나)?\s*{EqualRegex}))
+  def: (?:(이|보다)?)?\s*(({LessRegex}\s*(거나)?\s*{EqualRegex}(은|다)?))
   references: [ LessRegex, EqualRegex ]
 LessOrEqualSuffix: !simpleRegex
-  def: \s*(이상)
+  def: \s*(이하)
+OneNumberRangeMoreSeparateRegex: !simpleRegex
+  # TODO: modify below regex according to the counterpart in English
```
tellarin, line 327: If these first two are not necessary, change the TODO comment to "# Regex not applicable in Korean".
If they may be written later, you can leave as is for now.

neudurgeshp: Sure, we'll update the TODO comment.

Patterns/Korean/Korean-Numbers.yaml
neudurgeshp: It should be. We'll put '>' back to the MoreRegex.

tellarin: Please also add specs that use those symbols.

neudurgeshp: For '>' and '<' symbols, cases are there in Spec. 
For example.
- x>10 그리고 y<20
- 아직도 >30인 경우인가요
If something additional you are expecting, please let us know.

PR #2152, opened 2020-06-01
[.NET] Hindi DateTime Holiday enhancements

LionbridgeCSII:
Support for additional cases in Holiday (6 in Extractor and 8 in Parser).
All new cases pass.

Added 2 holidays (RakshaBandhan and Vaishakhi) to the HoliDiwali dictionary (now HoliDiwaliRakshabandhanBaisakhiDates).

We investigated the possibility to calculate these dates, but it seems that even using a lunar calendar it is difficult to get reliable results for more than a few years in the future. 
Also, after inspecting the codebase, we found references to the lunar calendar only in Chinese, Japanese and Korean, but it seems that the parsed result simply gets a flag 'isLunar' if a LunarRegex is matched. Are we missing something? e.g. a method that converts lunar date to solar date (or that returns dates of new moons).

No changes in DateTimeModel.


tellarin:
@LionbridgeCSII Yeah, it seems the isLunar property is not being enforced in parser/resolution output. 
It should be enforced across languages for any lunar-based date resolution though.
Can you create an issue for it? Thanks.

PR #2159, opened 2020-06-04
[.NET] Dutch DateTime Duration support

LionbridgeCSII:
All test cases pass.

Added variations of test cases with abbreviations 'dgn' (dagen) and 'jr' (jaar).

Two cases were updated in DateParser because they used "couple of" instead of "few" which has a different resolution.

In DateTimeModel, 125 pass, 217 fail.

### DIFF
--- a/Patterns/Dutch/Dutch-Numbers.yaml
+++ b/Patterns/Dutch/Dutch-Numbers.yaml
@@ -80,15 +80,15 @@ FractionNotationWithSpacesRegex: !simpleRegex
 FractionNotationRegex: !simpleRegex         
   def: (((?<=\W|^)-\s*)|(?<![/-])(?<=\b))\d+[/]\d+(?=(\b[^/]|$))
 FractionUnitsRegex: !simpleRegex
-  def: ((?<onehalf>anderhalve|anderhalf)|(?<quarter>driekwart)|half|halve|helft|kwart)
+  def: ((?<onehalf>anderhalve|anderhalf)|(?<quarter>driekwart)|half|halve|helft|kwart|kwartier)
 FractionHalfRegex: !simpleRegex
-  def: (ënhalf|enhalve|ëneenhalf)$
+  def: (ënhalf|enhalve|ëneenhalf|ëneenhalf)$
 OneHalfTokens: [een, half]
 FractionNounRegex: !nestedRegex
   def: (?<=\b)(({AllIntRegex}\s+(en\s+)?)?(({AllIntRegex})(\s+|\s*-\s*|\s*/\s*)((({AllOrdinalRegex})|({RoundNumberOrdinalRegex}))n?|halven|vierdes|kwart)|{FractionUnitsRegex}))(?=\b)
   references: [ AllIntRegex, AllOrdinalRegex, RoundNumberOrdinalRegex, FractionUnitsRegex ]
 FractionNounWithArticleRegex: !nestedRegex
-  def: (?<=\b)(({AllIntRegex}\s+(en\s)?)?(een)(\s+|\s*-\s*|\s*/\s*)(({AllOrdinalRegex})|({RoundNumberOrdinalRegex})|({FractionUnitsRegex}))|{AllIntRegex}[eë]n(half|halve|helft|kwart))(?=\b)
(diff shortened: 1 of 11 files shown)


tellarin:
@iMicknl, can you take a look this please?

Patterns/Dutch/Dutch-Numbers.yaml
```diff
@@ -80,15 +80,15 @@ FractionNotationWithSpacesRegex: !simpleRegex
 FractionNotationRegex: !simpleRegex         
   def: (((?<=\W|^)-\s*)|(?<![/-])(?<=\b))\d+[/]\d+(?=(\b[^/]|$))
 FractionUnitsRegex: !simpleRegex
-  def: ((?<onehalf>anderhalve|anderhalf)|(?<quarter>driekwart)|half|halve|helft|kwart)
+  def: ((?<onehalf>anderhalve|anderhalf)|(?<quarter>driekwart)|half|halve|helft|kwart|kwartier)
```
iMicknl, line 83: Will this not cause issues with the NumberParser where it will parse kwartier as 0,25? Kwartier should only be used in a DateTime context. Kwart is 0,25 in a number context.

Specs/DateTime/Dutch/DateParser.json
```diff
@@ -1243,19 +1243,19 @@
...
         },
         "Start": 11,
-        "Length": 24
+        "Length": 22
       }
     ]
   },
   {
-    "Input": "wie heb ik een paar dagen geleden een email gestuurd",
+    "Input": "wie heb ik enkele dagen geleden een email gestuurd",
```
iMicknl, line 1251: `een paar dagen geleden` is also a valid Dutch sentence and frequently used. Is this one covered by another spec already?

Specs/DateTime/Dutch/DurationParser.json
```diff
@@ -595,43 +573,41 @@
...
         "Start": 7,
         "Length": 14
       }
-    ],
-    "NotSupported": "dotNet",
-    "NotSupportedByDesign": "javascript,python,java"
+    ]
   },
   {
     "Input": "Ik ben een kwartier uur weg",
```
iMicknl, line 603: This spec was incorrect, sorry. It could be `een kwartier` or `een kwart uur`.

Specs/DateTime/Dutch/DateParser.json
LionbridgeCSII: the Input was updated because the original English case uses "few" that has a different resolution than "a couple of". There are other test cases where "een paar" is used.

Patterns/Dutch/Dutch-Numbers.yaml
LionbridgeCSII: removed 'kwartier' from Numbers.
This required a small change in BaseDurationParser in order to parse 'een kwartier' correctly as '0.25' instead of '1.25'.

Specs/DateTime/Dutch/DurationParser.json
LionbridgeCSII: updated the test case to "Ik ben een kwartier weg"

Specs/DateTime/Dutch/DurationParser.json
```diff
@@ -600,14 +600,14 @@
     ]
   },
   {
-    "Input": "Ik ben een kwartier uur weg",
+    "Input": "Ik ben een kwartier weg",
```
tellarin, line 603: Do we have a spec somewhere for the other form Mick mentioned? "Ik ben een kwart uur"

LionbridgeCSII: added

Specs/DateTime/Dutch/DateParser.json
tellarin: But do we have a separate case with "een paar"? We should at least keep it and mark temporarily as NotSupported if not yet passing. We should never remove spec cases that are plausible in a target language.

LionbridgeCSII: in DurationParser there are cases with "en paar" (e.g. "Ik ben een paar dagen weg.").
There are no such cases in DateParser (neither in Dutch nor in English)

PR #2161, opened 2020-06-05
[JA Number] Coverage Improvement (Number and Percent Model) and Spec Updates

neudurgeshp:
**PR Includes**
- Number Definitions
- Spec
   - NumberModel
   - PercentModel

- Number Extractor
   - Total Cases 484: (Extractor Passed - 471, Full Passed - 386, Skipped - 13)
   - New Test Spec: Cases 386 (Extractor Passed - 370, Full Passed - 288, Skipped - 16)
   - Integer 154: Extraction Passed-151, Skipped-3
   - Fraction 149: Extraction Passed-149
   - Decimal 30: Extraction Passed-27, Skipped-3
   - Power 30: Extraction Passed-30
   - Negative 23 : Extraction Passed-13, Skipped-10

- Percentage Extractor
   - Total Test Cases 93: (Extractor Passed - 93, Full Passed - 93, Skipped - 93)
   - Existing Cases 53: Extractor Passed - 53, Full Passed - 53
   - New Cases 40: Extractor Passed - 40, Full Passed - 40


.NET/Microsoft.Recognizers.Definitions.Common/Japanese/NumbersDefinitions.cs
```diff
@@ -113,6 +113,7 @@ public static class NumbersDefinitions
         {
             '十'
         };
+      public const string UnitsRegex = @"(k|mil|t|g|b|km|ml|m|ミリリットル(入れら)?|キロメートル|メートル|ミリメート)";
```
tellarin, line 116: What do these represent? All multipliers? If so, the regex should be named NumberMultiplierRegex as in other languages and reference the one in BaseUnit if re-using all values.

Specs/Number/Japanese/NumberModel.json
```diff
@@ -3931,7 +3932,9 @@
   {
     "Input": "この袋の厚さは0.08ミリメートルです。",
     "IgnoreResolution": true,
+    "Comment": "Need language review",
     "NotSupportedByDesign": "javascript, java, python",
+    "NotSupported": "dotnet",
```
tellarin, line 3937: It was passing previously and broke in this PR?

neudurgeshp: After review we realized that this should be negative test cases. That's why it's marked as notsupported for now. We'll be reviewing again with language expert.

.NET/Microsoft.Recognizers.Definitions.Common/Japanese/NumbersDefinitions.cs
neudurgeshp: Agreed. We'll do that.

PR #2162, opened 2020-06-08
[.NET] Dutch DateTime DateTime support

LionbridgeCSII:
All test cases pass.

Some Inputs such as "Ik ga morgen terug om 8:00" ("I'll go back tomorrow 8:00am") were updated because 'terug' (back) separates the date and time entities.
Here the list of updated cases:
- "Ik ga morgen terug om 8:00" -> "Ik ga morgen om 8:00 terug"
- "Ik ga terug om 20.15u" -> "Ik ga terug 20.00 op 15"
- "Ik ga vanavond terug om 7" -> "Ik ga vanavond om 7 terug"
- "Ik ga op de 5e terug om 4u ‘s morgens" -> "Ik ga terug op de 5e om 4u ‘s morgens"
- "Ik ga op 16-12-2016 terug om 12.23:59u" -> "Ik ga terug op 16-12-2016 om 12.23:59u"
- "Ik ga de 15e, 20:00 terug" -> "Ik ga op de 15e, 20:00 terug"
- "Ik ga 20.00 terug op de 15e" -> "Ik ga terug op de 15e om 20.00"
- "Ik ga om zeven uur terug op de 15e" -> "Ik ga terug om zeven uur op de 15e"

In DateTimeModel, 130 pass, 212 fail.


Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -314,9 +314,9 @@ RangeUnitRegex: !simpleRegex
...
+  def: (?<minnum>nul|een|één|twee|drie|vier|vijf|zes|zeven|acht|negen|tien|elf|elven|twaalf|dertien|veertien|vijftien|zestien|zeventien|achttien|negentien|twintig|eenentwintig|éénentwintig|tweeentwintig|tweeëntwintig|drieëntwintig|vierentwintig|vijfentwintig|vijventwintig|zesentwintig|zevenentwintig|achtentwintig|negenentwintig|dertig|eenendertig|tweeëndertig|drieëndertig|vierendertig|vijfendertig|zesendertig|zevenendertig|achtendertig|negenendertig|veertig|eenenveertig|tweeënveertig|drieënveertig|vierenveertig|vijfenveertig|zesenveertig|zevenenveertig|achtenveertig|negenenveertig|eenenvijftig|vijftig|tweeënvijftig|drieënvijftig|vierenvijftig|vijfenvijftig|zesenvijftig|zevenenvijftig|achtenvijftig|negenenvijftig)
```
tellarin, line 317: Minor: Instead of full repetitions like "vijfentwintig|vijventwintig|zesentwintig|zevenentwintig", you can write "vij[fv]entwintig|ze[sv]entwintig", which helps simplify review and debugging.

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -463,15 +463,15 @@ TimeOfTodayBeforeRegex: !nestedRegex
...
+  def: ({HourNumRegex}|{BaseDateTime.HourRegex}(\s*:\s*{BaseDateTime.MinuteRegex})?)(\s*({OclockRegex}|u))?\s*(,\s*)?(in\s+de\s+)?{DateTimeSpecificTimeOfDayRegex}
+  references: [ HourNumRegex, BaseDateTime.HourRegex, BaseDateTime.MinuteRegex, DateTimeSpecificTimeOfDayRegex, OclockRegex ]
 SimpleTimeOfTodayBeforeRegex: !nestedRegex
   def: '\b{DateTimeSpecificTimeOfDayRegex}(\s*,)?(\s+(om|rond|tegen|op))?\s*({HourNumRegex}|{BaseDateTime.HourRegex})\b'
   references: [ DateTimeSpecificTimeOfDayRegex, HourNumRegex, BaseDateTime.HourRegex ]
 SpecificEndOfRegex: !simpleRegex
-  def: (het\s+)?einde van(\s+de)?\s*$
+  def: ((het\s+)?einde? van(\s+de)?\s*$|^\s*(het\s+)?einde? van(\s+de(\s+dag)))
 UnspecificEndOfRegex: !simpleRegex
-  def: \b(((om|rond|tegen|op)\s+)?het\s+)?(einde\s+van\s+de\s+dag)\b
```
tellarin, line 474: "einde van de dag" is not possible? @iMicknl 
Perhaps a "(de\s+)?" would be enough here to cover both.

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -719,11 +730,15 @@ UnitValueMap: !dictionary
     minuten: 60
     minuut: 60
     min: 60
+    min.: 60
     secondes: 1
     seconden: 1
     seconde: 1
     secs: 1
     sec: 1
+    kwartier: 3600
```
tellarin, line 739: This seems incorrect. Especially as there are not the amounts of seconds in a half/quarter hour. Is this really necessary?

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -681,22 +681,30 @@ UnitMap: !dictionary
...
     minuten: M
     minuut: M
     min: M
+    min.: M
     secondes: S
+    seconden: S
     seconde: S
     secs: S
     sec: S
+    kwartier: H
```
tellarin, line 705: Are these also necessary? How does English output the unit for similar cases?

Patterns/Dutch/Dutch-Numbers.yaml
```diff
@@ -82,13 +82,13 @@ FractionNotationRegex: !simpleRegex
 FractionUnitsRegex: !simpleRegex
   def: ((?<onehalf>anderhalve|anderhalf)|(?<quarter>driekwart)|half|halve|helft|kwart)
 FractionHalfRegex: !simpleRegex
-  def: (ënhalf|enhalve|ëneenhalf)$
+  def: (ënhalf|enhalve|ëneenhalf|ëneenhalf)$
```
tellarin, line 85: Duplicate values.

Specs/DateTime/Dutch/DateParser.json
```diff
@@ -1224,14 +1224,14 @@
     ]
   },
   {
-    "Input": "Wie heb ik een paar maanden geleden een email gestuurd",
+    "Input": "Wie heb ik enkele maanden geleden een email gestuurd",
```
tellarin, line 1227: Same as in previous reviews, if the original sentence is plausible, do not remove it when adding a new one with edits.
If the original one is not currently supported, please mark as NotSupported and add a Comment. Edit its resolution accordingly if that's the issue. Or move the original sentence to either its right type or to the model-level json spec (in case it's an issue of language differences that actually represent a different type).
@iMicknl, can you take a look and see if we should keep both?

Specs/DateTime/Dutch/DateTimeExtractor.json
```diff
@@ -35,21 +33,19 @@
         "Start": 15,
         "Length": 10
       }
-    ],
-    "NotSupported": "dotNet"
+    ]
   },
   {
-    "Input": "Ik ga terug op 15 om 8.00:30",
```
tellarin, line 42: Same.

Specs/DateTime/Dutch/DateTimeExtractor.json
```diff
@@ -126,60 +117,55 @@
         "Start": 15,
         "Length": 17
       }
-    ],
-    "NotSupported": "dotNet"
+    ]
   },
   {
-    "Input": "Ik ga morgen terug om 8:00",
```
tellarin, line 133: Same. If both orders are correct, we need both specs. It's totally fine that the new one returns a datetime here, but the original one should be moved to the model-level spec to show that it then returns a date and a time because of the interfering term.

tellarin: Same for the next ones. @iMicknl, sorry for tagging you so much. But I need your help to confirm that both forms would happen in Dutch. :)

Specs/DateTime/Dutch/DateTimeExtractor.json
```diff
@@ -204,34 +189,31 @@
         "Start": 15,
         "Length": 41
       }
-    ],
-    "NotSupported": "dotNet"
+    ]
   },
   {
-    "Input": "Ik ga terug om 20.15u",
```
tellarin, line 211: Same, and so on.

tellarin: Please always add comments at the PR level to describe the reason of spec edits and where did the original went, if they've been moved. Otherwise it's very hard to review spec changes (specially not knowing the language). Thanks.

Specs/DateTime/Dutch/DateTimeExtractor.json
```diff
@@ -490,21 +453,7 @@
         "Start": 6,
         "Length": 19
       }
-    ],
-    "NotSupported": "dotNet"
-  },
-  {
-    "Input": "Ik ga vanavond terug om 7",
```
tellarin, line 497: Why was this deleted? If it was duplicate, it would have broken previous builds, no?

LionbridgeCSII:
> Please be careful with spec changes. As we've discussed before, if cases exist, they should only be altered/removed if they really don't apply. Otherwise we want to cover both the original case and the new form. It may happen that sentences have to be moved to their right specs or to the model level spec, but we should never delete a previously existing spec without a very strong reason.

In this case, in the context of DateTime entity, I would say that they do not apply because they do not contain a DateTime expression but a Date and a Time entity (which are already covered in the corresponding specs). But ok, we will add the original cases to DateTimeModel.

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: done

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: added

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: it is necessary because 'kwartier' contains both the fraction 'quarter' and the unit 'hour', so its value in UnitValueMap is required, for example, by ParseNumberSpaceUnit in BaseDurationParser

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: same as above

Patterns/Dutch/Dutch-Numbers.yaml
LionbridgeCSII: removed

Specs/DateTime/Dutch/DateParser.json
LionbridgeCSII: this is exactly the same case discussed in previous review (this PR has been built on top of the previous one). As discussed there, it was updated because the original English case uses "few" instead of "couple of". Cases showing support for "een paar" appear in other entities (as for English)

Specs/DateTime/Dutch/DateTimeExtractor.json
LionbridgeCSII: reintroduced

Specs/DateTime/Dutch/DateTimeExtractor.json
LionbridgeCSII: added to DateTimeModel

Specs/DateTime/Dutch/DateTimeExtractor.json
LionbridgeCSII: this case was wrongly localized, the original English is "I'll go back 8pm on 15"

Specs/DateTime/Dutch/DateTimeExtractor.json
LionbridgeCSII: this case was modified because 'terug' breaks the DateTime entity. It is now included in DateTimeModel

Patterns/Dutch/Dutch-DateTime.yaml
tellarin: This seems strange, as kwartier would be a quarter of this value. Perhaps the best option would be for the code that consults the table to be aware of this?
I'm fine with keeping it for now, but please take a look to see if we can be more consistent.

Specs/DateTime/Dutch/DateParser.json
tellarin: Yes. I'm just re-stating the goal of keeping coverage as high as possible. Spec cases don't need to be only the localizations of the English cases. 
Any existing spec cases in target languages should be preserved as much as possible, if they are plausible in that target language (if the output type for a NotSupported spec case isn't correct for the current spec file, then they should be moved to the right type or model spec file).

Specs/DateTime/Dutch/DateTimeExtractor.json
tellarin: Same as the  "een paar"/"enkele" issue. If not incorrect, it should not be removed and we should also support it, even if not a direct match to a source spec in English.

Specs/DateTime/Dutch/DateTimeExtractor.json
tellarin: Got it.

tellarin:
> In this case, in the context of DateTime entity, I would say that they do not apply because they do not contain a DateTime expression but a Date and a Time entity (which are already covered in the corresponding specs). But ok, we will add the original cases to DateTimeModel.

I understand your comment. But the main goal for the project is always to guarantee language coverage across types and to accumulate variations between languages, so we shouldn't focus only on the specific PR when regarding data/specs.

Specs/DateTime/Dutch/DateTimeExtractor.json
tellarin: I mistakenly merged the other PR with this removal, when we should have kept both spec cases and not remove one.

LionbridgeCSII: added to DateTimeModel

Specs/DateTime/Dutch/DateParser.json
LionbridgeCSII: added to DateTimeModel

Issue #2036, opened 2020-02-14
[DE DateTimeV2] Enable recognition of dialectal clock expressions


Ablesius:
**Is your feature request related to a problem? Please describe.**
In German it is possible not only to relate to time with prepositional expressions ("Viertel nach"/"Viertel vor" (quarter past, quarter to)) but also without prepositions:

- "Viertel 6" = "05:15"
- "Halb 6" = "05:30" (cp. English "half six")
- "Dreiviertel 6" = 05:45

It would be really nice if the recognition for these expressions could be enabled. Shouldn't be too hard since it's really just these three expressions ("Halb" already works); other expressions aren't possible to resolve to these times.

**Describe the solution you'd like**
see above

**Describe alternatives you've considered**
No alternatives, as the only other solution - not implementing the recognition - is already in place.

**Additional context**
None

tellarin:
@LionbridgeCSII, can you take a look at this one?

Ablesius:
@LionbridgeCSII if you have a PR ready, I might be able to contribute there, too :)

Ablesius:
Thanks for enabling this! :1st_place_medal:

PR #2163, opened 2020-06-09
[DE DateTimeV2] Enable recognition of dialectal clock expressions (#2036)

LionbridgeCSII:
Fix to issue #2036.
Added relevant test cases to TimeParser and DateTimeModel.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
@@ -48,6 +48,11 @@ namespace Microsoft.Recognizers.Definitions.German
       public const string WrittenMonthRegex = @"((monat\s*)?(?<month>apr(il|\.)|aug(ust|\.)|dez(ember|\.)|feb(ruar|ber|\.)|j[äa]n(uar|ner|\.)|jul(e?i|l\.)|jun([io]|\.)|märz|mai|nov(ember|\.)|okt(ober|\.)|sept?(ember|\.)))";
       public static readonly string MonthSuffixRegex = $@"(?<msuf>(im\s*|des\s*)?({RelativeMonthRegex}|{WrittenMonthRegex}|{MonthNumRegex}))";
       public const string DateUnitRegex = @"(?<unit>jahre|jahr|monate|monat|wochen?|tage|tag)(s)?";
+      public const string HalfTokenRegex = @"^(halb)";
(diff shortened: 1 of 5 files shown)


tellarin:
@Ablesius, can you take a first look before we review? Thanks!

.NET/Microsoft.Recognizers.Text.DateTime/German/Parsers/GermanTimeParserConfiguration.cs
```diff
@@ -49,17 +64,25 @@ public void AdjustByPrefix(string prefix, ref int hour, ref int min, ref bool ha
             var deltaMin = 0;
             var trimmedPrefix = prefix.Trim();
 
-            if (trimmedPrefix.StartsWith("halb"))
+            if (HalfTokenRegex.IsMatch(trimmedPrefix))
```
tellarin, line 67: Thanks for moving these to the right place! :)

Ablesius:
The specs look good. I'm not sure about all the regexes; to my knowledge, "dreiviertel vor n" is nonidiomatic.

.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
```diff
@@ -48,6 +48,11 @@ public static class DateTimeDefinitions
       public const string WrittenMonthRegex = @"((monat\s*)?(?<month>apr(il|\.)|aug(ust|\.)|dez(ember|\.)|feb(ruar|ber|\.)|j[äa]n(uar|ner|\.)|jul(e?i|l\.)|jun([io]|\.)|märz|mai|nov(ember|\.)|okt(ober|\.)|sept?(ember|\.)))";
       public static readonly string MonthSuffixRegex = $@"(?<msuf>(im\s*|des\s*)?({RelativeMonthRegex}|{WrittenMonthRegex}|{MonthNumRegex}))";
       public const string DateUnitRegex = @"(?<unit>jahre|jahr|monate|monat|wochen?|tage|tag)(s)?";
+      public const string HalfTokenRegex = @"^(halb)";
+      public const string QuarterToTokenRegex = @"^(viertel\s+vor|dreiviertel)\s*$";
+      public const string QuarterPastTokenRegex = @"^(viertel\s+nach)\s*$";
+      public const string ThreeQuarterToTokenRegex = @"^(viertel|dreiviertel\s+vor)\s*$";
```
Ablesius, line 54: "dreiviertel vor" isn't something you'd use in German, but it's probably a nice idea to add it because non-native speakers might become a little creative with forming these expressions, maybe.

PR #2167, opened 2020-06-11
[.NET] Hindi DateTime Set spec additions

LionbridgeCSII:
Support for additional cases in Set (3 in Extractor and 1 in Parser).
All new cases pass.

### DIFF
--- a/Specs/DateTime/Hindi/SetExtractor.json
+++ b/Specs/DateTime/Hindi/SetExtractor.json
@@ -72,6 +72,18 @@
     ]
   },
   {
+    "Input": "मैं ऐनुअली एक बार जाता हूं",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "ऐनुअली एक बार",
+        "Type": "set",
+        "Start": 4,
+        "Length": 13
+      }
+    ]
+  },
+  {
     "Input": "मैं हर दो दिन में जाती हूं",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
@@ -84,6 +96,18 @@
     ]
   },
   {
+    "Input": "मैं प्रत्येक दूसरे दिन जाती हूं",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": [
+      {
+        "Text": "प्रत्येक दूसरे दिन",
+        "Type": "set",
+        "Start": 4,
+        "Length": 18
+      }
+    ]
+  },
+  {
     "Input": "मैं हरेक तीन हफ़्तों में जाऊँगा",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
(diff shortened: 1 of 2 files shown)


Specs/DateTime/Hindi/SetExtractor.json
```diff
@@ -313,6 +349,7 @@
   },
   {
     "Input": "क्या मैं 09th मई को 2 रातों के लिए बुकिंग करवा सकता हूँ ",
+    "Comment": "(Can I do a booking for the 09th of May for 2 nights?). Hindi does not use plurals of weekdays such as 'Mondays', 'mornings', 'nights'",
```
tellarin, line 352: So there is no "equivalent" sentence that would generate a Set?
Also, based on the translation, it seems this spec should move to Model and output one date and one period?

Specs/DateTime/Hindi/SetParser.json
```diff
@@ -172,6 +172,7 @@
     "Context": {
       "ReferenceDateTime": "2017-09-27T12:25:54.2909444+03:00"
     },
+    "Comment": "The input literally translates to 'I'll leave each in two days' and 'in two days' is processed as Date instead of Duration",
```
tellarin, line 175: Is this because the translation was incorrect from the corresponding case in English? If so, can we add a new case with the correct translation?
This one can be moved to DateParser then (if not duplicate).

Specs/DateTime/Hindi/SetParser.json
```diff
@@ -197,6 +198,7 @@
     "Context": {
       "ReferenceDateTime": "2017-09-27T12:25:54.2959472+03:00"
     },
+    "Comment": "The input literally translates to 'I'll leave each in three weeks' and 'in three weeks' is processed as Date instead of Duration",
```
tellarin, line 201: Same.

Specs/DateTime/Hindi/SetParser.json
LionbridgeCSII: The problem is the translation is correct but the structure of the pattern is different in Hindi, and attempts to support it break other sub-entities. The solution seems to require Hindi specific modifications of the base code, so we have skipped it for now.

LionbridgeCSII:
> Just to confirm, these are just new specs to guarantee coverage, right? Do they differ in form from the previous ones? Or where they missed translations from the first round?

right, they were missed translations from the first round.

Specs/DateTime/Hindi/SetParser.json
tellarin: Got it. I misunderstood the comment at first.

Specs/DateTime/Hindi/SetExtractor.json
tellarin: @LionbridgeCSII, is this an equivalent situation to the other spec question?

LionbridgeCSII: similar, if I understand correctly, there is no difference in Hindi between 'night' and 'nights' and 'night' is also used in other contexts, for example as 'pm' indicator in time expressions. So, adding it to Set causes problems in other sub-entities. I am double checking if it is possible to support it in certain cases. 
The case is already included in the Model.

Issue #568, opened 2018-05-21
[CN Number] Incorrect number extraction cases


linzijia07:
"在下面的价格中选一个，$20，300美圆, ￥300, ok?"

It should be three numbers, i.e., "20", "300" and "300", while the recognizer gives me "20，300" and "300". Note that "，" in "20，300" is a Chinese comma.

tellarin:
On triage, issue was marked as *very low priority* as it's an ambiguous corner case.

PR #2171, opened 2020-06-15
[CN Currency] Fix incorrect extraction cases

visionshao:
Fix the issue #568
Add codes for detecting conflicts to extract function of NumberWithUnitExtractor for all platforms
Add codes for solving conflicts when merging results to AbstractNumberWithUnitModel for .NET and Python
Add test case to Specs

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
@@ -79,6 +79,25 @@ namespace Microsoft.Recognizers.Text.NumberWithUnit
             {
                 var numbers = this.config.UnitNumExtractor.Extract(source).OrderBy(o => o.Start);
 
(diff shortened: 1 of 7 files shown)


.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -79,6 +79,25 @@ public List<ExtractResult> Extract(string source)
             {
                 var numbers = this.config.UnitNumExtractor.Extract(source).OrderBy(o => o.Start);
 
+                if (numbers.Count() > 0 && CheckExtractorType(Constants.SYS_UNIT_CURRENCY))
+                {
+
+                    foreach (var number in numbers)
```
tellarin, line 85: Can you add a code comment here explaining what this is doing?
Also, can you check if there's measurable performance impact?
Thanks.

visionshao: ok

tellarin:
Closing this as it has been superseded by #2193.

Issue #840, opened 2018-09-06
[EN NumberRange] Fix wrong fraction extraction like "5 in 3", "8 out of 5"


guom08:
**Describe the bug**
For fractions like "N in M", "N out of M", it's only valid when "N <= M"

**Expected behavior**
"3 in 5" should be 3/5
"6 in 5", we should not extract anything

PR #2185, opened 2020-06-20
[EN Number] Fix wrong fraction extraction (#840)

LionbridgeCSII:
Fix to issue #840 

Added relevant test cases to NumberModel.

2 test cases needed to be updated in NumberRangeModel:
- "The number is equals to 20000 in 1998"
- "The number is from 200 to 3000000 in 2008"

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Recognizers.Text.Number
 
         public virtual NumberOptions Options { get; } = NumberOptions.None;
 
+        public virtual BaseNumberParser NumberParser { get; }
+
         internal abstract ImmutableDictionary<Regex, TypeTag> Regexes { get; }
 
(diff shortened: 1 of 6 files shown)


.NET/Microsoft.Recognizers.Text.Number/Extractors/BaseNumberExtractor.cs
```diff
@@ -64,6 +66,20 @@ public virtual List<ExtractResult> Extract(string source)
                         continue;
                     }
 
+                    // Matches containing separators 'in', 'out of' should be considered fractions only when numerator < denominator
+                    if (m.Groups["ambiguousSeparator"].Success)
+                    {
+                        var numerator = m.Groups["numerator"];
+                        var denominator = m.Groups["denominator"];
+                        int num = ParseNumber(numerator);
```
tellarin, line 74: This can be very costly. We need to run some perf impact tests here.

LionbridgeCSII: yes I was worried about it, especially the use of NumberParser. But since it runs only when ambiguousSeparator is matched, it should have an impact only for the cases "N in M" or "N out of M"

tellarin (approved):
Preliminary analysis of available data didn't show high frequency of such cases, so perf hit is currently borderline acceptable.

PR #2188, opened 2020-06-23
[SV NumberRange] Preparations to add support for NumberRange

pete1854:
Added initial code for the Range Recognizer. Refactored current swedish number recognizer constructors w/ new signatures in order to be code aligned.


tellarin:
@pete1854, feel free to make a PR with all the NumberRange specs in Swedish marked as "NotSupported". We can then slowly enable them as support for cases is implemented/validated.

PR #2192, opened 2020-06-26
[SV PhoneNumbers] Add support for swedish phone numbers NEW

pete1854:
Adding support for phone numbers. Format should reflect format specified by PTS. All test pass.


.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30204.135
```
tellarin, line 4: Be careful not to change the solution unless intentional.

Patterns/Base-PhoneNumbers.yaml
```diff
@@ -35,6 +35,9 @@ ITPhoneNumberRegex: !paramsRegex
 NLPhoneNumberRegex: !paramsRegex
   def: ((((\(\s?(\+\s?|00)31\s?\)\s?)|(((?<!\d)\+\s?|{WordBoundariesRegex}00)31\s?))?((({WordBoundariesRegex}|(?<=31))0?\d{1,3}|\(\s?0?\d{1,3}\s?\)|\(0\)[-\s]?\d{1,3})((-?[\d]{5,11})|(\s[\d\s]{5,11}))\d))|\b\d{10,12})(?!-){EndWordBoundariesRegex}
   params: [ WordBoundariesRegex, EndWordBoundariesRegex  ]
+SEPhoneNumberRegex: !paramsRegex
+  def: ({WordBoundariesRegex}((\+|00)46\s\d{1,3}\s\d{2,3}\s\d{2,3}(\s\d{2})?{EndWordBoundariesRegex}|\d{2,4}-\d{2,3}\s\d{2,3}(\s\d{2})?){EndWordBoundariesRegex})
```
tellarin, line 39: Can you describe in the PR description some of the key cases not previously captured, but nor captured by this beyond just numbers with the country code for Sweden?

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -803,6 +803,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Turkish", "Turkish", "{2F41
 		..\Specs\DateTime\Turkish\TimePeriodParser.json = ..\Specs\DateTime\Turkish\TimePeriodParser.json
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Swedish", "Swedish", "{2AA6937E-36C0-4E6B-B536-AA0E08D30261}"
```
tellarin, line 806: Nitpick: Can you manually edit the file so this Swedish spec additional is near the other ones (clustered by culture)?

Specs/Sequence/Swedish/PhoneNumberModel.json
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Tel: 08-12 45 67",
+    "NotSupportedByDesign": "javascript, python, java",
+    "Results": [
+      {
+        "Text": "08-12 45 67",
```
tellarin, line 39: Are such short number still common?

tellarin: This specific format seems quite prone to false positives, especially with such high score (0.9).

pete1854: @tellarin - could you please elaborate a little around scoring and its meaning in this situation? Why is a score close to 1 not good?

.NET/Microsoft.Recognizers.Text.sln
pete1854: This was highly unintentional. I must check my settings because I had to keep manually revert this change. I forgot it this time.

Specs/Sequence/Swedish/PhoneNumberModel.json
tellarin: The more well formatted the number is and the more confident the recognizers are, the closer to 1 the score should be.
But for this particular case, in my opinion, it may be too prone to false positives, so this should ideally be somehow reflected in a lower score.
Are such short numbers still common? And specifically in this format?

pete1854: Thanks for the explanation. To be honest I don’t know how common this particular number construct is nowadays. I have simply followed the PTS standard for each possible construct a Swedish phone number can have. By false-positives, do you mean there’s a risk it will show a match for a non-swedish phone number? Just so I’ve got it right: this recognizer should match swedish phone numbers and swedish phone numbers only, right?

tellarin: By false-positives I mean, for example, matching something that is not a phone number. Maybe a date followed by some numbers. Or other cases.

tellarin: Do you have a link to the PTS standard? If so, can you add it in the PR description?

pete1854: @tellarin - I have added a reference to the standard in the description.

pete1854: @tellarin - I could extend the specs w/ an array of cases containing a mix of phone numbers and other numbers (like the english spec) to ensure the recognizer doesn’t hit these. Should I do that?

tellarin: @pete1854, that would be very helpful. Specially phrases with partial date and time, other numerical tokens, IDs, etc.
We don't need too many, but a couple cases focused on patterns that may be similar to one of the new phonenumber formats. Thanks!

tellarin: Based on the standard you linked, it seems the only NDC that can fall into the easier to mismatch format is Stockholm (08).
So let's at least add an extra validation that only numbers starting with 08 can following the "0N-AX XX XX" format.

tellarin: The others with 0xx-xxx or 0xx-xxx are much less likely to be problematic, I think.

pete1854:
Abandoning

PR #2195, opened 2020-06-26
[.NET] Dutch NumbersWithUnits Currency enhancements

LionbridgeCS:
Added support for 4 previously skipped test cases. One test case is not supported yet, "Input": "Hij is me nog een tientje schuldig.". Fractions are supported by this component, however amounts higher than 1 Euro are not, such as 'tientje, 10 Euro. Also, 'dubbeltje' and 'stuiver' are still in use in common Dutch language, but 'tientje' not, ref: https://nl.wikipedia.org/wiki/Nederlandse_gulden". Therefore, the unit "Tientje" is introduced in an attempt to solve this test case, although this test case remains unsupported.


Specs/NumberWithUnit/Dutch/CurrencyModel.json
```diff
@@ -178,13 +177,14 @@
...
     "Input": "Hij is me nog een tientje schuldig.",
+    "Comment": "Fractions are supported, however amounts higher than 1 Euro are not, such as 'tientje', 10 Euro. Also, 'dubbeltje' and 'stuiver' are still in use, but 'tientje' not, ref: https://nl.wikipedia.org/wiki/Nederlandse_gulden",
     "Results": [
       {
         "Text": "tientje",
         "TypeName": "currency",
         "Resolution": {
-          "unit": "Euro",
-          "value": "10"
+          "unit": "Tientje",
```
tellarin, line 186: I don't think this is the behaviour we want. It seems to me the original spec was the correct output. 
To properly support it we'd need some kind of conversion from Tientje to Euro in the parser.

Specs/NumberWithUnit/Dutch/CurrencyModel.json
```diff
@@ -195,69 +195,66 @@
     "Input": "Een kwartje is 25 cent.",
     "Results": [
       {
-        "Text": "kwartje",
+        "Text": "een kwartje",
         "TypeName": "currency",
         "Resolution": {
-          "unit": "Netherlands guilder",
-          "value": "0,25"
+          "unit": "Kwartje",
```
tellarin, line 201: Similar to the 10 Euro case.

Specs/NumberWithUnit/Dutch/CurrencyModel.json
tellarin: @iMicknl, what do you think? I'm OK to merge as-is as a temporary solution, if it is useful, until we have the next version. 
After all, by at least tagging these terms as their own entities, it allows downstream processing; while currently they are lost.

iMicknl: I am not sure, to be honest... Dubbeltje, kwartje and tientje were originally part of the Nederlandse Gulden. (https://nl.wikipedia.org/wiki/Nederlandse_gulden). However, nowadays tientje could also mean 10 € instead of 10 gulden. 

I don't think this will be commonly used. Dubbeltje / kwartje is more commonly used, even for Euro, but not sure how this should be mapped.

Issue #2197, opened 2020-06-28
When does Python package supports German culture?


jhoelzl:
Hi,

is there a plan or roadmap when approximately the python package (https://pypi.org/project/recognizers-text-suite/) also support the German culture?
Thanks!

tellarin:
Unfortunately we don't current have an estimated time frame for that. What your usage scenario?
Would you be interested in helping contribute support?

jhoelzl:
I am working on a NLP project in English, German and Dutch. So this package would be perfect.

> Would you be interested in helping contribute support?
Definitely! Do i "just" have to build a related German language folder and related files as defined in https://github.com/microsoft/Recognizers-Text/tree/master/Python/libraries/recognizers-number/recognizers_number/number?
Maybe there is a short documentation how to add new languages to the python packages?

tellarin:
That's basically it, but in creating the extractor/parser configs for German it may be necessary to port over some behaviour from the .NET implementation.

tellarin:
These two links give an overview of concepts and processes. Unfortunately they are .NET focused and may be a little out of date.
Let us know if you have questions, ok?

- https://blog.botframework.com/2018/02/01/contributing-luis-microsoft-recognizers-text-part-2/
- https://blog.botframework.com/2018/02/12/contributing-luis-microsoft-recognizers-text-part-3/

jhoelzl:
Added pull request: #2198.

jkondek1:
Hi guys, **the German culture still does not work in case one installs the library from PyPI**. If the repo is cloned and then everything installed manually step-by-step, no problems occur. Is it possible the version in PyPI was not updated?

BrianInGermany:
I've tried my hand at the German date-time modules for Python [#2728](https://github.com/microsoft/Recognizers-Text/pull/2728)

tellarin:
As of the merge of #2914, there's now German support for Number/Currency/DateTimeV2 in Python close to parity with .NET
Help is needed to complete the other NumberWithUnit types and close the support gap to .NET.

PR #2198, opened 2020-06-29
[Python] Add German support for recognizers-text-number

jhoelzl:
Add initial German files. Towards implementing #2197.


jhoelzl:
> Ideally we should mark all cases in German DateTime specs as NotSupported in python and re-enable later.

And how / where should i do this?

tellarin:
> > Ideally we should mark all cases in German DateTime specs as NotSupported in python and re-enable later.
> 
> And how / where should i do this?

This should be an attribute added to each spec test case in the German DateTime json files under /Specs. We don't currently have a script in the repo for it, so either we need to do manually, or add such script.

Python/libraries/recognizers-number/recognizers_number/resources/german_numeric.py
```diff
@@ -70,10 +70,10 @@ def DoubleWithoutIntegralRegex(placeholder):
     DoubleWithRoundNumber = f'(((?<!\\d+(\\s*{BaseNumbers.NumberMultiplierRegex})?\\s*)-\\s*)|((?<=\\b)(?<!\\d+[\\.,])))\\d+[\\.,]\\d+\\s+{RoundNumberIntegerRegex}(?=\\b)'
     DoubleAllFloatRegex = f'((?<=\\b){AllFloatRegex}(?=\\b))'
     ConnectorRegex = f'(?<spacer>und)'
-    NumberWithSuffixPercentage = f'(?<!%)({BaseNumbers.NumberReplaceToken})(\\s*)(%(?!{BaseNumbers.NumberReplaceToken})|(per\\s*cents?|prozent|cents?)\\b)'
+    NumberWithSuffixPercentage = f'(?<!%)({BaseNumbers.NumberReplaceToken})(\\s*)(%(?!{BaseNumbers.NumberReplaceToken})|(Prozent(e)?)\\b)'
```
tellarin, line 73: Such changes need to be made in the corresponding YAML file. The platform-specific resource files are auto-generated from it. 
If these changes are meant to support new cases not covered in the specs, please add new spec cases to show the new behaviour, ok?
Also, all terms in regexes should be lowercase (due to performance).

Python/libraries/recognizers-number/recognizers_number/resources/german_numeric.py
```diff
@@ -144,11 +144,14 @@ def DoubleWithoutIntegralRegex(placeholder):
                               ("achtzig", 80),
                               ("neunzig", 90),
                               ("hundert", 100),
+                              ("einhundert", 100),
                               ("tausend", 1000),
+                              ("eintausend", 1000),
```
tellarin, line 149: Is this really needed?

tellarin:
@jhoelzl, I've made a few tweaks that I'll push here soon.
Two main comments: 
- Instead of changing the resources file in Python manually, the edits should happen on the patterns YAML file, which is used to automatically generated the resource code across platforms. Otherwise, any changes are lost on build;
- As the test cases current pass in .NET, test failures are likely to not be caused by pattern issues. I suggest checking where your new Python implementation differs from the corresponding .NET one for German.

Python/libraries/recognizers-number/recognizers_number/resources/german_numeric.py
jhoelzl: I added it because i though it would fix a failed test, but it does not solve it. I will remove it.

tellarin:
@jhoelzl Can you give me permission to push to your PR branch? I'm getting an error.
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

jhoelzl:
I want to fix he 24 remaning test cases, however, i am bit confused.

> As the test cases current pass in .NET, test failures are likely to not be caused by pattern issues. I suggest checking where your new Python implementation differs from the corresponding .NET one for German.

For example the word "einhundert" which means "one hundred" is not recognized in the test cases although the definition of ``RoundNumberIntegerRegex`` in Python covers the same as the one in .NET:

* .NET: https://github.com/jhoelzl/Recognizers-Text/blob/python-german-support/.NET/Microsoft.Recognizers.Definitions.Common/German/NumbersDefinitions.cs#L28
* Python: https://github.com/jhoelzl/Recognizers-Text/blob/python-german-support/Python/libraries/recognizers-number/recognizers_number/resources/german_numeric.py#L21

tellarin:
@jhoelzl it's not a problem with patterns or definitions now. There may be behaviour differences in two places compared to .NET:
- differences in the German extractor/parser configs you created in Python (as they were created from the English ones);
- differences in base classes (that utilize the data and methods in the config to do the actual work), as the .NET classes usually evolve faster.
I suggest picking one case that fails in Python and then debugging the same input in .NET to see where things differ.

tellarin (approved):
Approving to merge as-is without registering the number model in German as this can serve as the new support skeleton structure.

PR #2202, opened 2020-06-30
[Sequence - CreditCard] Add structure for Credit/Bank Card recognizer in Chinese

visionshao:
Following the implement of PhoneNumber.
Now, it can only work by SimpleConsole. Can not be tested;

Added Modules
Patterns
Base-CreditCard.yaml
Chinese-CreditCard.yaml

Microsoft.Recognizers.Text.Sequence
Extractor:
BaseCreditCardExtractor.cs

Config:
CreditCardExtractorConfiguration.cs

Models:
CreditCardModel.cs

Parser:
CreditCardParser.cs

Constants
SYS_CREDIT_CARD = "builtin.creditcard"
MODEL_CREDIT_CARD = "creditcard"
CREDIT_CARD_REGEX_CN = "CNCreditCard"

Definition
BaseCreditCard.cs
Chinese\CreditCardDefinitions.cs

Chinese
Extractors
ChineseCreditCardExtractorConfiguration.cs

SequenceRecognizer
Public static List RecognizeCreditCard
Public IModel GetCreditCardModel
RegisterModel

Other Changes
IExtractor.cs
Add new fields : Validation and Issuer (Maybe this can be done better)
IParser.cs
Add new fields : Validation and Issuer (Maybe this can be done better)
ResolutionKey
Add new fields : Validation and Issuer
Program.cs
Add SequenceRecognizer.RecognizeCreditCard(query, culture)

Test Cases
这是中国的银行卡6200 0000 0000 0000 123
这是中国的银行卡6200 0000 0000 0000
这是中国的银行卡6200000000000000123
这是中国的银行卡6200000000000000
这是中国的银行卡6200 0000 00000000123
这是中国的银行卡62 0 00 0000 0000 000
这是中国的银行卡3100 0000 0000 0000 123
这是中国的银行卡3100000000000000123
这是中国的银行卡3100 0000 00000 000123


.NET/Microsoft.Recognizers.Definitions.Common/BaseCreditCard.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+
+    public static class BaseCreditCard
+    {
+      public const string WordBoundariesRegex = @"\b";
+      public const string NonWordBoundariesRegex = @"\B";
+      public const string EndWordBoundariesRegex = @"\b";
+      public static readonly Func<string, string, string> CNCreditCardRegex = (WordBoundariesRegex, EndWordBoundariesRegex) => $@"((62|81)(\s?\d){{2}}(\s|-)?(\s?\d){{4}}(\s|-)?(\s?\d){{4}}(\s|-)?(\s?\d){{4}}(\s|-)?(\s?\d){{0,3}}|31(\s?\d){{2}}(\s|-)?(\s?\d){{4}}(\s|-)?(\s?\d){{4}}(\s|-)?(\s?\d){{4}}(\s|-)?(\s?\d){{3}})";
```
tellarin, line 27: Base should be used for all regexes, not only CN. 
Also, my understanding is that all numbers will fall in a similar range, and later prefixes can be checked to assign issuer, which also doesn't always directly map to country.
We should avoid any country mapping them. And card numbers can appear in any language.

.NET/Microsoft.Recognizers.Text.Sequence/Chinese/Extractors/ChineseCreditCardExtractorConfiguration.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions.Chinese;
+
+namespace Microsoft.Recognizers.Text.Sequence.Chinese
```
tellarin, line 6: As all cards should be recognized in all languages. It's preferable to move the basic implementation to English, and make all languages fallback to it, unless they need special config. Similar to how URL works.

.NET/Microsoft.Recognizers.Text.Sequence/Constants.cs
```diff
@@ -69,5 +73,7 @@ public static class Constants
         public const string URL_REGEX = "Url";
 
         public const string GUID_REGEX = "Guid";
+
+        public const string CREDIT_CARD_REGEX_CN = "CNCreditCard";
```
tellarin, line 77: This shouldn't be necessary, right?

.NET/Microsoft.Recognizers.Text.Sequence/Chinese/Extractors/ChineseCreditCardExtractorConfiguration.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class ChineseCreditCardExtractorConfiguration : CreditCardConfiguration
+    {
+
+        public ChineseCreditCardExtractorConfiguration(SequenceOptions options)
+            : base(options)
+        {
+            var regexes = new Dictionary<Regex, string>
+            {
+                {
+                    new Regex(CreditCardDefinitions.ChinaUnionPayRegex, RegexOptions.Compiled),
```
tellarin, line 17: This is very inefficient just to check for prefixes. And it would get worst as more and more options are added here.

.NET/Microsoft.Recognizers.Text.Sequence/Parsers/CreditCardParser.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions;
+
+namespace Microsoft.Recognizers.Text.Sequence.Parsers
+{
+    public class CreditCardParser : BaseSequenceParser
+    {
+
+        // @TODO move regexes to base resource files
```
tellarin, line 11: Parser should figure out the issuer/etc, crc, and filter out false positive matches.

.NET/Microsoft.Recognizers.Text/Extractors/IExtractor.cs
```diff
@@ -11,6 +11,8 @@ public interface IExtractor
 
     public class ExtractResult : ICloneableType<ExtractResult>
     {
+        public string Issuer { get; set; } = null;
```
tellarin, line 14: These attributes should not be added to the base class for ExtractResult, and also not in the Text project.
These are specific of the CC entity type. Perhaps add them in a resolution attribute. Like the case for relative ordinals, or datetimes with its many attributes.

.NET/Microsoft.Recognizers.Text/Models/ResolutionKey.cs
```diff
@@ -11,5 +11,7 @@ public static class ResolutionKey
         public const string SubType = "subtype";
         public const string Offset = "offset";
         public const string RelativeTo = "relativeTo";
+        public const string Issuer = "issuer";
```
tellarin, line 14: This is not ideal, but can be acceptable.
It would be better to do how DateTime does it.

.NET/Microsoft.Recognizers.Text/Parsers/IParser.cs
```diff
@@ -19,6 +19,8 @@ public ParseResult(ExtractResult er)
             Text = er.Text;
             Type = er.Type;
             Metadata = er.Metadata;
+            Issuer = er.Issuer;
```
tellarin, line 22: Same.

.NET/Samples/SimpleConsole/Program.cs
```diff
@@ -15,7 +15,7 @@ namespace SimpleConsole
     public static class Program
     {
         // Use English for the Recognizers culture
-        private const string DefaultCulture = Culture.English;
+        private const string DefaultCulture = Culture.Chinese;
```
tellarin, line 18: Be careful not to change the default language in the demo code.

tellarin:
Closing this as it's incomplete and the original contributor notified us offline he won't be able to continue it.

PR #2216, opened 2020-07-17
[Sample] Update Azure Function to .NET core and V3 SDK

iMicknl:


tellarin:
We used the hackathon to change the build system. Re-running this build now. :)

Issue #2037, opened 2020-02-14
[* Unit] Utterances with compound entity units do not work correctly - dimension


asishsinha:
**Describe the bug**
For utterance, "12 ft 2 in" prebuilt dimension entity considers "12 ft 2" as one dimension entity and "2 in" another.

**To Reproduce**
1. Go to LUIS portal add a prebuilt entity - dimension, train and test.
2. Type '12 ft 2 in'
3. See this wrong outputs:
![image](https://user-images.githubusercontent.com/10179672/74543937-4d4a5200-4f6c-11ea-8d62-3ce02452eabb.png)

**Expected behavior**
12 ft - dimension
2 in - dimension

**Sample input/output**
Input: 12 ft 2 in
Output: 
12 ft 2 - dimension
2 in - dimension

**Platform (please complete the following information):**
 - Environment: LUIS, portal
 - Version V .01

**Additional screenshots**
![image](https://user-images.githubusercontent.com/10179672/74544380-090b8180-4f6d-11ea-828d-ce386742e1d8.png)

tellarin:
The expected behaviour according to design is that the output is one entity, as it is compound: "2ft 2in".

asishsinha:
> The expected behaviour according to design is that the output is one entity, as it is compound: "2ft 2in".

But then, even this shows two entities.
![image](https://user-images.githubusercontent.com/10179672/75346474-caf35380-58c4-11ea-9139-8fdba716fe11.png)

tellarin:
@LionbridgeCSII, can you take a look at this one?

LionbridgeCSII:
@tellarin, currently only Currency supports compound units. We can extend support to Dimension, but it wil require maybe 20-30 hours.

tellarin:
@LionbridgeCSII, it seems there are two issues in the current system. 
1) "12 ft 2 in" returning "12 ft 2", instead of the two separate entities; and
2) the lack of proper support for compound entities as "12 ft 2 in" as a single entity.

Ideally we want the known issue to be fixed first. I.e. return two correct entities.
And the new compound behaviour also implemented, but only active if a new config flag is enabled (EnableCompoundTypes). We need to put this behaviour behind a flag to avoid a breaking change for current users.

If this fits into your estimate, I'd say go for it.
For the flag behaviour, you can see a similar flag usage in Numbers for SuppressExtendedTypes.

PR #2217, opened 2020-07-21
[* Unit] Fix unit extraction in utterances with compound entity (#2037)

LionbridgeCSII:
fix to first issue pointed out in #2037 ("12 ft 2 in" returning "12 ft 2", instead of the two separate entities)
added relevant test cases to DimensionModel


tellarin (approved):

tellarin:
@LionbridgeCSII can you resolved the conflict? We should probably still merge this before the compound units PR is ready to merge. What do you think?

LionbridgeCSII:
@tellarin, conflicts solved

PR #2218, opened 2020-07-23
[ES DateTimeV2] Add new cases to Spanish DateTimeModel spec

t-hanyinMS:
New cases for Spanish DateTimeModel spec are added. The newly enabled cases are all marked as "NotSupported": "javascript, python, java", so the gaps for .NET can be seen. 
The current new cases are revised by vendors.

- previous number of cases: 49

- number of cases now: 679

- cases added: 630


Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -1491,5 +1491,17039 @@
...
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(XXXX-WXX-1T09:30,XXXX-WXX-1T15:00,PT5H30M)",
+              "type": "datetimerange",
+              "start": "2019-01-21 09:30:00",
+              "end": "2019-01-21 15:00:00"
+            },
+            {
+              "timex": "(XXX
```
tellarin, line 14322: Let's see the build output now, but typically articles should not be included in the extracted entity unless they change its meaning.

tellarin, line 14429: This is incorrect. The last trimester/quarter are the last 3 months of the year.
This resolution is the 3rd trimester.

tellarin, line 14517: Same

tellarin:
@t-hanyinMS, some test failures seem real unsupported case, but there are a lot of mistakes in specs that need review/fixing.

t-hanyinMS:
> @t-hanyinMS, some test failures seem real unsupported case, but there are a lot of mistakes in specs that need review/fixing.

Thanks, these specs will be updated again after they are revised.

tellarin (approved):
OK to merge. Disabled tests are clustered with comments and issues will be open for each issue category.

PR #2221, opened 2020-07-27
[* Unit] Support for compound dimension units (#2037)

LionbridgeCSII:
Added support for compound dimension units so that they can be extracted as a single entity.
The compound behaviour is controlled by the flag EnableCompoundTypes as requested in #2037.
Relevant test cases have been added to the new spec file DimensionModelCompoundMode.json.


tellarin:
@LionbridgeCSII, @LionbridgeCS2, should #2217 be merged first? Or does this one replace it?

LionbridgeCSII:
@tellarin, this one contains also the commit from #2217.
But it seems that the use of 'double[]' in the new dictionary DimensionUnitMultiplesMap is breaking the javascript build

tellarin:
> @tellarin, this one contains also the commit from #2217.
> But it seems that the use of 'double[]' in the new dictionary DimensionUnitMultiplesMap is breaking the javascript build

Does javascript support something like "number[]"? If so, the code generator can be modified to translate the type. Can you test this and reply here? If it would work, we can plan to change the generator soon.

LionbridgeCSII:
> Does javascript support something like "number[]"? If so, the code generator can be modified to translate the type. Can you test this and reply here? If it would work, we can plan to change the generator soon.

I remember trying using other types like "int[]" and "string[]", (not sure about "number[]" but will try again), but they all gave the same problem

LionbridgeCSII:
@tellarin, with "number[]" javascript builds correctly, but .NET does not support it.

tellarin:
I'll try to commit a change to JavaScript code generator in this PR soon to make the conversion. The YAML should still declare it as double.

tellarin:
I'm making the mentioned change in another PR that will also update all generated resources across platforms. Once that's merged, this PR can be rebased on /master and it should pass. Then we can go through the full review.
Sorry for the delay.

tellarin:
@LionbridgeCSII, PR #2273 included the code gen change. You should now be able to use double[] in the YAML dictionary.

LionbridgeCSII:
@tellarin, it seems that now java compilation fails

LionbridgeCSII:
> @tellarin, it seems that now java compilation fails

I tried replacing 'double[]' with other types ('number[]', 'int[]', 'float[]'....), but only 'string[]' seems to be supported and does not break the build.

Issue #2214, opened 2020-07-16
[ES NumberRange] Compound disjoint range not extracted correctly


tellarin:
Separately they work fine, but the second range is not extracted when both occur in the same sentence.
"no menos de 194 y menos de 12"

PR #2227, opened 2020-08-03
[EN/ES NumberRange] Fixed compound disjoint range extraction (#2214)

LionbridgeCSII:
Fixes issue  #2214 (for both English and Spanish).
Added relevant test cases in NumberRangeModel.

### DIFF
--- a/Patterns/Spanish/Spanish-Numbers.yaml
+++ b/Patterns/Spanish/Spanish-Numbers.yaml
@@ -179,7 +179,7 @@ LessRegexNoNumberSucceed: !simpleRegex
 EqualRegexNoNumberSucceed: !simpleRegex
   def: ((igual(es)?|equivalentes?|equivalen?)((?!\s+(al?|que|del?))|(\s+(al?|que|del?)(?!(\s*\d+)))))
 OneNumberRangeMoreRegex1: !nestedRegex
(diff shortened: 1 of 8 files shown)


Patterns/Spanish/Spanish-Numbers.yaml
```diff
@@ -190,6 +193,9 @@ OneNumberRangeMoreSeparateRegex: !nestedRegex
 OneNumberRangeLessRegex1: !nestedRegex
   def: ({LessOrEqual}|{LessRegex})\s*((el|las?|los)\s+)?(?<number2>({NumberSplitMark}.)+)
   references: [ LessOrEqual, LessRegex, NumberSplitMark ]
+OneNumberRangeLessRegex1LB: !nestedRegex
```
tellarin, line 196: What does LB stand for? Wouldn't it be possible to treat it as a prefix instead of altering all regexes?

LionbridgeCSII: It stands for LookBehind. The problem is that the solution breaks the build in Java, so the new regex is only used in .NET

tellarin (changes requested):
Could you also apply the same change to Python and JS?

LionbridgeCSII:
> Could you also apply the same change to Python and JS?

but it seems that at the moment NumberRange is not implemented in javascript and python.

tellarin (approved):
Approving as-is to unblock fix.

Issue #2210, opened 2020-07-16
[ES DateTimeV2] Support "de 2015" as a date range under Experimental/Analysis mode


tellarin:
Entities like "de 2015" should be recognized as a range under ExperimentalMode (Analysis mode). Parity with English.

PR #2230, opened 2020-08-04
[ES DateTimeV2] Added support for "de 2015" as a date range under Experimental mode (#2210)

LionbridgeCSII:
Added support for ranges like "de 2015" in Spanish (under Experimental mode) in analogy to the English case "from 2015" (#2210). 
Test case added to the new spec file DateTimeModelExperimentalMode.json

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
@@ -183,7 +183,8 @@ namespace Microsoft.Recognizers.Definitions.Spanish
       public static readonly string HolidayRegex3 = $@"\b(?<holiday>(d[ií]a( del?( las?)?)? )(trabajador|madres?|padres?|[aá]rbol|mujer(es)?|solteros?|niños?|marmota|san valent[ií]n|maestro))(\s+(del?\s+)?({YearRegex}|(?<order>(pr[oó]xim[oa]?|est[ea]|[uú]ltim[oa]?|en))\s+año))?\b";
       public const string BeforeRegex = @"(antes(\s+del?(\s+las?)?)?)";
       public const string AfterRegex = @"(despu[eé]s(\s*del?(\s+las?)?)?)";
-      public const string SinceRegex = @"(desde(\s+(las?|el))?|de)";
+      public const string SinceRegex = @"(desde(\s+(las?|el))?)";
+      public const string SinceRegex1 = @"(desde(\s+(las?|el))?|de)";
       public const string AroundRegex = @"(?:\b(?:cerca|alrededor|aproximadamente)(\s+de\s+(las?|el))?\s*\b)";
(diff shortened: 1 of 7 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishMergedParserConfiguration.cs
```diff
@@ -6,12 +6,15 @@ namespace Microsoft.Recognizers.Text.DateTime.Spanish
...
+        private const RegexOptions RegexFlags = RegexOptions.Singleline | RegexOptions.ExplicitCapture;
+
         public SpanishMergedParserConfiguration(IDateTimeOptionsConfiguration config)
             : base(config)
         {
             BeforeRegex = SpanishMergedExtractorConfiguration.BeforeRegex;
             AfterRegex = SpanishMergedExtractorConfiguration.AfterRegex;
-            SinceRegex = SpanishMergedExtractorConfiguration.SinceRegex;
+            SinceRegex = (config.Options & DateTimeOptions.ExperimentalMode) != 0 ?
+                new Regex(DateTimeDefinitions.SinceRegex1, RegexFlags) : new Regex(DateTimeDefinitions.SinceRegex, RegexFlags);
```
tellarin, line 17: This should read from the MergedExtractor config as above, instead of recreating here.

Specs/DateTime/Spanish/DateTimeModelExperimentalMode.json
```diff
@@ -0,0 +1,28 @@
+[
+  {
+    "Input": "El rango es de 2014.",
```
tellarin, line 3: Can you add a similar case in the default DateTimeModel.json to show the different behaviour there is this config is not enabled?

LionbridgeCSII: added

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishMergedParserConfiguration.cs
LionbridgeCSII: done, but the conditional statement seems to be required.

.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
```diff
@@ -184,6 +184,7 @@ public static class DateTimeDefinitions
       public const string BeforeRegex = @"(antes(\s+del?(\s+las?)?)?)";
       public const string AfterRegex = @"(despu[eé]s(\s*del?(\s+las?)?)?)";
       public const string SinceRegex = @"(desde(\s+(las?|el))?)";
+      public const string SinceRegex1 = @"(desde(\s+(las?|el))?|de)";
```
tellarin, line 187: Rename to SinceRegexExp.

tellarin (approved):
The rename can come in a followup PR. Thanks.

Issue #654, opened 2018-06-25
[EN DatetimeV2] Support "the end of " + specific "year/month/week"


guom08:
Also check to support cases like 
"from 2007 to the end of 2008"

PR #2232, opened 2020-08-05
[EN DatetimeV2] Added support for "the end of " in date period (#654)

LionbridgeCSII:
Added support for ranges like "from 2007 to the end of 2008" (#654).

Isolated occurrences of "the end of " + specific "year/month/week" seem to be already supported.
However, there is maybe an inconsistency in the parsed result when the extraction is recognized as daterange or duration.
For example:
"end of 2016"      -> "start": 2016-09-01
"end of this year" -> "start": 2016-07-01

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Recognizers.Definitions.English
       public static readonly string AllHalfYearRegex = $@"({HalfYearFrontRegex})|({HalfYearBackRegex})|({HalfYearRelativeRegex})";
       public const string EarlyPrefixRegex = @"\b(?<EarlyPrefix>early|beginning of|start of|(?<RelEarly>earlier(\s+in)?))\b";
       public const string MidPrefixRegex = @"\b(?<MidPrefix>mid-?|middle of)\b";
-      public const string LaterPrefixRegex = @"\b(?<LatePrefix>late|end of|(?<RelLate>later(\s+in)?))\b";
+      public const string LaterPrefixRegex = @"\b(?<LatePrefix>late|(the\s+)?end of|(?<RelLate>later(\s+in)?))\b";
       public static readonly string PrefixPeriodRegex = $@"({EarlyPrefixRegex}|{MidPrefixRegex}|{LaterPrefixRegex})";
       public const string PrefixDayRegex = @"\b((?<EarlyPrefix>early)|(?<MidPrefix>mid(dle)?)|(?<LatePrefix>later?))(\s+in)?(\s+the\s+day)?$";
       public const string SeasonDescRegex = @"(?<seas>spring|summer|fall|autumn|winter)";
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -279,7 +281,9 @@ public static class DateTimeDefinitions
       public static readonly string YearPeriodRegex = $@"((((from|during|in)\s+)?{YearRegex}\s*({TillRegex})\s*{YearRegex})|(((between)\s+){YearRegex}\s*({RangeConnectorRegex})\s*{YearRegex}))";
       public static readonly string StrictTillRegex = $@"(?<till>\b(to|(un)?till?|thru|through)\b|{BaseDateTime.RangeConnectorSymbolRegex}(?!\s*(h[1-2]|q[1-4])(?!(\s+of|\s*,\s*))))";
       public static readonly string StrictRangeConnectorRegex = $@"(?<and>\b(and|through|to)\b|{BaseDateTime.RangeConnectorSymbolRegex}(?!\s*(h[1-2]|q[1-4])(?!(\s+of|\s*,\s*))))";
-      public static readonly string ComplexDatePeriodRegex = $@"(?:((from|during|in)\s+)?(?<start>.+)\s*({StrictTillRegex})\s*(?<end>.+)|((between)\s+)(?<start>.+)\s*({StrictRangeConnectorRegex})\s*(?<end>.+))";
+      public static readonly string TillRegexWithDet = $@"({TillRegex}(\s+the\b)?)";
```
tellarin, line 284: Why do we need to create these new regexes and not reuse TillRegex? Does it cause unintended side-effects?
Having these partially duplicated regexes makes maintenance harder and also further diverges from the other platforms.
It would be preferable to either treat it at the original regex, or change the extractor/parsing code to account for the article where needed.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -377,8 +377,18 @@ private DateTimeResolutionResult ParseComplexDatePeriod(string text, DateObject
 
                         if (endResolution.Success)
                         {
-                            futureEnd = ((Tuple<DateObject, DateObject>)endResolution.FutureValue).Item1;
-                            pastEnd = ((Tuple<DateObject, DateObject>)endResolution.PastValue).Item1;
+                            // When the second group refers to the 'end of' a period, the end resolution of the period should be used.
+                            if (match.Groups["EndOf"].Success)
```
tellarin, line 381: It seems we'd also need a little more general solution to support "start of" and the like.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
LionbridgeCSII: I was worried of possible side effects in modifying TillRegex directly. But now I have updated it and the tests do not show any issue.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
LionbridgeCSII: generalized solution to take into account the modifiers "start of" and "middle of". 
Also applied it to both 'start' and 'end' groups of the range.

tellarin (approved):

tellarin:
@LionbridgeCSII can you resolve the conflict to merge? Thanks.

LionbridgeCSII:
@tellarin, conflicts solved

Issue #1316, opened 2019-01-15
[EN DateTimeV2] In the query { when is the next new year day }, “new year day” is resolved as 2019-01-01


jiayinge:
**Describe the bug**
In the query { when is the next new year day }, “new year day” is resolved as 2019-01-01.

**To Reproduce**

**Expected behavior**

**Sample input/output**

**Platform (please complete the following information):**

**Additional context**

PR #2236, opened 2020-08-10
[EN DateTimeV2] support for relative holiday (#1316)

LionbridgeCSII:
Added support for relative holidays like "next new year day" (#1316).
Relevant test cases added to DateTimeModel


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -221,9 +221,15 @@ public static class DateTimeDefinitions
...
+      public const string HolidayList1 = @"(?<holiday>mardi gras|(washington|mao)'s birthday|juneteenth|(jubilee|freedom)(\s+day)|chinese new year|(new\s+(years'|year\s*'s|years?)\s+eve)|(new\s+(years'|year\s*'s|years?)(\s+day)?)|may\s*day|yuan dan|christmas eve|(christmas|xmas)(\s+day)?|black friday|yuandan|easter(\s+(sunday|saturday|monday))?|clean monday|ash wednesday|palm sunday|maundy thursday|good friday|white\s+(sunday|monday)|trinity sunday|pentecost|corpus christi|cyber monday)";
+      public static readonly string HolidayRegex1 = $@"\b{HolidayList1}(\s+(of\s+)?({YearRegex}|{RelativeRegex}\s+year))?\b";
+      public static readonly string HolidayRegex1CS = $@"\b({StrictRelativeRegex}\s+)?{HolidayList1}(?(1)|(\s+(of\s+)?({YearRegex}|{RelativeRegex}\s+year))?)\b";
```
tellarin, line 226: What does the CS here mean?
And wouldn't it be better to then combine all separate lists into one single regex to match?
It seems perf of the current change would re-scan the input multiple times.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishHolidayExtractorConfiguration.cs
```diff
@@ -11,13 +11,13 @@ public class EnglishHolidayExtractorConfiguration : BaseDateTimeOptionsConfigura
             new Regex(DateTimeDefinitions.YearRegex, RegexFlags);
 
         public static readonly Regex H1 =
-            new Regex(DateTimeDefinitions.HolidayRegex1, RegexFlags);
+            new Regex(DateTimeDefinitions.HolidayRegex1CS, RegexFlags);
```
tellarin, line 14: I guess these should be combined at the resources level. Do you see a negative side-effect of matching only once? Or a benefit of keeping multiple regexes?

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
LionbridgeCSII: CS stands for Conditional Syntax which however is not supported in java, so the new regex is only used in .NET.
Ok, lists combined into a single regex.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishHolidayExtractorConfiguration.cs
LionbridgeCSII: no I do not see negative side-effects, I have updated the file to use a single combined regex.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
tellarin: Is this supported in python/java? If not, then perhaps we should re-design it and change the base code too, not solve only at the regex level.

LionbridgeCSII: It should be supported in python but not in java/javascript. But it should be possible to rewrite the regex without conditional, so that it is supported everywhere. 
I will try this solution and to modify the configuration files in python/java to use a single regex as in .net.

Issue #2226, opened 2020-08-02
[IT Number] String "tr" recognized as number zero


MaxiPigna:
If you execute the instruction `NumberRecognizer.RecognizeNumber("tr", Culture.Italian)` you expect no result since "tr" doesn't resemble any number. Instead, it returns zero as result.

The same it happens also on LUIS, where "tr" is recognized as a number entity.
![image](https://user-images.githubusercontent.com/33937977/89120103-d0784980-d4b3-11ea-8b10-a396c2abe527.png)

**To Reproduce**
NumberRecognizer.RecognizeNumber("tr", Culture.Italian)[0].Resolution.First().Value
"0"

**Expected behavior**
NumberRecognizer.RecognizeNumber("tr", Culture.Italian).Any()
false

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: nuget package, LUIS
 - Version of package v1.4.0

tellarin:
Thanks for the catch. A fix is on its way.

MaxiPigna:
Great, thanks! When you will release the new nuget package version, is the update immediately published also on LUIS?

tellarin:
@MaxiPigna, not automatically. But the fix is merged and the next package release should enter the LUIS pipeline soon.
Thanks for the bug report.

PR #2240, opened 2020-08-12
[IT Number] fix String "tr" recognized as number zero (#2226)

LionbridgeCSII:
Fixes instances of elided numbers being wrongly extracted (#2226).
For example 'settant' should be extracted only when occurring in compound numbers e.g. 'settantotto' (seventy-eight) or in elided form e.g. 'settant'anni' (seventy years).
Relevant test cases added to NumberModel.

### DIFF
--- a/Specs/Number/Italian/NumberModel.json
+++ b/Specs/Number/Italian/NumberModel.json
@@ -1847,5 +1847,28 @@
         }
       }
     ]
+  },
+  {
+    "Input": "tr",
+    "NotSupportedByDesign": "javascript,python,java",
+    "Results": []
+  },
+  {
+    "Input": "ha compiuto settant'anni",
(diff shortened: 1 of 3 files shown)


Specs/Number/Italian/NumberModel.json
```diff
@@ -1847,5 +1847,28 @@
...
+        "Text": "settant",
+        "TypeName": "number",
+        "Resolution": {
+          "value": "70"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "ha compiuto settant anni",
```
tellarin, line 1870: This is not a regular form, right? If the purpose here is to show this is not extracted, let's add a Comment attribute in the spec describing it.

LionbridgeCSII: comment added

Issue #2231, opened 2020-08-04
[EN DateTimeV2] 1am is resolved to 13:00 (Python)


godfatherofdevil:
text = "tonight at 1am" - the date part is resolved correctly however the time part is resolved to 13:00

here is the code snippet to reproduce it:

`from recognizers_date_time import DateTimeRecognizer
from recognizers_date_time import Culture
recognizer = DateTimeRecognizer(Culture.English)
model = recognizer.get_datetime_model()
text = "tonight at 1am"
result = model.parse(text)
for r in result:
  print(r.text, r.resolution)`

**Expected behavior**
1am should be resolved to 01:00 and not 13:00

**Sample input/output**
input text = "tonight at 1am"
output : `tonight at 1am {&apos;values&apos;: [{&apos;timex&apos;: &apos;2020-08-04T13&apos;, &apos;type&apos;: &apos;datetime&apos;, &apos;value&apos;: &apos;2020-08-04 13:00:00&apos;}]}`
As applicable, add examples of text input and json output to help explain the problem.

**Platform (please complete the following information):**
 - Platform:  Python
 - Environment: python 3.8 REPL
 - Version of package : recognizers-text-date-time        1.0.2a2

**Additional context**

PR #2242, opened 2020-08-13
[EN DateTimeV2] Fix for 1am being wrongly resolved to 13:00 and similar (#2231)

LionbridgeCSII:
Fix for cases like "tonight at 1am" where "1am" is resolved to 13:00 instead of 01:00 (#2231).
The solution has been implement in all frameworks.
Updated 1 test case in MergedParsed ("Add 649 midnight tonight") which had the wrong resolution 12:00 instead of 00:00.


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -16975,5 +16975,74 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I'll go back tonight at 2am",
```
tellarin, line 16980: Can you also add a spec case for the exact sentence reported in the issue?
It can come in a future PR. Thanks.

tellarin:
@LionbridgeCSII, can you resolve the conflict and add that spec case mentioned in the review?

LionbridgeCSII:
> @LionbridgeCSII, can you resolve the conflict and add that spec case mentioned in the review?

@tellarin, conflict solved and test case added.

Issue #2054, opened 2020-03-23
[EN DateTimeV2] "Till date" is not currently recognized as datetimerange


vatyagga:
**Describe the bug**
"Till date" in the input is not recognized as a datetime entity. 

**To Reproduce**
Sample question:
How many Azure clusters docked till date? 

**Expected behavior**
Till date is expected to be parsed by LUIS as a datetime

PR #2245, opened 2020-08-18
[EN DateTimeV2] Added support for "To/Till date" (#2054)

LionbridgeCSII:
Added support for expressions like "to/till date" in DatePeriod (#2054).


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -190,6 +190,7 @@ public static class DateTimeDefinitions
       public static readonly string TimeNumberCombinedWithUnit = $@"\b(?<num>\d+(\.\d*)?){TimeUnitRegex}";
       public static readonly string[] BusinessHourSplitStrings = { @"business", @"hour" };
       public const string NowRegex = @"\b(?<now>(right\s+)?now|as soon as possible|asap|recently|previously)\b";
+      public static readonly string NowDateRegex = $@"\b({NowRegex}|^(date)$)\b";
```
tellarin, line 193: Perhaps rename this to "NowParseRegex"?
Also, wouldn't it be best to treat "^(date)$" as a named group to makes the parse behaviour explicit, as in onewordperiodregex? I suggest adding a comment on the YAML that this component wouldn't need to be translated .

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
LionbridgeCSII: Renamed regex and added comment.
But I do not see the reason to use a named group here, "date" does not require a specific parse processing, it is handled as the other elements of NowRegex.

tellarin: Not a major issue. The point was to clarify for those localizing the regexes to other languages that this is English-specific and not exactly the same as "now".

tellarin: I'm fine with merging as-is once conflicts are resolved.

tellarin (approved):
Only pending resolving conflicts.

LionbridgeCSII:
> Only pending resolving conflicts.

solved

Issue #2011, opened 2019-12-27
[EN DateTimeV2] "end of" related bugs


wlifuture:
issues with e.g. "end of this month", "before end of this year", "by end of this year", etc.
1. "by end of this month": [end, 2019-12-16];
2. "before end of this year": [end, 2019-07-01];
3. "end of this year": with [start, 2019-07-01];
etc.

PR #2247, opened 2020-08-19
[EN DateTimeV2] Fix for "end of" related bugs (#2011)

LionbridgeCSII:
Fixed resolution for the first and second patterns pointed out in #2011. 
The resolution of the third pattern "end of this year" does not seem wrong (but as pointed out in #2232, there is a discrepancy in the start value respect to "end of 2016").
Also fixed analogous issue for patterns like "after the beginning of March".

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -17163,5 +17163,82 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I'll go back by end of this month",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupported": "javascript, python, java",
+    "Results": [
+      {
+        "Text": "by end of this month",
+        "Start": 13,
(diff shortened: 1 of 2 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -17067,5 +17067,115 @@
...
+              "start": "2016-07-01",
+              "end": "2017-01-01"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "I'll go back after the beginning of March",
```
tellarin, line 17149: If this is supported, why not "before the end of December"?

LionbridgeCSII: "before the end of December" is also supported, added test case.

tellarin: Sorry. My misunderstanding. Thanks for adding the case.

tellarin (approved):

Issue #1820, opened 2019-08-23
[EN Dimension] 5' isn't recognized.


RaysceneNS:
English feet dimensions writen as a number followed by the single prime character (') are not recognized.

**To Reproduce**
> I am 5' tall

**Expected behavior**
Expected the 5' to be recognized as type=dimension, unit=Foot, value=5.

PR #2250, opened 2020-08-21
[EN Dimension] added ' to Foot symbols (#1820)

LionbridgeCSII:
fix for issue #1820, 5' to be recognized as type=dimension, unit=Foot, value=5.
Test case added to DimensionModel


tellarin (changes requested):
Do you have an estimate to handle the mentioned "quote" issue?

LionbridgeCSII:
> Do you have an estimate to handle the mentioned "quote" issue?
I have not looked into the "quote" issue yet, I was trying to tackle the python failure first. The problem there is that list elements in the yaml file are converted to raw strings, but while " is correctly converted to r'"', ' becomes r'\''. 
Do you have any suggestion on how to solve it?

tellarin:
> > Do you have an estimate to handle the mentioned "quote" issue?
> 
> I have not looked into the "quote" issue yet, I was trying to tackle the python failure first. The problem there is that list elements in the yaml file are converted to raw strings, but while " is correctly converted to r'"', ' becomes r'\''.
> Do you have any suggestion on how to solve it?

This would also require a change in the Python code generator. We can add it once the .NET version is finalized.

LionbridgeCSII:
@tellarin, added regex to handle the quote issue (merged with the existing NonUnitRegex)

Issue #2243, opened 2020-08-14
[EN DateTimeV2] resolution is None for text like "from around 1pm"


godfatherofdevil:
**Describe the bug**
resolution is None for text containing an unique combination of `from` and some time, some examples:
1. "i am feeling sick from around 1pm"
2. "I have pain and pressure in my lower abdomen, blurry vision and headache from around 2pm"

**To Reproduce**
`python
from recognizers_date_time import DateTimeRecognizer
from recognizers_date_time import Culture
recognizer = DateTimeRecognizer(Culture.English)
model = recognizer.get_datetime_model()
text = "I have pain and pressure in my lower abdomen, blurry vision and headache from around 2pm"
result = model.parse(text)
for r in result:
	print(r.text, r.resolution, r.start, r.end, r.type_name)
`

**Expected behavior**
if the text is "from around 2pm today" then resolution is correct, should have the same behavior in above case.

**Sample input/output**
text = "I have pain and pressure in my lower abdomen, blurry vision and headache from around 2pm"
resolution = None
another_text = "i am feeling sick from around 1pm"
resolution = None
yet_another_text = "i will be back around 1am today"
resolution = {'values': [{'timex': '2020-08-14T13', 'Mod': 'since', 'type': 'datetimerange', 'sourceEntity': 'datetimepoint', 'start': '2020-08-14 13:00:00'}]}

**Platform (please complete the following information):**
 - Platform: Python
 - Environment: Python REPL
 - Version of package: 1.0.2a2

**Additional context**

godfatherofdevil:
(.venv38) avinash@avinash-ThinkPad-E490:~/repos/MMC$ python
Python 3.8.2 (default, Mar 21 2020, 16:51:37) 
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from recognizers_date_time import DateTimeRecognizer
>>> from recognizers_date_time import Culture
>>> recognizer = DateTimeRecognizer(Culture.English)
>>> model = recognizer.get_datetime_model()
>>> text = "I have pain and pressure in my lower abdomen, blurry vision and headache from around 2pm"
>>> result = model.parse(text)
>>> for r in result:
...   print(r.text, r.resolution)
... 
around 2pm None
>>> for r in result:
...   print(r.text, r.resolution, r.start, r.end, r.type_name)
... 
around 2pm None 78 87 datetimeV2.time
>>> another_text = "i will be back around 1 am today"
[...]

PR #2252, opened 2020-08-25
[EN DateTimeV2] Added support for text like "from around 1pm" (#2243)

LionbridgeCSII:
added support in .NET for cases like "from around 1pm" (#2243).

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
@@ -235,10 +235,10 @@ namespace Microsoft.Recognizers.Definitions.English
       public const string RecentlyTimeRegex = @"(recently|previously)";
       public const string AsapTimeRegex = @"(as soon as possible|asap)";
       public const string InclusiveModPrepositions = @"(?<include>((on|in|at)\s+or\s+)|(\s+or\s+(on|in|at)))";
-      public const string AroundRegex = @"(?:\b(?:around|circa)\s*\b)";
-      public static readonly string BeforeRegex = $@"((\b{InclusiveModPrepositions}?(?:before|in\s+advance\s+of|prior\s+to|(no\s+later|earlier|sooner)\s+than|ending\s+(with|on)|by|(un)?till?|(?<include>as\s+late\s+as)){InclusiveModPrepositions}?\b\s*?)(?<approx>\s*{AroundRegex})?|(?<!\w|>)((?<include><\s*=)|<))(\s+the)?";
-      public static readonly string AfterRegex = $@"((\b{InclusiveModPrepositions}?((after|(starting|beginning)(\s+on)?(?!\sfrom)|(?<!no\s+)later than)|(year greater than))(?!\s+or equal to){InclusiveModPrepositions}?\b\s*?)(?<approx>\s*{AroundRegex})?|(?<!\w|<)((?<include>>\s*=)|>))(\s+the)?";
-      public static readonly string SinceRegex = $@"(?:(?:\b(?:since|after\s+or\s+equal\s+to|starting\s+(?:from|on|with)|as\s+early\s+as|(any\s+time\s+)?from)\b\s*?)(?<approx>\s*{AroundRegex})?|(?<!\w|<)(>=))(\s+the)?";
+      public const string AroundRegex = @"(?:\b(?:around|circa)\s*?\b)(\s+the)?";
+      public static readonly string BeforeRegex = $@"((\b{InclusiveModPrepositions}?(?:before|in\s+advance\s+of|prior\s+to|(no\s+later|earlier|sooner)\s+than|ending\s+(with|on)|by|(un)?till?|(?<include>as\s+late\s+as)){InclusiveModPrepositions}?\b\s*?)|(?<!\w|>)((?<include><\s*=)|<))(\s+the)?";
+      public static readonly string AfterRegex = $@"((\b{InclusiveModPrepositions}?((after|(starting|beginning)(\s+on)?(?!\sfrom)|(?<!no\s+)later than)|(year greater than))(?!\s+or equal to){InclusiveModPrepositions}?\b\s*?)|(?<!\w|<)((?<include>>\s*=)|>))(\s+the)?";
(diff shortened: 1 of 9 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -233,10 +233,10 @@ public static class DateTimeDefinitions
...
-      public static readonly string BeforeRegex = $@"((\b{InclusiveModPrepositions}?(?:before|in\s+advance\s+of|prior\s+to|(no\s+later|earlier|sooner)\s+than|ending\s+(with|on)|by|(un)?till?|(?<include>as\s+late\s+as)){InclusiveModPrepositions}?\b\s*?)|(?<!\w|>)((?<include><\s*=)|<))(\s+the)?";
-      public static readonly string AfterRegex = $@"((\b{InclusiveModPrepositions}?((after|(starting|beginning)(\s+on)?(?!\sfrom)|(?<!no\s+)later than)|(year greater than))(?!\s+or equal to){InclusiveModPrepositions}?\b\s*?)|(?<!\w|<)((?<include>>\s*=)|>))(\s+the)?";
-      public const string SinceRegex = @"(?:(?:\b(?:since|after\s+or\s+equal\s+to|starting\s+(?:from|on|with)|as\s+early\s+as|(any\s+time\s+)?from)\b\s*)|(?<!\w|<)(>=))(the)?";
       public const string AroundRegex = @"(?:\b(?:around|circa)\s*\b)";
+      public static readonly string BeforeRegex = $@"((\b{InclusiveModPrepositions}?(?:before|in\s+advance\s+of|prior\s+to|(no\s+later|earlier|sooner)\s+than|ending\s+(with|on)|by|(un)?till?|(?<include>as\s+late\s+as)){InclusiveModPrepositions}?\b\s*?)(?<approx>\s*{AroundRegex})?|(?<!\w|>)((?<include><\s*=)|<))(\s+the)?";
```
tellarin, line 237: Why add it to the relative regexes and not have a separate match by itself, before the code that deals with the relative matches?
I worry a bit that the current solution mixes the semantics of the relative regexes with an unrelated concept.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
LionbridgeCSII: Added it to the relative regexes in order to modify the base code as little as possible (the same kind of approach is used with the InclusiveModPrepositions). It did not seem problematic because these regexes are only used in Merged and DateTimePeriod (e.g. "today before around 4pm") where this kind of pattern seems legit. 
But yes, I can add new regexes explicitly for these patterns and keep them separated from the relative regexes if you think it is better.

tellarin: Can you please add as separate regexes?

LionbridgeCSII: added as separate regexes

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -327,7 +327,23 @@ private List<ExtractResult> AddMod(List<ExtractResult> ers, string text)
         {
             foreach (var er in ers)
             {
-                var success = TryMergeModifierToken(er, config.BeforeRegex, text);
+                var success = TryMergeModifierToken(er, config.BeforeAroundRegex, text);
```
tellarin, line 330: Why do we need to match both BeforeAroundRegex and BeforeRegex later? This is basically the same as before and there's no benefit in keeping both (it would be better to keep combined).
Can't we just match on around at the entity level, then add mod would only match on before/after/etc, as the method name implies. This would also allow just the around + datetime form to always work, even if no relative prefix.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
LionbridgeCSII: Ah, I see what you mean now, yes, it makes more sense. I thought this was what you implied when saying to keep the regexes separated. I will update the code.

LionbridgeCSII: code updated, 'around' modifier is still matched in Merged instead that at the entity level, but the other modifiers (before/after...) are now allowed to match also when 'around' is found. In this way, both around mod ('approx') and mixed mods (e.g. 'before-approx') can be caught.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -327,6 +327,7 @@ private List<ExtractResult> AddMod(List<ExtractResult> ers, string text)
         {
             foreach (var er in ers)
             {
+                TryMergeModifierToken(er, config.AroundRegex, text);
                 var success = TryMergeModifierToken(er, config.BeforeRegex, text);
```
tellarin, line 331: This should happen always even if the previous succeeded, right?

Can you please add a code comment saying this and that it then will cascade on the relative modifiers?

tellarin: The comment will make the code intent clearer. Otherwise new people reading the code will think it's incorrect.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
LionbridgeCSII: comment added

tellarin (approved):

tellarin:
@LionbridgeCSII please keep an eye out for PRs with conflicts, ok? We can merge as soon as these are resolved.
Thanks.

LionbridgeCSII:
@tellarin, conflicts solved

Issue #450, opened 2018-04-11
[FR DateTimeV2] Can not get hours duration


thomasryck:
I have tried different syntax in order to get a duration in hour and I never managed to get one. Each time I do a request with a duration in hour as " pour 3 heures ", "pour une durée de 3 heures", I get as a result of the parsing a time or, depending on the context, a datetime which contain the duration. This issue only exist with hours from the different tests I did. I also think that the issue on duration provokes issue on datetime since when I say "le 17 pour 3 heures", the parsed entity is "17 pour 3 heures".  And I think that the parsing sould return a date and a duration and not a datetime.

tellarin:
@LionbridgeCSII, please take a look at this one and try to fix/implement similarly to English.

PR #2256, opened 2020-08-27
[FR DateTimeV2] Fix to get hours duration (#450)

LionbridgeCSII:
Fix to avoid extracting expressions like "3 heures" as time when preceded by "pour" or "durée de". The problem is "3 heures" is ambiguous in general and can also represent a time e.g. "il est 3 heures" (it is 3 o'clock).
Fix implemented only in .NET, test cases added to DateTimeModel.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
@@ -128,8 +128,10 @@ namespace Microsoft.Recognizers.Definitions.French
       public static readonly string TimeRegex2 = $@"(\b{TimePrefix}\s+)?(t)?{BaseDateTime.HourRegex}(\s*)?:(\s*)?{BaseDateTime.MinuteRegex}((\s*)?:(\s*)?{BaseDateTime.SecondRegex})?((\s*{DescRegex})|\b)";
       public static readonly string TimeRegex3 = $@"\b{BaseDateTime.HourRegex}\.{BaseDateTime.MinuteRegex}(\s*{DescRegex})(\s+{TimePrefix})?";
       public static readonly string TimeRegex4 = $@"\b{BasicTime}(\s*{DescRegex})?(\s+{TimePrefix})?\s+{TimeSuffix}\b";
+      public static readonly string TimeRegex4CS = $@"(?<=(?<DurationPrep>(pour|durée de)\s+)|){TimeRegex4}(?(DurationPrep)(?<!heures?))";
       public static readonly string TimeRegex5 = $@"\b{BasicTime}((\s*{DescRegex})(\s+{TimePrefix})?|\s+{TimePrefix})";
       public static readonly string TimeRegex6 = $@"{BasicTime}(\s*{DescRegex})?\s+{TimeSuffix}\b";
+      public static readonly string TimeRegex6CS = $@"(?<=(?<DurationPrep>(pour|durée de)\s+)|){TimeRegex6}(?(DurationPrep)(?<!heures?))";
       public static readonly string TimeRegex7 = $@"\b{TimeSuffix}\s+[àa]\s+{BasicTime}((\s*{DescRegex})|\b)";
       public static readonly string TimeRegex8 = $@"\b{TimeSuffix}\s+{BasicTime}((\s*{DescRegex})|\b)";
       public static readonly string TimeRegex9 = $@"\b{PeriodHourNumRegex}\s+{FivesRegex}((\s*{DescRegex})|\b)";
--- a/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
@@ -717,7 +717,7 @@ namespace Microsoft.Recognizers.Definitions.French
         };
       public static readonly Dictionary<string, string> AmbiguityTimeFiltersDict = new Dictionary<string, string>
         {
-            { @"heures?$", @"(pour|durée\s+de)\s*" }
+            { @"heures?$", @"\b(pour|durée\s+de)\s+(\S+\s+){1,2}heures?\b" }
         };
       public static readonly IList<string> MorningTermList = new List<string>
         {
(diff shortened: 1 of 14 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -128,8 +128,10 @@ public static class DateTimeDefinitions
       public static readonly string TimeRegex2 = $@"(\b{TimePrefix}\s+)?(t)?{BaseDateTime.HourRegex}(\s*)?:(\s*)?{BaseDateTime.MinuteRegex}((\s*)?:(\s*)?{BaseDateTime.SecondRegex})?((\s*{DescRegex})|\b)";
       public static readonly string TimeRegex3 = $@"\b{BaseDateTime.HourRegex}\.{BaseDateTime.MinuteRegex}(\s*{DescRegex})(\s+{TimePrefix})?";
       public static readonly string TimeRegex4 = $@"\b{BasicTime}(\s*{DescRegex})?(\s+{TimePrefix})?\s+{TimeSuffix}\b";
+      public static readonly string TimeRegex4CS = $@"(?<=(?<DurationPrep>(pour|durée de)\s+)|){TimeRegex4}(?(DurationPrep)(?<!heures?))";
```
tellarin, line 131: Would it be possible to solve without conditional statements? These are not supported in regexes in some of the other target platforms, right?
Also, couldn't this be solved by some ambiguity filter at the time level to discard extracted times if the prefix is present?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
tellarin: Duplicating regexes (one for .NET one for the rest) also introduces extra maintainability issues.

LionbridgeCSII: solving it without conditional statement at the regex level seems much more complicate and the use of lookbehinds (which are also not supported everywhere) seems unavoidable. 
An ambiguity filter in the TimeExtractor seems a more viable solution, I will try it (I have discarded this approach because it means modifying the base code for a language specific problem which is also not recommended).

tellarin: There is a generic mechanism for ambiguity filters already and it's usually applied "at the end" of a type.
The filter dictionary itself stays in the language specific resource file.

tellarin: Which platform doesn't support lookbehinds?

LionbridgeCSII: python supports only fixed width lookaround, and also java and javascript have some limitations (I encountered problems with lookarouds of infinite width or with multiple quantifiers).

LionbridgeCSII: added method FilterAmbiguity to BaseTimeExtractor and removed conditional regexes.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
```diff
@@ -163,5 +167,26 @@ private List<Token> SpecialCasesRegexMatch(string text, DateObject reference)
...
+            if (this.config.AmbiguityFiltersDict != null)
+            {
+                foreach (var regex in this.config.AmbiguityFiltersDict)
+                {
+                    foreach (var extractResult in extractResults)
+                    {
+                        if (regex.Key.IsMatch(extractResult.Text))
+                        {
+                            var matches = regex.Value.Matches(text).Cast<Match>();
+                            extractResults = extractResults.Where(er => !matches.Any(m => m.Index + m.Length == er.Start))
```
tellarin, line 182: The condition here is too dependent on the specific regex. People changing the resources file will not be aware of the criteria here. Ideally it should be only a check if a match is within the other.

Patterns/French/French-DateTime.yaml
```diff
@@ -1055,6 +1055,11 @@ AmbiguityFiltersDict: !dictionary
   entries:
     '^([eé]t[eé])$': '(?<!((l\s*[''`]\s*)|(cet(te)?|en)\s+))[eé]t[eé]\b'
     '^(mer)$': '(?<!((le|ce)\s+))mer\b'
+# Used to exclude duration matches (e.g. "pour 3 heures") from time extractions
+AmbiguityTimeFiltersDict: !dictionary
+  types: [ string, string ]
+  entries:
+    'heures?$': '(pour|durée\s+de)\s*'
```
tellarin, line 1062: Shouldn't the second regex check for the <number heures> part too? Otherwise if there are multiple "heures".
Similar to the general filter a few lines above.

Specs/DateTime/French/DateTimeModel.json
```diff
@@ -2465,5 +2465,121 @@
...
+            }
+          ]
+        },
+        "Start": 20,
+        "End": 27
+      }
+    ]
+  },
+  {
+    "Input": "Je vais partir pour une durée de 3 heures",
```
tellarin, line 2494: Can you add two more cases?
One sentence that only has a time entity using "heures" and another one that has both a time and a duration.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
LionbridgeCSII: Agreed, however here it is crucial to check that the match and the extractResult are concatenated. One could change the second regex in AmbiguityTimeFiltersDict to '(pour|durée\s+de)\s+(.+)' and then check if its match contains the extractResult, but then cases like "Je vais partir pour la mer à 3 heures" (I will leave for the sea at 3 o'clock) will also be treated as duration.

Specs/DateTime/French/DateTimeModel.json
LionbridgeCSII: cases added

Patterns/French/French-DateTime.yaml
LionbridgeCSII: It does not seem necessary given the current implementation of the filter because it excludes only the extractions ending with "heures" that are immediately preceded by "(pour|durée\s+de)"

tellarin: As discussed in the code file, the current approach there is too brittle. Either we need a way to represent the "contiguity" as part of the filter definition, or we need a filter regex that checks the key too.
Maybe something like (pour|durée\s+de)\s*(\S+){1,2}\sheures?

LionbridgeCSII: updated regex as suggested and reverted FilterAmbiguity to its original form (from BaseMergedDateTimeExtractor) which checks for overlapping matches.

tellarin (approved):

Issue #908, opened 2018-10-15
[EN-DatetimeV2] Wrongly recognized some of DateTime entities with "." splitter


guom08:
**Describe the bug**
For cases like "in 5.1935", it should be extracted as "in May 1935".
Now, only the "1935" part is extracted as "year 1935"

**To Reproduce**
Input "in 5.1935", will get output for only "1935" as a DateRange entity

**Expected behavior**
"5.1935" should only be extracted as "May 1935"
"5.11" can be extracted as "Nov 5th" or "May, 2011" or time "5:11", this case needs to be distinguished by special prepositions.
    a. If someone say "on 5.11", it should be "on Nov 5th".
    b. If someone say "in/during 5.11", it should be recognized as "in/during May, 2011".
    c. If someone say "at 5.11", it should be recognized as "at 5:11"

To avoid possible confusion, for two digit year, we require certain prepositions to indicate a Date or DateRange. We will not extract expression with two digit year but without prepositions as any DateTime entity.

**Sample input/output**
"5.1935" -> "May 1935"
"5.11" -> null
"on 5.11" -> "Nov 5th"
"in/during 5.11" -> "May, 2011"
"at 5.11" -> "5:11"
"on 5.15" -> null
"at 5.85" -> null
"in/during 5.85" -> "May, 1985"

tellarin:
I suggest we don't cover all these now. Let's focus on day.month and day.time only. For the others, just make sure the year part is not extracted by itself.

PR #2260, opened 2020-09-02
[EN DateTimeV2] Fix for wrongly recognized DateTime entities with "." splitter (#908)

LionbridgeCSII:
Fix for issues reported in #908. 

We did not update the pattern "on 5.11" which is currently parsed as "May 11th" instead of "Nov 5th" (as requested in the issue).   It seems to be a breaking change because there are test cases that rely on the current implementation e.g.:
"Cortana, please set up a Skype call sometime on Friday 7.6 with Jim." -> "July 6th"
"This task should be done on 5.12" -> "May 12th"


tellarin (approved):

tellarin:
@LionbridgeCSII Same here. Thanks.

LionbridgeCSII:
> @LionbridgeCSII Same here. Thanks.

resolved.

tellarin:
@LionbridgeCSII Sorry. Simple conflicts again.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19968,5 +19968,96 @@
...
+  },
+  {
+    "Input": "I'll go back on 5.1935",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "Results": []
+  },
+  {
+    "Input": "I'll go back 5.11",
```
tellarin, line 20008: Can you add a comment here that this is not supproted as the preposition specifies the interpretation and without it it doesn't make sense?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19968,5 +19968,96 @@
...
+              "type": "date",
+              "value": "2017-05-11"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "I'll go back on 5.1935",
```
tellarin, line 20001: Can you add a comment in this one to say that in the future it could be supported as "May-1935"?
Can you also add a spec with some numbers like "0.1993 is not a date" and a spec like "12.12.12.12"?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19968,5 +19968,96 @@
...
+  },
+  {
+    "Input": "I'll go back 5.11",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "Results": []
+  },
+  {
+    "Input": "I'll go back during 5.11",
```
tellarin, line 20015: I think we can remove this one.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19968,5 +19968,96 @@
...
+              "type": "time",
+              "value": "17:11:00"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "I'll go back at 5.85",
```
tellarin, line 20050: Please reword this to "The price ended the day at 5.85".

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19968,5 +19968,96 @@
...
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "Results": []
+  },
+  {
+    "Input": "I'll go back in 5.85",
+    "Context": {
```
tellarin, line 20058: Please, add the same comment as in "5.1935" as in the future this could be interpreted as a date.

tellarin (changes requested):
Sorry, took a closer look at the specs and it would be good to add some info in them.

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: done

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: done and test cases added.

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: removed.

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: done

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: added

Issue #2262, opened 2020-09-02
[EN DateTimeV2] Specific datetimeperiod phrase not extracted


LionbridgeCSII:
**Describe the bug**

Expressions like "between 7 and 9" and "between 7 and 9 last night" are not recognized as DateTimePeriod, instead, in the second case, only "9 last night" is extracted as DateTime.

**Expected behavior**
"between 7 and 9" and "between 7 and 9 last night" should be extracted as a DateTimePeriod.

PR #2264, opened 2020-09-07
[EN DateTimeV2] Fix for specific datetimeperiod phrase not extracted (#2262)

LionbridgeCSII:
Fix for DateTimePeriods containing pure numbers (e.g. "between 7 and 9 last night") not being extracted (#2262).

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
@@ -756,6 +756,8 @@ namespace Microsoft.Recognizers.Text.DateTime
             return new Token(startOut, endOut);
         }
 
+        // The method matches pure number ranges. It is used inside MatchTimeOfDay, so the condition IsNullOrWhiteSpace(midStr) implies
+        // that the range must be contiguous to a TimeOfDay expression (e.g. "last night from 7 to 9").
         private List<Token> MatchPureNumberCases(string text, Token tok, bool before)
         {
             var ret = new List<Token>();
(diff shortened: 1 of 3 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
```diff
@@ -736,5 +756,30 @@ private Token MatchWithinNextPrefix(string subStr, string text, Token duration,
             return new Token(startOut, endOut);
         }
 
+        private List<Token> MatchPureNumberCases(string text, Token tok, bool before)
```
tellarin, line 759: What exactly is happening here? Are you taking advantage of the fact that a range with pure number is detected if at the end of the sentence? Is that it?
Can you add a comment to the method describing the assumption?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -17680,5 +17678,30 @@
         }
       }
     ]
+  },
+  {
+    "Input": "I'll be out next evening from 7 to 9",
+    "Context": {
```
tellarin, line 17684: Could you add a morning case, also using a different sentence structure (i.e., not "i'll be out ...")? Thanks.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimePeriodExtractor.cs
LionbridgeCSII: I dropped the requirement that the pure number range must be at the end of sentence when it is contiguous to a time of day match. Will add comment.

LionbridgeCSII: comment added

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: added text case with 'morning'

tellarin (approved):

Issue #2263, opened 2020-09-03
[EN DateTimeV2] Missing detection for "weekend" or "weekends" unless preposition is present


dunalduck0:
Recognizer can detect entities of "weekend" or "weekends" if they are with preposition, but fail to detect if no preposition is present. Below are examples of successful and failed cases.

Successful detection examples:
“dinner with mom this weekend”  "timex": "2020-W36-WE"
“dinner with mom next weekend”  "timex": "2020-W37-WE"

Failed detection examples:
“dinner with mom weekend”   no entity is found

It's not uncommon to omit preposition when users type in casual text. Can you please add detection in such cases?

tellarin:
@LionbridgeCSII, can you take a look? We can discuss a design by e-mail if necessary.

PR #2265, opened 2020-09-07
[EN DateTimeV2] Fix for missing detection for "weekend" unless preposition is present (#2263)

LionbridgeCSII:
removed "weekend" from UnspecificDatePeriodRegex used by FilterUnspecificDatePeriod in BaseMergedDateTimeExtractor in order to keep the extraction "weekend" when preposition is missing e.g. “dinner with mom weekend” (#2263).

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -17751,7 +17751,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-11T00:00:00"
     },
-    "NotSupported": "javascript, python",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "any weekend",
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -14808,6 +14808,21 @@
             }
           ]
         }
+      },
+      {
+        "Text": "any year",
+        "Start": 19,
+        "End": 26,
+        "TypeName": "datetimeV2.set",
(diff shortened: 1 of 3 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -17680,5 +17680,30 @@
...
+    },
+    "NotSupported": "javascript, python",
+    "Results": [
+      {
+        "Text": "weekend",
+        "Start": 16,
+        "End": 22,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
```
tellarin, line 17697: As this is not explicit. Shouldn't we return a past and a future resolution like in other places? Why is this not happening?
Can you add cases for sentences with the "the" article too. Or other cases?
BTW, "any weekend" and "all weekends" should be supported as Set. Can we add this in the same PR?

tellarin (changes requested):

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: Past and future resolutions are returned in DatePeriodParser (with identical values) but not in DateTimeModel for cases like "(this|the|next) (week|weekend|month|year)".

LionbridgeCSII: added test case "the weekend", and support for 'any' + [Unit] in Set. 
Support for 'all' + [Unit] will come later because additional work is needed to check that the unit is plural.

tellarin (approved):

Issue #1850, opened 2019-09-04
[* Currency] Currency not recognized when used before value... eg.  USD 200


antonn:
**Describe the bug**
200 USD --> Gets recognized as Money with both isoCurrency and Value.
USD 200 --> here only 200 is getting recognized as Number. 

**To Reproduce**
1. Create prebuilt entities Money and Number
2. Type 200 USD in the Test interface. -> It works fine as Money gets recognized
3. Type USD 200 --> In this case only number. USD is not getting identified anywhere

**Expected behavior**
USD 200  should be recognized as Money with value 200 and isoCurrency USD

**Sample input/output**
I would like to buy this item for USD 200

**Platform (please complete the following information):**
 - LUIS English
 

**Additional context**

antonn:
Any update on when this will be fixed ?

tellarin:
Hi @antonn, it should come out in vNext+1. Tentatively planned for early next month.

antonn:
Hi tellarin

Any update on the date of fix to luis.ai for this issue reported?

antonn:
Any further update on this?

tellarin:
Hi @antonn,
this is in the near-term plans, but there is no ETA yet.

PR #2266, opened 2020-09-08
[* Currency] Fix for currency 'USD' not recognized when used before value (#1850)

LionbridgeCSII:
Added 'usd' to CurrencyPrefixList in order to solve issue #1850. 
Should other currencies be added to the list?
Test case added to CurrencyModel.


tellarin (changes requested):
Unfortunately this is not a proper fix. We should have a mechanism where any existing currency code can be either a prefix or a suffix. Otherwise we depend on error-prone duplication of data there's already in the YAML file.
Perhaps some pre-processing step when building the structures at init?
CurrencyNameToIsoCodeMap has the code values. We should support all the regular codes plus the "_"{3_char_code} ones (removing the "_" for matching), but not the ones starting with underscore and less then 3 char codes.

LionbridgeCSII:
@tellarin, now CurrencyNameToIsoCodeMap is merged into CurrencyPrefixList in the currency configuration file (excluding the codes starting with underscore). The same could be done with CurrencySuffixList, but it seems that this list already contains the iso codes.

tellarin (approved):

Issue #2267, opened 2020-09-08
[ZH Number]"千分之一"、"仟分之贰"、"万分之三"、"萬分之肆" can't be recognized as number or percentage number


SoaringTiger:
**Describe the bug**
"千分之一"、"仟分之贰"、"万分之三"、"萬分之肆" can't be recognized as number or percentage number

**Expected behavior**
"千分之一" and "一千分之一" mean 0.001  (1‰) ， but "千分之一" can't be recognized as number or percentage number.
Similarly，
"仟分之贰" and "壹仟分之贰" mean 0.002  (2‰) 
"万分之三" and "一万分之三" mean 0.0003  (3‱）
"萬分之肆" and "壹萬分之肆" mean 0.0004  (4‱）
"百万分之五" and "一百万分之五" mean 5E-06 (0.5‱）
"千万分之六" and "一千万分之六" mean 6E-07 (0.06‱）
"万万分之七" and "一万万分之七" mean 7E-08 (0.007‱）
etc.

**Platform (please complete the following information):**
 - Platform: [.NET]
 - Version of package [v1.4.1]

tellarin:
Closing it as duplicate of #2257 and being addressed in PR #2269.

Issue #2257, opened 2020-08-28
[ZH Percent]  "十分之一" and also "*分之*" can't be recognized as a percentage number


LukeALee:
when I try to recognize percentage number from "北京地区十分之一二手房总价不超过350万，其中中心城区房源占比仅为0.49%", the recognize_percentage() method can only catch the number of "0.49%", but in Chinese I think *分之* is really a common usage which also means percentage. Good news is recognize_number() can catch them all. But I still think that is not enough.

tellarin:
Happens in multiple platforms. Not Python only. We'll prioritize.
Thanks for reporting!

PR #2269, opened 2020-09-09
[ZH Number|Percent] Improve support for “分之” in Number and Percent in Chinese (#2267 #2257)

iofu728:
## Summarization

1. Percentage module support "[十百千万]分之*". 
2. Number module support "\d[十百千万]分之\d点\d*".

## ⏳ The detail of fixing the bug

Refer to issue #2267 and #2257, the previous model only support "\d[百]分之*" in number module and can't deal "[十百千万]分之*" pattern.
And in the test processing, we find that the previous model also can't solve the "\d[十百千万]分之\d点\d*" pattern.

One more thing, there are some ambiguous cases in this pattern, like "百分之一二" and "百分之一二手房".

The bug which can't recognize the "分之" in percentage module due to the regex string only has "[百佰]" pattern.
And it can recognize the "\d[百]分之*" in the number module which means there is some specific numerical value in front of "[十百千万]分之" like the span “一百分之一”.

## ToDo:
1. Solve the ambiguous case in number module.


tellarin (changes requested):
Could you add all cases mentioned in the issues as new specs to make sure these don't fail in future changes?
Also, why are the two edited cases not supported in JS? Is there an specific issue?

tellarin (approved):

tellarin:
@ChopperMan33, thanks for the help reviewing it.

Issue #2097, opened 2020-05-08
[EN Unit Dimension] Include recognition of sub-types of dimension unit, such as weight, distance, volume


ashly1yeo:
**Is your feature request related to a problem? Please describe.**
Not a problem but it provides better definition of the Unit Dimension type to have the Weight, Distance, Speed, Volume and Length.

**Describe the solution you'd like**
Instead of recognizing "100 lbs" as Unit-Dimensions, it will be typed as Unit-Weight just like "36.4°C" is recognized as Unit-Temperature. 

**Describe alternatives you've considered**
None.

**Additional context**
None

tellarin:
Modifying the type would be considered a breaking change.
Would it be reasonable to return an extra attribute with the Dimension sub-type? This schema is already used for numbers. E.g.,

`
"TypeName": "number",
"Resolution": {
    "subtype": "fraction",
     ...
}
`

ashly1yeo:
Yes that is what I am proposing, sorry it is not clear.  

Unit-Weight is TypeName: Unit, Subtype: Weight

tellarin:
@joaopsousa FYI and tracking.

PR #2274, opened 2020-09-15
[EN Unit Dimension] Added resolution of sub-types of dimension unit (#2097)

LionbridgeCSII:
Added "subtype" attribute to Resolution field for Dimension units (#2097) in .NET.


tellarin (approved):
Looks OK. The only issue to propagate to all languages is adding sub-types in their specs?

tellarin:
Re-building PR.

LionbridgeCSII:
> Looks OK. The only issue to propagate to all languages is adding sub-types in their specs?

yes exactly, should I add them to the other languages?

tellarin:
> > Looks OK. The only issue to propagate to all languages is adding sub-types in their specs?
> 
> yes exactly, should I add them to the other languages?

Please do, but in another PR once this one merges. Thanks!

Issue #2270, opened 2020-09-09
[ZH Dimension] "12千米"、"2.3千克"  etc. dimension value is null.   Only 1 digit number dimension value can be recognized.


SoaringTiger:
**Describe the bug**
"12千米"、"2.3千克"  etc. dimension value is null。
Only 1 digit number dimension value can be recognized. Such as "1千克"、 "2千米".

**Expected behavior**
```
{
        "text": "2.3千克",
        "start": 0,
        "end": 3,
        "typeName": "dimension",
        "resolution": {
            "unit": "Kilogram",
            "value": "2.3"
        }
    }

{
        "text": "12千米",
        "start": 0,
        "end": 3,
        "typeName": "dimension",
        "resolution": {
            "unit": "Kilometer",
            "value": "12"
        }
    }
```
etc

**Sample input/output**
```
input: "1千克"
{
        "text": "1千克",
        "start": 0,
        "end": 2,
        "typeName": "dimension",
        "resolution": {
            "unit": "Gram",
            "value": "1000"
        }
    }

---
input:  "2.3千克"

{
        "text": "千克",
        "start": 3,
        "end": 4,
        "typeName": "dimension",
        "resolution": {
            "unit": "Kilogram",
            "value": null
        }
    }
```

**Platform (please complete the following information):**
 - Platform: [ .NET ...]
 - Version of package [v1.4.1]

PR #2280, opened 2020-09-18
[ZH NumberUnit|Dimension] Improve support for “*千克” in Number with Unit and Dimension in Chinese (#2270)

iofu728:
## Summarization
1. improve support kilo unit in Chinese like "千克", "千米". Refer to #2270 
2. improve support multi-dimension digits recognizer, like "12千米".

## ⏳ The detail of fixing the bug

In this part, "千" is a morpheme of the word "千克/千米/千磅" rather than a single token word.
Due to `{NotSingleRefgex}|{SingleRegex}` match the pattern like "2千" and ignore the next word is "米", the previous model can't recognizer the kilo unit pattern like“千克”, "千米".

So we add one new pattern which wants to match the kilo unit, we call "KiloEndRegex".


Patterns/Chinese/Chinese-Numbers.yaml
```diff
@@ -176,14 +176,19 @@ PairRegex: .*[双对雙對]$
...
+KiloEndRegex: !nestedRegex
```
tellarin, line 183: This seems to be a bit too complex as a fix. And it would also fail for cases like "300m", right?

tellarin: Perhaps it could follow a design more like #2353?

tellarin (changes requested):

tellarin:
@iofu728, I'll close the PR as @ChopperMan33 is already working on a more general fix.

PR #2281, opened 2020-09-18
[EN Python DateTimeV2] Added support for text like "from around 1pm" (#2243)

LionbridgeCSII:
Ported .NET solution for issue #2243 to Python.

The test case "I am feeling sick around 1pm" is still unsupported in Python because there is an inconsistency between the two platforms:

- In .NET, the entity type is always modified to its corresponding range-type independently of the modifier (so for "around 1pm", the type "time" changes to "timerange")
- In Python, only 'before', 'after' and 'since' modifiers modify the type (so for "around 1pm", the returned type remains "time")

### DIFF
--- a/Python/libraries/recognizers-date-time/recognizers_date_time/date_time/base_merged.py
+++ b/Python/libraries/recognizers-date-time/recognizers_date_time/date_time/base_merged.py
@@ -357,6 +357,9 @@ class BaseMergedExtractor(DateTimeExtractor):
     def add_mod(self, extract_results: List[ExtractResult], source: str) -> List[ExtractResult]:
 
         for extract_result in extract_results:
+            # AroundRegex is matched non-exclusively before the other relative regexes in order
+            # to catch also combined modifiers e.g. "before around 1pm"
+            self.try_merge_modifier_token(extract_result, self.config.around_regex, source)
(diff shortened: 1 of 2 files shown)


tellarin (approved):

tellarin:
@LionbridgeCSII, we'll check where the discrepancy came from and decide which way it should be unified. Can you open an issue for it?

tellarin:
@LionbridgeCSII build is failing here.

LionbridgeCSII:
> @LionbridgeCSII, we'll check where the discrepancy came from and decide which way it should be unified. Can you open an issue for it?

Ok, will do.

LionbridgeCSII:
> @LionbridgeCSII build is failing here.

this is strange, it builds correctly on my side. I will take a closer look later.

LionbridgeCSII:
> > @LionbridgeCSII build is failing here.
> 
> this is strange, it builds correctly on my side. I will take a closer look later.

The pattern "from around 1pm" fails because of PR #2292, since now "from" is matched as modifier only in Experimental mode (and in Standard mode only "around 1pm" is extracted which is affected by the discrepancy pointed out above).
Does Python support Experimental mode? In case, I can add support for the test case there.

Issue #2283, opened 2020-09-22
[ES DateTimeV2] Null resolutions for ranges


tellarin:
Some ranges are extracted, but have null resolution. For example:

- "El rango es de 2014 hasta mayo de 2018"
- "Las ventas aumentaron durante 2017-t1"
- "el conflicto duró de enero a abril de 2015"

Spanish DateTimeModel.json contains specs to cover this issue (marked in the _Comment_ tag as **RANGECRIT**).

tellarin:
@LionbridgeCSII, could you prioritize this issue? Please try to fix all existing specs tagged with this issue and re-enable them (or also add new specs).
But feel free to create PRs that solve parts of the problem (with the respective cases enabled.
Thanks.

tellarin:
@LionbridgeCSII #2286 and #2284 would be the next ones.

PR #2290, opened 2020-09-23
[ES DateTimeV2] Fixed null resolutions for ranges (#2283)

LionbridgeCSII:
Fixed null resolution for spec cases witch comment tag RANGECRIT (#2283).
Test cases re-enabled in .NET.

### DIFF
--- a/Specs/DateTime/Spanish/DateTimeModel.json
+++ b/Specs/DateTime/Spanish/DateTimeModel.json
@@ -7185,8 +7185,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "Comment": "RANGECRIT. Null resolution.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "de 2014 hasta mayo de 2018",
@@ -7211,8 +7210,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "Comment": "RANGECRIT. Null resolution.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "desde 2014 hasta 2 mayo 2018",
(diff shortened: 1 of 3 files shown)


Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -14889,16 +14873,16 @@
     ]
   },
   {
-    "Input": "Saldré de las 11:30 a.m. a las 12:30 a.m. 27 de diciembre",
+    "Input": "Saldré de las 11:30 a.m. a las 12:30 27 de diciembre",
```
tellarin, line 14876: Does a similar sentence with am/pm that makes sense on both sider work correctly?
Can you add such spec if it doesn't currently exist?

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -18092,8 +18070,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "Comment": "RANGECRIT. Null resolution.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
```
tellarin, line 18074: I just noticed there is a range case marked "CRIT Range support". Can yuo check it and see if the problem is related to this PR or a different one?

tellarin (approved):
This looks good. We'll merge once builds pass.
Meanwhile, please check the two comments on this review. Thanks.

Specs/DateTime/Spanish/DateTimeModel.json
LionbridgeCSII: yes, the original sentence works correctly too, I removed the second 'a.m.' to conform the input to the corresponding English case. Added new test case.

Specs/DateTime/Spanish/DateTimeModel.json
LionbridgeCSII: the case is still failing, the problem seems to be caused by the single datepoints not being extracted correctly. Should I fix it in this PR?

tellarin: It can be a separate fix after the higher priority issues are done. Thanks for checking. I had forgotten why it had been classified differently from the ones in this issue.

Issue #2287, opened 2020-09-22
[EN DateTimeV2] Inconsistency between English and Spanish in handling "from" and issue with list


tellarin:
Currently there is an inconsistency between English and Spanish in handling cases like "from 2015".

Spanish correctly handles it as range if in Experimental mode (#2230), however English is always treating it as such. This behaviour should be changed to match the one for Spanish.

Also, even if under Experimental mode, in cases like "from 2015 and 2016" the recognizers should return only two range years "2015" and "2016", which currently fails.

tellarin:
@LionbridgeCSII, can you fix this after completing #2283 or in parallel to it? This is a new blocking issue for some consumers of the packages.

LionbridgeCSII:
@tellarin, only the pattern 'from' + [year] should be handled in Experimental mode, right? 
While other expressions like 'from' + [date|dateperiod] (e.g. 'from today', 'from october', 'from the end of 1989') should continue to be supported in standard mode.

tellarin:
This can get tricky. In Spanish any case of "de <datetimeV2>" is only treated as "since" under Experimental mode, right? If so, then English should behave the same.

LionbridgeCSII:
> This can get tricky. In Spanish any case of "de " is only treated as "since" under Experimental mode, right? If so, then English should behave the same.

ok, then I will follow the Spanish implementation.
And for the affected cases in the model, I will modify their expected results ( and move the originals to the experimental model).

tellarin:
> And for the affected cases in the model, I will modify their expected results ( and move the originals to the experimental model).

Please do, and keep the "non-range" version of the results in the original Model specs file. I.e. each sentence in both files, showing the difference in extraction/resolution.

Thanks!

PR #2292, opened 2020-09-23
[EN DateTimeV2] Fixed inconsistency between English and Spanish in handling "from" and issue with list (#2287)

LionbridgeCSII:
Moved support for ranges like "from 2015" from Standard to Experimental mode in analogy with Spanish behaviour (#2287).

Added pattern "from 2015 and 2016"  to PotentialAmbiguousRangeRegex (FromToRegex) to avoid considering 'from' a modifier in this case.

Updated affected test cases in DateTimeModel and moved the originals to DateTimeModelExperimentalMode.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
@@ -238,7 +238,8 @@ namespace Microsoft.Recognizers.Definitions.English
       public const string AroundRegex = @"(?:\b(?:around|circa)\s*?\b)(\s+the)?";
       public static readonly string BeforeRegex = $@"((\b{InclusiveModPrepositions}?(?:before|in\s+advance\s+of|prior\s+to|(no\s+later|earlier|sooner)\s+than|ending\s+(with|on)|by|(un)?till?|(?<include>as\s+late\s+as)){InclusiveModPrepositions}?\b\s*?)|(?<!\w|>)((?<include><\s*=)|<))(\s+the)?";
       public static readonly string AfterRegex = $@"((\b{InclusiveModPrepositions}?((after|(starting|beginning)(\s+on)?(?!\sfrom)|(?<!no\s+)later than)|(year greater than))(?!\s+or equal to){InclusiveModPrepositions}?\b\s*?)|(?<!\w|<)((?<include>>\s*=)|>))(\s+the)?";
-      public const string SinceRegex = @"(?:(?:\b(?:since|after\s+or\s+equal\s+to|starting\s+(?:from|on|with)|as\s+early\s+as|(any\s+time\s+)?from)\b\s*?)|(?<!\w|<)(>=))(\s+the)?";
+      public const string SinceRegex = @"(?:(?:\b(?:since|after\s+or\s+equal\s+to|starting\s+(?:from|on|with)|as\s+early\s+as|(any\s+time\s+)from)\b\s*?)|(?<!\w|<)(>=))(\s+the)?";
+      public static readonly string SinceRegexExp = $@"({SinceRegex}|\bfrom(\s+the)?\b)";
       public const string AgoRegex = @"\b(ago|before\s+(?<day>yesterday|today))\b";
       public static readonly string LaterRegex = $@"\b(?:later(?!((\s+in)?\s*{OneWordPeriodRegex})|(\s+{TimeOfDayRegex}))|from now|(from|after) (?<day>tomorrow|tmr|today))\b";
       public const string InConnectorRegex = @"\b(in)\b";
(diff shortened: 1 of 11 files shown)


tellarin (changes requested):
Currently fails for "3 minutes from today", which ideally should be a single entity in either mode.
But it's ok to be two separate entities for now.

LionbridgeCSII:
> Currently fails for "3 minutes from today", which ideally should be a single entity in either mode.
> But it's ok to be two separate entities for now.

Modified test cases in Merged according to the corresponding case in Model. 
Strangely in Visual Studio the original test cases (with extracted text 'from today') pass (in Merged) and the updated cases (with extracted text 'today') fail.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishMergedExtractorConfiguration.cs
```diff
@@ -16,8 +16,9 @@ public class EnglishMergedExtractorConfiguration : BaseDateTimeOptionsConfigurat
         public static readonly Regex AfterRegex =
             new Regex(DateTimeDefinitions.AfterRegex, RegexFlags);
 
-        public static readonly Regex SinceRegex =
-            new Regex(DateTimeDefinitions.SinceRegex, RegexFlags);
+        // used in Experimental mode
+        public static readonly Regex SinceRegexExp =
```
tellarin, line 20: Can you place the two declarations close to each other and have the decision to which one to assign explicitly show the choices? This will help readability.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -255,7 +256,7 @@ public static class DateTimeDefinitions
       public static readonly string RelativeDurationUnitRegex = $@"(?:(?:(?<=({NextPrefixRegex}|{PreviousPrefixRegex}|{ThisPrefixRegex})\s+)({DurationUnitRegex}))|((the|my))\s+({RestrictedTimeUnitRegex}))";
       public static readonly string ReferenceDatePeriodRegex = $@"\b{ReferencePrefixRegex}\s+(?<duration>week|month|year|decade|weekend)\b";
       public const string ConnectorRegex = @"^(-|,|for|t|around|@)$";
-      public const string FromToRegex = @"\b(from).+(to)\b.+";
+      public const string FromToRegex = @"(\b(from).+(to)\b.+|\b(from)\s+\d{4}\s+and\s+\d{4})";
```
tellarin, line 259: I don't get this regex. Why is this needed?
It seems this is more of a workaround to the issue? Then it seems to be very brittle and will fail for cases like:
"from january and february" or for "from 2002 or 2003".

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishMergedParserConfiguration.cs
```diff
@@ -11,7 +11,8 @@ public EnglishMergedParserConfiguration(IDateTimeOptionsConfiguration config)
         {
             BeforeRegex = EnglishMergedExtractorConfiguration.BeforeRegex;
             AfterRegex = EnglishMergedExtractorConfiguration.AfterRegex;
-            SinceRegex = EnglishMergedExtractorConfiguration.SinceRegex;
+            SinceRegex = (config.Options & DateTimeOptions.ExperimentalMode) != 0 ? EnglishMergedExtractorConfiguration.SinceRegexExp :
```
tellarin, line 14: Do we need this logic here if it's been properly assigned in EnglishMergedExtractorConfiguration.cs?

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
tellarin: If this regex is just kind of a filter to avoid other cases falling into the ExperimentalMode behaviour, then we should rename it.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishMergedExtractorConfiguration.cs
LionbridgeCSII: moved declarations close to each other.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
LionbridgeCSII: it was meant to exclude only the cases like "from 2002 and 2003". Now I modified it to 
(\b(from).+(to|and|or)\b.+)
but maybe this is too unrestricted?

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishMergedParserConfiguration.cs
LionbridgeCSII: yes, as commented in #2230 this seems to be necessary. Otherwise, the regex in MergedParser always gets the standard expression even in Experimental mode.

LionbridgeCSII: I do not know if there is a different way to do the assignment in the Extractor configuration file which allows to avoid this logic.

tellarin: Let's see. For now let's use this version.

tellarin (approved):

Issue #2284, opened 2020-09-22
[ES DateTimeV2] Recognition issues with relative ranges and periods


tellarin:
Some relative datetime mentions are not properly recognized in Spanish For example:

- "¿Qué pasará en los 5 próximos años?"
- "Iré a China los próximos cinco días o los próximos cuarenta días."

Spanish DateTimeModel.json contains specs to cover this issue (marked in the _Comment_ tag as **RELPERIOD**).

PR #2305, opened 2020-10-14
 [ES DateTimeV2] Improved recognition with relative ranges and periods (#2284)

LionbridgeCSII:
Fixed recognition issues with relative ranges and periods in Spanish (#2284).

### DIFF
--- a/Specs/DateTime/Spanish/DateTimeModel.json
+++ b/Specs/DateTime/Spanish/DateTimeModel.json
@@ -4233,8 +4233,7 @@
     "Context": {
       "ReferenceDateTime": "2018-03-23T00:00:00"
     },
-    "Comment": "RELPERIOD. Refine support for relative periods.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "próximos cinco días",
@@ -4305,8 +4304,7 @@
     "Context": {
       "ReferenceDateTime": "2018-03-25T01:00:00"
     },
-    "Comment": "RELPERIOD. Refine support for relative periods.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "2 horas",
@@ -6358,8 +6356,7 @@
     "Context": {
       "ReferenceDateTime": "2018-06-26T00:00:00"
     },
-    "Comment": "RELPERIOD. Refine support for relative periods.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "martes antes de las dos de la tarde",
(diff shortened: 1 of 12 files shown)


Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -554,7 +554,7 @@
     "Context": {
       "ReferenceDateTime": "2019-08-23T01:00:00"
     },
-    "NotSupported": "javascript, python",
+    "NotSupported": "java, javascript, python",
```
tellarin, line 557: Why have these stopped working in Java? Because it misses the new merger method?

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -7581,13 +7582,12 @@
...
     },
-    "Comment": "RELPERIOD. Refine support for relative periods.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
-        "Text": "el año anterior",
-        "Start": 20,
-        "End": 34,
+        "Text": "año",
```
tellarin, line 7588: Isn't this incorrect? I'd rather it recognize "the previous year" in such sentences.

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -12017,8 +12007,8 @@
     "Context": {
       "ReferenceDateTime": "2019-01-31T00:00:00"
     },
-    "Comment": "RELPERIOD. Refine support for relative periods.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "Comment": "'miércoles pasado' seems to translate to 'last Wednesday' instead of 'past Wednesday'. Only 'este miércoles pasado' corresponds to 'this past Wednesday'.",
```
tellarin, line 12010: We need to check this one with a fluent Spanish speaker. I know a bit, but I'm not native fluent.

Specs/DateTime/Spanish/DateTimeModel.json
tellarin: We can keep this behaviour for now. But please open an issue for this case and tag it with the 'open question' label.

Specs/DateTime/Spanish/DateTimeModel.json
tellarin: If so, let's open a Java issue to port this change there and re-enable these. Lower priority than continuing improving Spanish support in .NET for now.

tellarin (approved):
Approved with comments.

Specs/DateTime/Spanish/DateTimeModel.json
LionbridgeCSII: The problem is that this pattern is ambiguous, it could mean "the previous year", but in this context it means "the year before the revolution".
So I conformed it with the English behaviour where for the case "The story happened the year before independence." only "the year" is extracted.
Maybe I could modify the behaviour to take into account "anterior" when it is not followed by prepositions.

LionbridgeCSII: anyway, yes I will open the issue

Specs/DateTime/Spanish/DateTimeModel.json
LionbridgeCSII: yes, I think so.
Ok, will open the issue

Specs/DateTime/Spanish/DateTimeModel.json
LionbridgeCSII: @tellarin, we asked a native Spanish speaker and he said:

The problem with Spanish is that we use “pasado” for both “last” and “past” and only context will help to disambiguate.
Some tips to help to avoid ambiguity are:
- Adding “este” -> to refer the day of the current week.
- Adding “de la semana pasada” -> to refer to the day of the previous week.

PR #2312, opened 2020-10-17
[ZH Number] fix #2297 which about cases like '1,100万' 、'1,100.9亿'

### LINKED ISSUE #2297: [ZH Number] "1,100万" 、"1,100.9亿" etc. can't be recognized as number
**Describe the bug**
"1100万"、"1100.9亿"  etc. can be be recognized as number.
but
**"1,100万" 、"1,100.9亿" etc. can't be recognized as number**

**Expected behavior**
```
{
        "Text": "1,100万",
        "Start": 0,
        "End": 5,
        "TypeName": "number",
        "Resolution": {
            "Value": "11000000"
        }
 }

{
        "Text": "1,100.9亿",
        "Start": 0,
        "End": 7,
        "TypeName": "number",
        "Resolution": {
            "Value": "110090000000"
        }
 }
```

 - Platform: [.NET ...]
 - Version of package [v1.4.1]

ChopperMan33:
Fix for issues pointed out in #2297
The previous model can't recognize thousands separators in cases like "1,100万”, so I changed the regex pattern in "DoubleWithThousandsRegex" to recognize the thousands separators.
And enabled test cases in NumberModel.


Patterns/Chinese/Chinese-Numbers.yaml
```diff
@@ -234,7 +234,7 @@ DoubleWithMultiplierRegex: !nestedRegex
   def: ({NegativeNumberTermsRegexNum}\s*)?{ZeroToNineFullHalfRegex}+[\.．]{ZeroToNineFullHalfRegex}+\s*{BaseNumbers.NumberMultiplierRegex}
   references: [NegativeNumberTermsRegexNum, ZeroToNineFullHalfRegex, BaseNumbers.NumberMultiplierRegex]
 DoubleWithThousandsRegex: !nestedRegex
-  def: '{NegativeNumberTermsRegex}?{ZeroToNineFullHalfRegex}+([\.．]{ZeroToNineFullHalfRegex}+)?\s*[多几幾余]?[万亿萬億]{1,2}'
+  def: '({NegativeNumberTermsRegex}?{ZeroToNineFullHalfRegex}+([\.．]{ZeroToNineFullHalfRegex}+)?\s*[多几幾余]?[万亿萬億]{1,2})|({NegativeNumberTermsRegex}?{ZeroToNineFullHalfRegex}{1,3}(,{ZeroToNineFullHalfRegex}{3})+([\.．]{ZeroToNineFullHalfRegex}+)?\s*[多几幾余]?[万亿萬億]{1,2})'
```
tellarin, line 237: Could this pattern be simplified? It seems to mostly duplicate the previous one. If simplifying it now would be hard, we can use the new one for now, as long as there is no negative performance impact.

tellarin (changes requested):

Patterns/Chinese/Chinese-Numbers.yaml
ChopperMan33: Thank you for your advice. I've simplified the regex pattern.

tellarin (approved):

PR #2317, opened 2020-10-19
[ZH Dimension] fix #2301 which about cases like '一吨半' 、'两米半' 

### LINKED ISSUE #2301: [ZH Dimension]"一吨半"、"两米半"、"三斤半" etc. can't be recognized as dimension
**Describe the bug**
"一吨半"、"两米半"、"十五斤半" etc. can't be recognized as dimension
"一吨半" means "**one and a half tons**" that can be recognized in culture "en" but not in "zh-cn".
Similarly, 
"两米半" means "**two and a half meters**";
...

**Expected behavior**
```
{
        "Text": "一吨半",
        "Start": 0,
        "End": 2,
        "TypeName": "dimension",
        "Resolution": {
            "unit":"Ton",
            "Value": "1.5"
        }
 }

{
        "Text": "两米半",
        "Start": 0,
        "End": 2,
        "TypeName": "dimension",
        "Resolution": {
            "unit":"Meter",
            "Value": "2.5"
        }
 }

{
        "Text": "十五斤半",
        "Start": 0,
        "End": 3,
        "TypeName": "dimension",
        "Resolution": {
            "unit":"Jin",
            "Value": "15.5"
        }
 }
```

 - Platform: [.NET ...]
 - Version of package [v1.4.1]

ChopperMan33:
Fix for issues pointed out in #2301.

- The previous model can't recognize ‘半’ after unit in cases like ‘一吨半’. And there're two numerics and one unit in this phrase. So I add an expanding strategy in the `NumberWithUnitExtractor` class, separately parse each numeric and add up these two numbers.
- To support JavaScript which can't import regex resources from different modules, I add `HalfUnitRegex` in `chinese_NumberWithUnit.yaml`.
- Remove previous half strategy in Chinese data time module and the corresponding unit test cases in `DurationParser`.
- And enabled test cases in `DimensionModel`.


Specs/DateTime/Chinese/DurationParser.json
```diff
@@ -98,7 +98,7 @@
     },
     "Results": [
       {
-        "Text": "三年",
+        "Text": "三年半",
```
tellarin, line 101: Can you add a case like this as a full sentence in DateTimeModel.json too?

Patterns/Chinese/Chinese-NumbersWithUnit.yaml
```diff
@@ -756,4 +756,6 @@ TemperatureAmbiguousValues: !list
   entries:
     - 度
     - k
+HalfUnitRegex: !simpleRegex
+  def: 半
```
tellarin, line 760: Shouldn't this regex check if the match is at the end of the string or similar?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
```diff
@@ -318,6 +323,54 @@ public List<ExtractResult> Extract(string source)
                 result = SelectCandidates(source, result, unitIsPrefix);
             }
 
+            // Expand Chinese Half Patterns
```
tellarin, line 326: If this is only used in Chinese, it shouldn't be in the base NumberWithUnit extractor/parser code. Perhaps a method in the corresponding Chinese config that only has implementation there? In the default config the method would just not do anything.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Extractors/FrenchNumberWithUnitExtractorConfiguration.cs
```diff
@@ -50,6 +50,8 @@ protected FrenchNumberWithUnitExtractorConfiguration(CultureInfo ci)
 
         public virtual Regex AmbiguousUnitNumberMultiplierRegex => null;
 
+        public virtual Regex HalfUnitRegex => null;
```
tellarin, line 53: The move to a method in a Chinese config would eliminate all these unused regex references too.

Specs/DateTime/Chinese/DurationParser.json
tellarin: Can you also add a sentence here that has two examples of such cases in the same sentence, to show that the current solution doesn't misbehave in such situations?

tellarin: Same for an example with two durations in DateTimeModel.json

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
tellarin: Can you also add some code comment here explaining what this code is doing? That will help future maintainers.

tellarin (changes requested):

Specs/DateTime/Chinese/DurationParser.json
ChopperMan33: I've add some unit test cases both on DateTimeModel.json and DurationParser.json.

Patterns/Chinese/Chinese-NumbersWithUnit.yaml
ChopperMan33: Due to the different regex results in Java, we need to match the "半" pattern from the source string rather than the pre-recognizer phrases. And the index check makes sure the "半" immediately follows a phrase which composed of a number and a unit.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Extractors/FrenchNumberWithUnitExtractorConfiguration.cs
ChopperMan33: Same as previous discussion, this part is not necessary for now and has been removed.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/NumberWithUnitExtractor.cs
tellarin: @ChoppperMan33, sorry, that's not what I meant. We should avoid specializing the base extractor/parser classes if possible. 
The extension points ideally would be the language-specific configuration classes.
For example, NumberWithUnitExtractor takes a INumberWithUnitExtractorConfiguration by language. That config would have a method for this, which basically does nothing by default. But the Chinese instance of the config would implement the new behaviour. Can you change to that?
Let us know if you have any questions.
BTW, do you know if Japanese would have the same case with 'half'?

ChopperMan33: Do you means set a default HalfUnitRegex in INumberWithUnitExtractorConfiguration construct function like https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.Sequence/Config/IpConfiguration.cs#L9, or structure a Extract function in ChineseNumberWithUnitExtractorConfiguration and put all code to that language-specific configuration classe?
Can you give me an example?
One more thing, the Japanese also have similar cases like "1トン半".
Thank you.

tellarin: Not only for the regex. That whole logic specific to processing the 'half' case should be a method in the language-specific config where it's necessary. I meant something like AdjustBySuffix in EnglishTimeParserConfiguration (but this specific method is not really well written). The base extractor/parser would then call the method from the config. 
Basically there could be a default empty implementation in some base config, and then only the Chinese and Japanese configs would actually implement the method. In most languages nothing would be done, in Chinese and Japanese , the 'half' case would be processed. 
Do you think you can also fix in Japanese after we have the final design for Chinese?

ChopperMan33: Understand, I try to fix it and support the Japanese language.

tellarin: Please write descriptive comments for each commit message too, ok? Thanks a lot!

ChopperMan33: I've finished the code reconstruct.  For Japanese, I find our model doesn't support any dimension cases even if "1トン" which means "1 Ton", ref this link https://github.com/microsoft/Recognizers-Text/blob/master/Specs/NumberWithUnit/Japanese/DimensionModel.json#L344. So I don't change that code. Thank you.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/NumberWithUnitParser.cs
```diff
@@ -104,10 +117,16 @@ public ParseResult Parse(ExtractResult extResult)
                     var numValue = string.IsNullOrEmpty(numberResult.Text) ?
                         null :
                         this.Config.InternalNumberParser.Parse(numberResult);
+                    var resolution_str = numValue?.ResolutionStr;
```
tellarin, line 120: Code style: Please rename to resolutionStr.

tellarin (approved):
Minor rename request for code style. Please fix in the next PR.
Do you think you can implement the suffix expansion for Japanese too and add a couple test specs for it in a new same PR?
Thanks!

PR #2320, opened 2020-10-21
Adding subtype info in Spanish Number specs

tellarin:
- Adding subtype info in Spanish Number specs;
- Minor resource edits in Spanish and Portuguese;
- Re-gen resources across platforms.


iofu728 (approved):
It seems like support the `son` and `igual` pattern in Spanish.

wgx998877 (approved):

PR #2327, opened 2020-10-25
[ES DateTimeV2] Fixed incorrect entity boundaries / overly aggressive merging (#2285)

LionbridgeCSII:
Fix for issue #2285.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
@@ -25,7 +25,8 @@ namespace Microsoft.Recognizers.Definitions.Spanish
       public const bool CheckBothBeforeAfter = false;
       public const string TillRegex = @"(?<till>\b(hasta|hacia|al?)\b|~|--|-|—|——)(\s+(el|la(s)?)\b)?";
       public static readonly string RangeConnectorRegex = $@"(?<and>\b(y\s*(el|(la(s)?)?))\b|{BaseDateTime.RangeConnectorSymbolRegex})";
-      public const string DayRegex = @"(?<day>01|02|03|04|05|06|07|08|09|10|11|12|13|14|15|16|17|18|19|1|20|21|22|23|24|25|26|27|28|29|2|30|31|3|4|5|6|7|8|9)(?=\b|t)";
+      public const string WrittenDayRegex = @"(?<day>uno|dos|tres|cuatro|cinco|seis|siete|ocho|nueve|diez|once|doce|trece|catorce|quince|dieciséis|diecisiete|dieciocho|diecinueve|veinte|veintiuno|veintidós|veintitrés|veinticuatro|veinticinco|veintiséis|veintisiete|veintiocho|veintinueve|treinta|treinta y uno)";
+      public const string DayRegex = @"\b(?<day>01|02|03|04|05|06|07|08|09|10|11|12|13|14|15|16|17|18|19|1|20|21|22|23|24|25|26|27|28|29|2|30|31|3|4|5|6|7|8|9)(?=\b|t)";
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
```diff
@@ -25,7 +25,8 @@ public static class DateTimeDefinitions
       public const bool CheckBothBeforeAfter = false;
       public const string TillRegex = @"(?<till>\b(hasta|hacia|al?)\b|~|--|-|—|——)(\s+(el|la(s)?)\b)?";
       public static readonly string RangeConnectorRegex = $@"(?<and>\b(y\s*(el|(la(s)?)?))\b|{BaseDateTime.RangeConnectorSymbolRegex})";
-      public const string DayRegex = @"(?<day>01|02|03|04|05|06|07|08|09|10|11|12|13|14|15|16|17|18|19|1|20|21|22|23|24|25|26|27|28|29|2|30|31|3|4|5|6|7|8|9)(?=\b|t)";
+      public const string WrittenDayRegex = @"(?<day>uno|dos|tres|cuatro|cinco|seis|siete|ocho|nueve|diez|once|doce|trece|catorce|quince|dieciséis|diecisiete|dieciocho|diecinueve|veinte|veintiuno|veintidós|veintitrés|veinticuatro|veinticinco|veintiséis|veintisiete|veintiocho|veintinueve|treinta|treinta y uno)";
```
tellarin, line 28: Why is this required in Spanish if there's no equivalent in English?

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishDateParserConfiguration.cs
```diff
@@ -42,7 +42,7 @@ public SpanishDateParserConfiguration(ICommonDateTimeParserConfiguration config)
             UpcomingPrefixRegex = new Regex(DateTimeDefinitions.UpcomingPrefixRegex, RegexFlags);
             PastPrefixRegex = new Regex(DateTimeDefinitions.PastPrefixRegex, RegexFlags);
 
-            DayOfMonth = config.DayOfMonth;
+            DayOfMonth = DateTimeDefinitions.DayOfMonth.ToImmutableDictionary();
```
tellarin, line 45: Why this change?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/Spanish/DateTimeDefinitions.cs
LionbridgeCSII: We followed the Italian implementation where WrittenDayRegex is also used. We introduced this new regex there because in English these cases are covered by FlexibleDayRegex: (?<DayOfMonth>([A-Za-z]+\s)?[A-Za-z\d]+) which is constrained by (?<=(st|nd|rd|th)). But in Italian and Spanish, cardinal numbers are used instead of ordinals and without constraint the orignal regex matches too much.

.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishDateParserConfiguration.cs
LionbridgeCSII: config.DayOfMonth gets its values from BaseDateTime.DayOfMonthDictionary which only contains digit numbers. We use instead the Spanish version of the dictionary which contains also written day numbers in order to parse the cases like "on the twenty-six". 
But this can probably be avoided by enabling ForTheRegex in Spanish, instead of using OnRegex to parse these patterns. I will try this solution.

tellarin (approved):

Issue #2289, opened 2020-09-22
[EN NumberRange] Parsing exception in sentence with multiple range mentions


tellarin:
The sentence "find sales greater than 30 and less than or equal to 40 and greater than or equal to 50" causes an exception during extraction. The expected results would be the extraction on the different ranges.

Each of the three mentions is extracted correctly if in a sentence by themselves.

tellarin:
"greater than 30 and less than or equal to 40" by itself works correctly and returns a single range too.
"less than or equal to 40 and greater than or equal to 50" by itself is incorrectly interpreted. It should return two disjoint ranges.

tellarin:
@LionbridgeCSII, can you take a look if this still happens?

LionbridgeCSII:
> @LionbridgeCSII, can you take a look if this still happens?

yes, it does. I started looking into this issue yesterday.

PR #2328, opened 2020-10-26
[EN NumberRange] Fixed parsing exception in sentence with multiple range mentions (#2289)

LionbridgeCSII:
Fixed issue #2289. 

But I fear similar problems could still arise, especially in long sentences, given the current structure of the range regexes (which pose only few constraints on what is initially interpreted as number, e.g. "(?\<number1\>({NumberSplitMark}.)+)").

So, I am wondering if it would not be better to reorganize the extractor following an approach similar to that of MergedDateTimeExtractor. That is, first extract all numerical entities and then check their afterString and beforeString for modifiers and connectors to determine if they represent ranges.


tellarin:
> So, I am wondering if it would not be better to reorganize the extractor following an approach similar to that of MergedDateTimeExtractor. That is, first extract all numerical entities and then check their afterString and beforeString for modifiers and connectors to determine if they represent ranges.

Yes, this is a good point. We should plan for this, but this would be a bigger effort than issue fixing. Let's follow up internally first on that.

tellarin (approved):

Issue #2334, opened 2020-10-26
[DE DateTimeV2] Set and time being incorrectly mixed during recognition


mujdricza:
**Describe the bug**
dateTimex entity is not recognized, if something follows in the sentence.

**To Reproduce**
1. Make a LUIS app with prebuiltEntity "datetimeV2"
2. Type 'jeden Tag um 19:13 Uhr bitte' or "jeden Tag um 19:13 Uhr Alarm stellen" or something according to the template "jeden Tag um xx:xx: Uhr <something>"
4. See error in the sample input/output

**Expected behavior**
a dateTimex like "T1913" 
Correctly working example: "jeden Abend um 19:13 Uhr Alarm stellen"; among the entities, also a dateTimex entity is detected: (Central European Time Zone)
```
"dateTimex": [
                  {
                    "value": "T19:13",
                    "literal": "jeden Abend um 19:13 Uhr"
                  }
                ]
```

**Sample input/output with the bug**
input: "jeden Tag um 19:13 Uhr Alarm stellen"
output: (with further entities "all", "number", and "skill_name"; datetime entity is missing!)
```
"entities": {
              "all": [
                {
                  "value": "all",
                  "literal": "jeden"
                }
              ],
              "number": [
                {
                  "value": "19",
                  "literal": "19:13"
                },
                {
                  "value": "13",
                  "literal": "19:13"
                }
              ],
              "skill_name": [
                {
                  "value": "alarm",
                  "literal": "Alarm"
                }
              ]
            }
          },
```

**Platform (please complete the following information):**
 - Platform: Python/Java
 - Environment: LUIS
 - Version of package (not sure)

tellarin:
We've replicated a bug with sentences like "jeden Tag um 19:13 Uhr Alarm stellen" that's different from the reported issue.
Currently the output for this case is "den Tag um 19:13 Uhr" which is incorrect and also mixes the recurring date and the time mentions.
@LionbridgeCSII, can you take a look? "jeden Abend", "jeden Tag" should be recognized as Set. And "den" should not be parsed out of "jeden" into a date (word boundary issue).

PR #2340, opened 2020-10-27
[DE DateTimeV2] Fixed set and time being incorrectly mixed during recognition (#2334)

LionbridgeCSII:
Fix to issue #2334. 

The behaviour of the extractor has been conformed to the English one. 
But, should we modify it to extract "jeden Tag um 19:13 Uhr" as a single entity? This happens already when the time entity is at the end of the utterance (e.g. "Ich gehe jeden Tag um 3 Uhr Nachmittags").

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
@@ -25,8 +25,8 @@ namespace Microsoft.Recognizers.Definitions.German
       public const bool CheckBothBeforeAfter = false;
       public const string TillRegex = @"(?<till>zu|bis\s*zum|zum|bis|bis\s*hin(\s*zum)?|--|-|—|——)";
       public const string RangeConnectorRegex = @"(?<and>und|--|-|—|——)";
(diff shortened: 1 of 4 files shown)


tellarin:
> The behaviour of the extractor has been conformed to the English one.
> But, should we modify it to extract "jeden Tag um 19:13 Uhr" as a single entity? This happens already when the time entity is at the end of the utterance (e.g. "Ich gehe jeden Tag um 3 Uhr Nachmittags").

Ideally yes, I think. But what's the behaviour in English for the sentences corresponding to these two exact cases? Always separate?

tellarin (approved):

LionbridgeCSII:
> Ideally yes, I think. But what's the behaviour in English for the sentences corresponding to these two exact cases? Always separate?

Yes, in English they are always separate because the regex that should check for "every day" before the time entity (BeforeEachDayRegex) is set to null. So at the moment, only cases where "every day" follows the time are extracted as a single Set entity (e.g. "7:13 p.m. every day").

tellarin:
Let's leave as-is in this PR for now then. Can you resolve the conflicts to merge? Thanks!

LionbridgeCSII:
> Let's leave as-is in this PR for now then. Can you resolve the conflicts to merge? Thanks!

ok, conflict solved.

PR #2349, opened 2020-11-02
[ZH Number] Fix interpretation of numbers mixing Chinese and Arabic numerals, like "1千零20" and "2千零30万" (#2302)

### LINKED ISSUE #2302: [ZH Number] "1千零1"、"1千零20"、"2千零30万"、"3万4000"、"25万6千" etc. can't be recognized as number
**Describe the bug**
"1千零1"、"1千零20"、"2千零30万"、"3万4000"、"25万6千" etc. can't be recognized as number

"1千零1" means 1001
"1千零20" means 1020
"2千零30万" means 20300000
"3万4000" means 34000
"25万6千" means 256000
……

**Expected behavior**
```
{
        "Text": "1千零1",
        "Start": 0,
        "End": 3,
        "TypeName": "number",
        "Resolution": {
            "Value": "1001"
        }
 }

{
        "Text": "1千零20",
        "Start": 0,
        "End": 4,
        "TypeName": "number",
        "Resolution": {
            "Value": "1020"
        }
 }

{
        "Text": "2千零30万",
        "Start": 0,
        "End": 5,
        "TypeName": "number",
        "Resolution": {
            "Value": "20300000"
        }
 }

{
        "Text": "3万4000",
        "Start": 0,
        "End": 5,
        "TypeName": "number",
        "Resolution": {
            "Value": "34000"
        }
 }

{
        "Text": "25万6千",
        "Start": 0,
        "End": 4,
        "TypeName": "number",
        "Resolution": {
            "Value": "256000"
        }
 }
```

 - Platform: [.NET ...]
 - Version of package [v1.4.1]

ChopperMan33:
Fix for issues pointed out in #2302 

The previous model can't recognize Chinese integer phrase with the digit in cases like "1千零1"、"1千零20"、"2千零30万"、"3万4000"、"25万6千", so I changed the regex pattern in `AllIntRegex`, `NotSingleRegex`, and add some code to parse that mode.
The model passes all the test cases in NumberModel.


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -571,22 +572,43 @@ private double GetIntValue(string intStr)
                         }
                         else
                         {
-                            beforeValue = Config.ZeroToNineMap[intStr[i]];
+                            double tmp = Config.ZeroToNineMap[intStr[i]];
```
tellarin, line 575: Rename "tmp" to something like "currentDigit".

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -488,6 +488,7 @@ private double GetIntValue(string intStr)
             var isRoundBefore = false;
             long roundBefore = -1, roundDefault = 1;
             var isNegative = false;
+            var isPreDigit = false;
```
tellarin, line 491: This flag indicates previous digits have been processed, right? I'd suggest renaming it to something like isMultiDigit

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -571,22 +572,43 @@ private double GetIntValue(string intStr)
...
+                                beforeValue = tmp;
+                            }
+
                             isRoundBefore = false;
                         }
                     }
                     else
                     {
-                        if (i == intStr.Length - 1 && (Config.CultureInfo.Name == "ja-JP" || Config.CultureInfo.Name == "ko-KR"))
+                        if ((Config.CultureInfo.Name == "ja-JP" || Config.CultureInfo.Name == "ko-KR") || char.IsDigit(intStr[i]))
```
tellarin, line 590: I'm not sure this new 'if' is correct. The previous behaviour would only enter here is both the first clause was true and the language was either JA or KO. Can you double check the behaviour and clean it up?

tellarin (changes requested):
Mostly looks OK. But that 'if' needs to be checked. Thanks.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -571,22 +572,43 @@ private double GetIntValue(string intStr)
...
+                            {
+                                beforeValue = tmp;
+                            }
+
                             isRoundBefore = false;
                         }
                     }
                     else
                     {
-                        if (i == intStr.Length - 1 && (Config.CultureInfo.Name == "ja-JP" || Config.CultureInfo.Name == "ko-KR"))
```
ChopperMan33, line 580: We have already judged it in line 565, so I delet it.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
ChopperMan33, line 590: When the last character is a digit, set the roundDefault to 1. For example, "一千200“, when we at the last position, we will set roundDefault from 1000 to 1.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
ChopperMan33, line 491: Can I name it  as "isDigitBefore"?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
tellarin: But why to check if the culture is Japanese or Korean?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
tellarin: isMultiDigit means that it's not a single digit, then it must have previous digits.
Or it could be hasPreviousDigits too.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
ChopperMan33: OK, I will check it. Thank you.

tellarin (approved):

Issue #2347, opened 2020-11-01
[* Number] Parsing of vulgar fractions in unicode


hansmbakker:
**Is your feature request related to a problem? Please describe.**
Currently, Recognizers-Text can parse fractions, but it does not seem to parse vulgar fractions written in unicode characters:

A regex for these characters is described in https://stackoverflow.com/questions/49460802/regex-match-all-vular-fractions
It seems possible to get the numerical value using Char: https://stackoverflow.com/questions/12630007/find-numeric-value-of-a-digit-character-in-c-sharp

**Describe the solution you'd like**
Recognizers-Text being able to parse vulgar fractions written in unicode characters.

**Describe alternatives you've considered**
Pre-parse those unicode characters to decimal values before feeding to Recognizers-text

PR #2351, opened 2020-11-03
[EN Number] Added support for vulgar fractions (#2347)

LionbridgeCSII:
Added support for single character fractions (¼½¾⅐⅑⅒⅓⅔⅕⅖⅗⅘⅙⅚⅛⅜⅝⅞) in English (#2347).

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Recognizers.Definitions.English
       public static readonly string OrdinalRoundNumberRegex = $@"(?<!an?\s+){RoundNumberOrdinalRegex}";
       public static readonly string OrdinalEnglishRegex = $@"(?<=\b){AllOrdinalRegex}(?=\b)";
       public const string FractionNotationWithSpacesRegex = @"(((?<=\W|^)-\s*)|(?<=\b))\d+\s+\d+[/]\d+(?=(\b[^/]|$))";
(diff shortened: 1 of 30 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -55,7 +55,7 @@ public static class NumbersDefinitions
       public static readonly string OrdinalRoundNumberRegex = $@"(?<!an?\s+){RoundNumberOrdinalRegex}";
       public static readonly string OrdinalEnglishRegex = $@"(?<=\b){AllOrdinalRegex}(?=\b)";
       public const string FractionNotationWithSpacesRegex = @"(((?<=\W|^)-\s*)|(?<=\b))\d+\s+\d+[/]\d+(?=(\b[^/]|$))";
-      public const string FractionNotationRegex = @"(((?<=\W|^)-\s*)|(?<![/-])(?<=\b))\d+[/]\d+(?=(\b[^/]|$))";
+      public const string FractionNotationRegex = @"((((?<=\W|^)-\s*)|(?<![/-])(?<=\b))\d+[/]\d+(?=(\b[^/]|$))|[\u00BC-\u00BE\u2150-\u215E])";
```
tellarin, line 58: It seems this regex is the same everywhere, right? It would be better to move it to BaseNumber.yaml for it to be reused by all languages.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
LionbridgeCSII: Moved regex to BaseNumber.yaml and added test case to other languages.

tellarin (approved):

PR #2353, opened 2020-11-04
[ZH Dimension] Fixed extraction for cases like "12千米"、"2.3千克" (#2270).

LionbridgeCSII:
Fixed extraction of multi digit numbers when followed by units like "千米" (kilometer) and "千克" (kilogram) (#2270).
Test cases added to DimensionModel.


Patterns/Chinese/Chinese-Numbers.yaml
```diff
@@ -175,7 +175,7 @@ SpeGetNumberRegex: !nestedRegex
 PairRegex: .*[双对雙對]$
 #IntegerExtractor
 RoundNumberIntegerRegex: !simpleRegex
-  def: (((?<![十百千拾佰仟])[十百千拾佰仟])|([万亿兆萬億]))
+  def: (((?<![十百千拾佰仟])[十百千拾佰仟])|([万亿兆萬億]))(?![米克字])
```
tellarin, line 178: This fix goes in the right direction, but it's too simple and will cause failures for cases like "300m" or just number like "12千".

tellarin (changes requested):

tellarin:
@LionbridgeCSII,  I'll close the PR as @ChopperMan33 is already working on a more general fix.

PR #2354, opened 2020-11-05
[ZH Percent] Added support for percentage like "[Den]分之[Num]" (#2257)

LionbridgeCSII:
Added support for percentage like "[Den]分之[Num]" e.g. "十分之一" (10%) in .NET and Python (#2257).
Test cases added to PercentModel.

### DIFF
--- a/Specs/Number/Chinese/PercentModel.json
+++ b/Specs/Number/Chinese/PercentModel.json
@@ -2025,5 +2025,50 @@
         }
       }
     ]
+  },
+  {
+    "Input": "十分之一",
+    "NotSupported": "javascript, java",
+    "Results": [
+      {
+        "Text": "十分之一",
+        "TypeName": "percentage",
+        "Resolution": {
+          "value": "10%"
+        },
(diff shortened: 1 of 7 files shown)


LionbridgeCSII:
did not notice that this fix is overlapping with PR #2269, so maybe it is not needed.

Specs/Number/Chinese/PercentModel.json
```diff
@@ -2025,5 +2025,50 @@
...
+  },
+  {
+    "Input": "十分之一",
+    "NotSupported": "javascript, java",
+    "Results": [
+      {
+        "Text": "十分之一",
+        "TypeName": "percentage",
+        "Resolution": {
+          "value": "10%"
```
tellarin, line 2037: This both overlaps with PR #2269 and also only covers simple cases.

tellarin (changes requested):

tellarin:
Closing it as #2269 seems more general.

PR #2358, opened 2020-11-12
[SV Currency] Adding Swedish support for Currency

pete1854:


Patterns/Swedish/Swedish-NumbersWithUnit.yaml
```diff
@@ -765,13 +767,17 @@ FractionalUnitNameToCodeMap: !dictionary
     Hào: HAO
     Ngwee: NGWEE
 CompoundUnitConnectorRegex: !simpleRegex
-    def: (?<spacer>and)
+    def: (?<spacer>och)
 CurrencyPrefixList: !dictionary
   types: [ string, string ]
   entries:
 # Currency Prefix Symbols/Terms
+    Swedish krona: sek
```
tellarin, line 775: You should not have to add the ISO codes to the prefix list. This should be automatically part of the process.
Perhaps the code here got a bit out of date compared to English?

Specs/NumberWithUnit/Swedish/CurrencyModel.json
```diff
@@ -0,0 +1,425 @@
...
+  {
+    "Input": "priset för den andra simulatorn varierar med 16,4 miljoner US$",
+    "NotSupportedByDesign": "javascript, python, java",
+    "Results": [
+      {
+        "Text": "16,4 miljoner us$",
+        "TypeName": "currency",
+        "Resolution": {
+          "value": "16400000",
+          "unit": "United States dollar"
```
tellarin, line 127: There should be a isoCurrency property here. How does the test pass without it?
Can you double check if some validation is not happening properly?

tellarin (changes requested):

Specs/NumberWithUnit/Swedish/CurrencyModel.json
pete1854: I found out that I called the TestNumberWithUnit test instead of the TestCurrency test for the Currency model. I also updated the extractor to follow the English one. Now it all works. Thanks for making me aware of this :)

tellarin (approved):
Thanks!

PR #2362, opened 2020-11-13
[SV Number] Performance upgrade (cache implementation)

pete1854:
Existing number extractors have been upgraded to use the CachedNumberExtractor where applicable. They should now be aligned w/ the English extractor implementations.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Swedish/Extractors/NumberExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Swedish/Extractors/NumberExtractor.cs
@@ -6,28 +6,22 @@ using Microsoft.Recognizers.Definitions.Swedish;
 
 namespace Microsoft.Recognizers.Text.Number.Swedish
 {
-    public class NumberExtractor : BaseNumberExtractor
+    public class NumberExtractor : CachedNumberExtractor
     {
         private const RegexOptions RegexFlags = RegexOptions.Singleline | RegexOptions.ExplicitCapture;
 
         private static readonly ConcurrentDictionary<(NumberMode, NumberOptions), NumberExtractor> Instances =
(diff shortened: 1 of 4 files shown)


.NET/Microsoft.Recognizers.Text.Number/Swedish/Extractors/NumberExtractor.cs
```diff
@@ -6,28 +6,22 @@
...
         private NumberExtractor(BaseNumberOptionsConfiguration config)
             : base(config.Options)
         {
 
-            this.mode = config.Mode;
+            keyPrefix = string.Intern(ExtractType + "_" + config.Options + "_" + config.Mode + "_" + config.Culture);
 
-            NegativeNumberTermsRegex = new Regex(NumbersDefinitions.NegativeNumberTermsRegex + "$", RegexFlags);
-
-            AmbiguousFractionConnectorsRegex = new Regex(NumbersDefinitions.AmbiguousFractionConnectorsRegex, RegexFlags);
```
tellarin, line 26: Are these two removed as no longer relevant? Or not implemented in Swedish?
I need to do a comparison with English before completing the review.

pete1854: So far they've not been used. But, I guess you have a point in letting them remain. I'm working w/ the Range implementation and also improving the fraction recognizer now and can it be that they could potentially need these? I haven't checked yet. Once again - it might be better keeping them for the future.

tellarin: Let's bring it back here then. Thanks.

tellarin (approved):

PR #2370, opened 2020-11-16
[SV Numbers ] Improvements to ordinal and fraction recognition of compound number expressions

pete1854:
I'm trying to improve the recognition of ordinals and fraction numbers.

I'm creating this PR as a Draft PR since I need to discuss possible ways of solving the challenge of parsing written compound numbers in swedish.

An example:
"En tjugoförstedel" ("A twenty-firsts" in english) cannot be resolved using the current implementation of the parser. I checked the Italian solution and found that they have actually added the **specific fraction expression** to the **OrdinalNumberMap**.

Should I do this I would en up w/ at least 10 * 10 new entries (10 for each tenth), .e.g

```
tjugoförstedel: 21
tjugoandradel: 22
tjugotredjedel: 23
.
.
.
tjugoniondel: 29
```

I would also certainly need to add entries for all the variations to these expressions, e.g. for the entries for 21 I would need to add entries for

```
tjugoförst(a)del: 21
tjugoförst(adelar): 21
tjugoförst(adels): 21
tjugoförst(edelar): 21
tjugoförst(edels): 21
```
a.s.o

I would need to add hundreds of entries to cover for all possible expressions. I don't mind doing so, given there is no other way of implementing recognition for these kinds of expressions in the core.

Any suggestions are very welcomed.


tellarin:
> I would also certainly need to add entries for all the variations to these expressions, e.g. for the entries for 21 I would need to add entries for

@pete1854, perhaps we could add a "normalization" method to be called before consulting the relevant map. So, "tjugoförstadelar" -> "tjugoförsta", and that would be in the map. What do you think?

pete1854:
Yes, that could certainly be one way of solving it. Do you picture this normalization to happen in the `NormalizeTokenSet` function in `SwedishNumberParserConfiguration`, or?

pete1854:
@tellarin - I have made a prototype for one way of resolving this type of fractional expressions. 
The idea is separate Ordinal number constructs from Fractional ones and I have therefore put the latter in a separate map `SwedishWrittenFractionLookupMap`. I have then overridden the `ResolveCompositeNumber` function in `SwedishNumberParserConfiguration` class. Here I first try to resolve the expression using the default Ordinal map. If this fails (which it does for these expressions) it tries to find the value in the `SwedishWrittenFractionLookupMap`. 
I have not considered performance at this point. 
What do you think of this approach?

tellarin:
> @tellarin - Yes, that could certainly be one way of solving it. Do you picture this normalization to happen in the `NormalizeTokenSet` function in `SwedishNumberParserConfiguration`, or?

Yes.

tellarin:
> @tellarin - I have made a prototype for one way of resolving this type of fractional expressions.
> The idea is to separate Ordinal number constructs from Fractional ones and I have therefore put the latter in a separate map `SwedishWrittenFractionLookupMap`. I have then overridden the `ResolveCompositeNumber` function in `SwedishNumberParserConfiguration` class. Here I first try to resolve the expression using the defaults. If this fails (which it does for these expressions) it tries to find the value in the `SwedishWrittenFractionLookupMap`.
> I have not considered performance at this point.
> What do you think of this approach?

This looks good to me.

.NET/Microsoft.Recognizers.Text.Number/Swedish/Parsers/SwedishNumberParserConfiguration.cs
```diff
@@ -69,5 +73,23 @@ public override IEnumerable<string> NormalizeTokenSet(IEnumerable<string> tokens
 
             return fracWords;
         }
+
+        public override long ResolveCompositeNumber(string numberStr)
```
tellarin, line 77: Just add a code comment here to describe why the override.

Specs/Number/Swedish/NumberModel.json
```diff
@@ -768,22 +768,6 @@
       }
     ]
   },
-  {
-    "Input": "femtio - två",
```
tellarin, line 772: I'd prefer if we had some normalization for this case into "femtio två" for the lookup, so we could still support it.

Specs/Number/Swedish/NumberModel.json
```diff
@@ -882,22 +866,6 @@
       }
     ]
   },
-  {
-    "Input": "femtio-två",
```
tellarin, line 886: Same.

Specs/Number/Swedish/NumberModel.json
```diff
@@ -1376,30 +1358,30 @@
           "value": "0,0476190476190476"
         },
         "Start": 0,
-        "End": 17
+        "End": 16
       }
     ]
   },
   {
-    "Input": "en tjugo femtedel",
```
tellarin, line 1384: "en tjugo femtedel" is not a valid case?

Specs/Number/Swedish/NumberModel.json
```diff
@@ -1376,30 +1358,30 @@
...
+        "End": 17
       }
     ]
   },
   {
     "Input": "tre tjugoförstedelar",
-    "Comment": "Correspons to writing 'Three twentyfirsts'. Compound fraction expressions are currently not handled.",
-    "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript, python, java",
+    "Debug": true,
```
tellarin, line 1384: Please be careful to not leave the "Debug" markers in the specs in the repo.

tellarin (changes requested):
Looks mostly good to me. But couldn't we support the hyphen cases with a simple normalized lookup?

Specs/Number/Swedish/NumberModel.json
pete1854: This has been fixed

Specs/Number/Swedish/NumberModel.json
pete1854: This case as well as "femtio - två" is a very strange construct. This is not how numbers are written in swedish. I could stretch the recognizer to support "femtio två" (which is also not correct since swedish numbers are written as compound expressions). Should we really encourage completely erroneous grammar?

Specs/Number/Swedish/NumberModel.json
pete1854: This has been removed

Specs/Number/Swedish/NumberModel.json
tellarin: Feel free to remove cases if they really would never show up in practice.
If you can add a list of cases removed with reasons, that helps a lot. Another option is to re-writte really incorrect sentences into new cases, if relevant.
Thanks!

pete1854: Ok. Is there any reason we'd like to support completely erroneous constructs? In an ideal world the recognizer would understand the language in question - not all kind of wild constructs (that might be completeley correct in another language). Supporting non-language specific grammar, in my mind, increases the risk of malinterpretations of the text that is processed. Or, am I missing the point? :)

pete1854:
Is there anything else I should do for this PR to be accepted? I also have no clue as to why the build fails. Seems to be something that relates to JS which I have not touched in my PR.

pete1854:
I have made further tests pertaining to fraction handling. I have hoped that the current Ordinal implementation would solve issues regarding the correct resolution of compound fraction expressions, e.g. "två hundratjugosjättedelar" or "nitton femhundraåttiotredjedels". It has turned out that the recognizer cannot currently resolve these types of expressions.

I'm struggling to understand the relationship between the ordinal patterns and how they can be reused for fractional expressions.

An ordinal expression such as "den hundratjugosjätte" resolves as expected and I get the notion that I "only" need to add something to these expressions to denote a fraction [e.g. "del(s|ar|en)?"], or something similar.

If I can trick the extractor to believe it has extracted an ordinal type I can see that it resolves the fractional part correctly ("hundratjugosjättedels" = 126).

I have tried to look into the dutch and german solutions, but, the swedish language is even "more" compund i nature than these. Also, they don't seem to try and test for these complex expressions at all.

Do you have any ideas how I should go about solving this? I would like to solve this so that I can continue and create PRs for range and datetime understanding which are about to be ready.

tellarin:
@pete1854 Thanks a lot for all the updates! I'll take a closer look at the changes and your question soon. Sorry for the delay.
I believe we can merge most of it soon.

Specs/Number/Swedish/NumberModel.json
tellarin: Clearly erroneous cases should not be supported. But malformed cases that may be common-place should, if they don't interfere with regular cases. For example, common typos or very informal language/format, or formats that may not be native that are somewhat commonly used by non-native speakers. 
This is common in English, for example, as people format numbers/datetimes in different styles in international contexts.
Anything "completetly erroneus" or "wild" can be removed. :-)

.NET/Microsoft.Recognizers.Definitions.Common/Swedish/QuotedTextDefinitions.cs
```diff
@@ -25,7 +25,7 @@ public static class QuotedTextDefinitions
       public const string QuotedTextRegex1 = @"(“([^“”]+)”)";
       public const string QuotedTextRegex2 = @"(‘([^‘’]+)’)";
       public const string QuotedTextRegex3 = @"(""([^""]+)"")";
-      public const string QuotedTextRegex4 = @"('([^']+)')";
+      public const string QuotedTextRegex4 = @"(\\'([^\']+)\\')";
```
tellarin, line 28: Can you revert this change? This escaping is invalid and is not automatically generated in the CI build.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/Swedish/QuotedTextDefinitions.cs
tellarin: I just went ahead and reverted and re-triggered the PR build.

tellarin (approved):

Issue #2332, opened 2020-10-26
[JavaScript | EN DateTime V2] Years spelled out as words resolve to '0001-01-01'


apolya01:
**Describe the bug**
Phrases that may be year numbers written out as words ("two thousand twenty", "nineteen ninety-nine") resolve to 
"{"start\":"0001-01-01","end":"0001-01-01"}"

**To Reproduce**
- Call `Recognizers.recognizeDateTime("two thousand twenty", Recognizers.Culture.English)`
- Observe that `results[0].resolution.values[0]` has 'start', 'end' fields of unexpected value
- no other result/resolution is present

**Expected behavior**
These kinds of phrases, if they are to be resolved, should resolve to accurate date values; e.g. "two thousand twenty" should resolve to `"{"start":"2020-01-01","end":"2020-12-31"}"`

**Sample input/output**
`Recognizers.recognizeDateTime("two thousand twenty", Recognizers.Culture.English)`

```
"resolution":{"values":[{"timex":"0000","type":"daterange","start":"0001-01-01","end":"0001-01-01"}]},"text":"two thousand and twenty","typeName":"datetimeV2.daterange"}'
```

**Platform (please complete the following information):**
 - Platform: JavaScript
 - Environment: npm package
 - Version of package: 1.3.0

**Additional context**

PR #2373, opened 2020-11-17
[JavaScript/Python | EN DateTime V2] Fixed missing resolution for years spelled out as words (#2332)

LionbridgeCSII:
Added support for years spelled out as words (e.g. "two thousand twenty") in DatePeriod in Javascript and Python (#2332).


tellarin (commented):
Looks good so far. We'll complete the review once the builds pass.
But please prioritize the issues for .NET unless some specific issue is highlighted, ok?
Thanks!

LionbridgeCSII:
> But please prioritize the issues for .NET unless some specific issue is highlighted, ok?

Sure, I worked on this one just because there were no new .NET issues (and the old ones are mostly labelled as open questions, so it is not clear if they should be addressed).

tellarin (approved):

PR #2377, opened 2020-11-18
[ZH Dimension|Number] Handle ambiguity caused by "两" (#2324)

### LINKED ISSUE #2324: [ZH Number / Dimension] Number ending with "两"  should not be recognized as Number but Dimension，such as "两万两"、"二两"、"三十两"、"三千两百两" etc.
**Describe the bug**
"两" is unit of mass in Chinese.  
Number ending with "两"  should not be recognized as Number but Dimension，such as "两万两"、"二两"、"三十两"、"三千两百两" etc.

**Expected behavior**
```
{
        "Text": "两万两",
        "Start": 0,
        "End": 2,
        "TypeName": "dimension",
        "Resolution": {
            "unit": "Liang",
            "value": "20000"
        }
 }

{
        "Text": "二两",
        "Start": 0,
        "End": 1,
        "TypeName": "dimension",
        "Resolution": {
            "unit": "Liang",
            "value": "2"
        }
 }

{
        "Text": "三十两",
        "Start": 0,
        "End": 2,
        "TypeName": "dimension",
        "Resolution": {
            "unit": "Liang",
            "value": "30"
        }
 }

{
        "Text": "三千两百两",
        "Start": 0,
        "End": 4,
        "TypeName": "dimension",
        "Resolution": {
            "unit": "Liang",
            "value": "3200"
        }
 }

{
        "Text": "1234两",
        "Start": 0,
        "End": 4,
        "TypeName": "dimension",
        "Resolution": {
            "unit": "Liang",
            "value": "1234"
        }
 }
```
 - Platform: [NET, JavaScript, Python, Java...]
 - Version of package [v1.4.1]

**Additional context**
Particularly,  _"两两"_  is neither a number nor a dimension, it means "something in pair".

Especially,  if _"两"_ is used alone, it still means "two".
```
{
  "Text": "两",
  "Start": 0,
  "End": 0,
  "TypeName": "number",
  "Resolution": {
    "value": "2"
  }
}
```

ChopperMan33:
Fixed #2324 (conflicts between number "两"(two) and unit "两"(50g)).
Separately extract single number "两" and unit "两".
For example, in cases like "两万两" or "两两".
Added test cases to DimensionModel.json.


Patterns/Chinese/Chinese-Numbers.yaml
```diff
@@ -188,15 +188,21 @@ PercentageNumRegex: !simpleRegex
...
+ContinuouslyNumberRegex: !nestedRegex
```
tellarin, line 191: Can you add a comment in the file here to explain the need for ContinuouslyNumberRegex and SingleLiangRegex and to let people implementing other languages that they should remove these two if they don't apply?

tellarin (changes requested):

Patterns/Chinese/Chinese-Numbers.yaml
ChopperMan33: I have added a comment for it.

tellarin (approved):

PR #2380, opened 2020-11-20
[ZH NumberRange] Added support for number pattern "万多亿" (#2365)

### LINKED ISSUE #2365: [ZH Numberrange] "三十万多亿"，"四百六十万多亿" etc.  numberrange results are error, but "三十一万多亿", "四百六十一万多亿" can be recognized correctly.
**Describe the bug**
"三十万多亿" equals "30万多亿" , "三十万多亿" numberrange result is "(300000,)", it's an error result, 
but "30万多亿" can be recognized correctly as "(30000000000000,)".

Strangely， "三十一万多亿" equals "31万多亿"， both of them can be recognized correctly.
Similar examples, such as "四百六十万多亿" and "四百六十一万多亿".

**Expected behavior**
{
  "Text": "三十万多亿",
  "Start": 0,
  "End": 4,
  "TypeName": "numberrange",
  "Resolution": {
    "value": "(30000000000000,)"
  }
}

{
  "Text": "四百六十万多亿",
  "Start": 0,
  "End": 6,
  "TypeName": "numberrange",
  "Resolution": {
    "value": "(460000000000000,)"
  }
}

 - Platform: [ .NET...]
 - Version of package [v1.4.2]

**Additional context**

ChopperMan33:
Fixed #2365 (can't parse the pattern "万多亿") by sorting Config.UnitMap in BaseCJKNumberParser.
For example, in cases like "三十万多亿".
Added test cases to NumberRangeModel.json.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
@@ -498,8 +498,7 @@ namespace Microsoft.Recognizers.Text.Number
         // Parse unit phrase. "万", "億",...
         private string ReplaceUnit(string resultText)
         {
-            var unitMap = Config.UnitMap.OrderBy(o => o.Key.Length).ToDictionary(o => o.Key, p => p.Value);
-            foreach (var unit in unitMap.Keys)
+            foreach (var unit in Config.UnitMap.Keys)
             {
                 resultText = resultText.Replace(unit, Config.UnitMap[unit]);
             }
--- a/.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
@@ -498,7 +498,8 @@ namespace Microsoft.Recognizers.Text.Number
         // Parse unit phrase. "万", "億",...
         private string ReplaceUnit(string resultText)
         {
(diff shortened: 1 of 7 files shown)


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -498,7 +498,8 @@ private string NormalizeCharWidth(string text)
         // Parse unit phrase. "万", "億",...
         private string ReplaceUnit(string resultText)
         {
-            foreach (var unit in Config.UnitMap.Keys)
+            var unitMap = Config.UnitMap.OrderBy(o => o.Key.Length).ToDictionary(o => o.Key, p => p.Value);
```
tellarin, line 501: This should only happen once in initialization, can you move it out of this method?

tellarin: Same in other platforms.

tellarin (changes requested):

Specs/DateTime/Chinese/DurationExtractor.json
```diff
@@ -75,5 +75,17 @@
         "Length": 4
       }
     ]
+  },
+  {
+    "Input": "还剩8天20时",
+    "NotSupported": "dotnet,javascript,python,java",
```
tellarin, line 81: Why are these cases marked as NotSupported in .NET?

tellarin (approved):
Approving as the spec comment is for an unrelated issue.

Issue #2381, opened 2020-11-24
[ZH DateTimeV2] Issues in recognizing dates like “三十日”, “三十一日”, and ”一月二十三“


ChopperMan33:
**Describe the bug**
1. “日” in “三十日” or “三十一日” is not be extracted. The model can correctly recognize the date, but will miss the word "日“ in the sentence.
2. ”一月二十三“ (23/Jan) will be incorrectly recognized as "一月二”（02/Jan).

**Expected behavior**
```json
 {
    "Input": "五月三十日",
    "Results": [
      {
        "Text": "五月三十日",
        "Type": "date",
        "Start": 0,
        "Length": 5
      }
    ]
  }，
 {
    "Input": "五月三十一日",
    "Results": [
      {
        "Text": "五月三十一日",
        "Type": "date",
        "Start": 0,
        "Length": 6
      }
    ]
  },
{
    "Input": "一月二十三",
    "Context": {
      "ReferenceDateTime": "2020-11-22T00:00:00"
    },
    "Results": [
      {
        "Text": "一月二十三",
        "Type": "date",
        "Value": {
          "Timex": "XXXX-01-23",
          "FutureResolution": {
            "startDate": "2021-01-23",
            "endDate": "2021-01-23"
          },
          "PastResolution": {
            "startDate": "2020-01-23",
            "endDate": "2020-01-23"
          }
        },
        "Start": 0,
        "Length": 5
      }
    ]
  }
```

**Sample input/output**
```json
{
    "Input": "五月三十日",
    "Results": [
      {
        "Text": "五月三十",
        "Type": "date",
        "Start": 0,
        "Length": 5
      }
    ]
  }，
{
    "Input": "一月二十三",
    "Context": {
      "ReferenceDateTime": "2020-11-22T00:00:00"
    },
    "Results": [
      {
        "Text": "一月二",
        "Type": "date",
        "Value": {
          "Timex": "XXXX-01-02",
          "FutureResolution": {
            "startDate": "2021-01-02",
            "endDate": "2021-01-02"
          },
          "PastResolution": {
            "startDate": "2020-01-02",
            "endDate": "2020-01-02"
          }
        },
        "Start": 0,
        "Length": 3
      }
    ]
  }
```

**Platform (please complete the following information):**
 - Platform: [.NET, JavaScript, Python, Java]
 - Environment: [nuget package, npm package, LUIS, console app, in browser]
 - Version of package [v1.4.2]

Issue #2248, opened 2020-08-19
[ZH DateTimeV2] Add support for Chinese dynasties


Green-li:
Will support the analysis of Chinese dynasties?

## Example
Chinese sentence:`神龙元年， 十一月，武则天去世。` (`Wu Zetian died in November of the first year of the dragon.`)
The date in sentence is `神龙元年， 十一月`. In ISO 8601, it should be `0705-11-01`.

tellarin:
@Green-li Would you be interested in helping contribute support?

Green-li:
@tellarin Sorry, I'm busy recently. I will fork this repo and try to add this feature when free.

tellarin:
@Green-li Any contribution is greatly appreciated. Let us know how we can facilitate. Thanks.

PR #2382, opened 2020-11-24
[ZH DateTimeV2] Supported recognizing dynasties, fixed recognizing dates like “三十日” (#2248 #2381)

ChopperMan33:
Supported recognizing Chinese dynasties like "康熙元年“ (#2248).
Fixed recognizing dates like “三十日” (#2381).
Added test cases.


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
```diff
@@ -739,6 +739,22 @@ private int ConvertChineseYearToInteger(string yearChsStr)
...
             var num = 0;
 
+            var regionTitleMatch = this.config.RegionTitleRegex.MatchExact(yearChsStr, trim: true);
+            if (regionTitleMatch != null && regionTitleMatch.Value != string.Empty)
+            {
+                // handle "康熙元年" refer to https://zh.wikipedia.org/wiki/%E5%B9%B4%E5%8F%B7
+                int basicYear = this.config.DynastyYearMap[regionTitleMatch.Value];
+                string biasYearStr = yearChsStr.Substring(regionTitleMatch.Value.Length, yearChsStr.Length - regionTitleMatch.Length);
+                int biasYear = 1;
+                if (biasYearStr != "元")
```
tellarin, line 749: Please move hardcoded values such as "元" to the resources YAML file.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
```diff
@@ -739,6 +739,22 @@ private int ConvertChineseYearToInteger(string yearChsStr)
             var year = 0;
             var num = 0;
 
+            var regionTitleMatch = this.config.RegionTitleRegex.MatchExact(yearChsStr, trim: true);
+            if (regionTitleMatch != null && regionTitleMatch.Value != string.Empty)
+            {
+                // handle "康熙元年" refer to https://zh.wikipedia.org/wiki/%E5%B9%B4%E5%8F%B7
+                int basicYear = this.config.DynastyYearMap[regionTitleMatch.Value];
+                string biasYearStr = yearChsStr.Substring(regionTitleMatch.Value.Length, yearChsStr.Length - regionTitleMatch.Length);
```
tellarin, line 747: Instead of parsing the value like these, can't named groups be used in the regex so you capture what you need?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -207,6 +207,22 @@ private int ConvertChineseToInteger(string yearChsStr)
             var year = 0;
             var num = 0;
 
+            var regionTitleMatch = ChineseDateExtractorConfiguration.RegionTitleRegex.MatchExact(yearChsStr, trim: true);
```
tellarin, line 210: Ideally we should not have such duplicated code. Can't it be restructured to re-use at least a common method?

Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -7,9 +7,9 @@ MonthRegex: !simpleRegex
...
+  def: (?<day>二十一日|三十一日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|二十一日|三十一日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|二十一号|三十 一号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|二十一号|三十一号|二十二 号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|十一日|十二日|十三日|十四日|十五日| 十六日|十七日|十八日|十九日|十一日|二十日|三十日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十日|三十日|10日|11日|12日|13日|14日|15日|16日|17日|18日|19日|20日|21日|22日|23日|24日|25日|26日|27日|28日|29日|30日|31日|十一号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|十一号|二十号|三十号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十号|三十号|10号|11号|12号|13号|14号|15号|16号|17号|18号|19号|20号|21号|22号|23号|24号|25号|26号|27号|28号|29号|30号|31号|初一|三十|一日|二日|三日|四日|五日|六日|七日|八日|九日|一日|十日|二日|三日|四日|五日|六日|七日|八日|九日|十日|1日|2日|3日|4日|5日|6日|7日|8日|9日|一号|二号|三号|四号|五号|六号|七号|八号|九号|一号|十号|二号|三号|四号|五号|六号|七号|八号|九号|十号|1号|2号|3号|4号|5号|6号|7号|8号|9号)
```
tellarin, line 10: Can you explain this change and the DayRegexNumInChinese one?

Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -19,9 +19,14 @@ SimpleYearRegex: !simpleRegex
...
   def: '[一二三四五六七八九零壹贰叁肆伍陆柒捌玖〇两千俩倆仨]'
+RegionTitleRegex: !simpleRegex
+  def: (贞观|开元|神龙|洪武|建文|永乐|景泰|天顺|成化|嘉靖|万历|崇祯|顺治|康熙|雍正|乾隆|嘉庆|道光|咸丰|同治|光绪|宣统|民国)
+DynastyYearRegex: !nestedRegex
+  def: ({RegionTitleRegex}(元|\d{1,3}|[十拾]?({ZeroToNineIntegerRegexChs}[十百拾佰]?){0,3}))
+  references: [RegionTitleRegex, ZeroToNineIntegerRegexChs]
 DateYearInChineseRegex: !nestedRegex
-  def: (?<yearchs>({ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}|{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}|{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}))
-  references: [ZeroToNineIntegerRegexChs]
+  def: (?<yearchs>({DynastyYearRegex}|{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}|{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}|{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}{ZeroToNineIntegerRegexChs}))
```
tellarin, line 28: Does {DynastyYearRegex} need to be the first?

Python/libraries/recognizers-date-time/recognizers_date_time/date_time/french/dateperiod_extractor_config.py
```diff
@@ -209,6 +209,14 @@ def week_day_of_month_regex(self) -> Pattern:
     def all_half_year_regex(self) -> Pattern:
         return self._all_half_year_regex
 
+    @property
+    def region_title_regex(self) -> Pattern:
```
tellarin, line 213: This should not happen. We should not add such new attributes to all languages, unless it applies in most.
Can it be changed to affect only CJK?

Specs/DateTime/Chinese/DatePeriodParser.json
```diff
@@ -1458,5 +1458,180 @@
         "Length": 9
       }
     ]
+  },
+  {
+    "Input": "神龙元年十一月，武则天去世",
```
tellarin, line 1463: When adding new specs in DateTime, please add a couple examples per sub-type extractor/parser, but always add relevant cases in DateTimeModel.json too, ok?

JavaScript/packages/recognizers-date-time/src/dateTime/english/dateConfiguration.ts
```diff
@@ -126,6 +126,8 @@ export class EnglishDateParserConfiguration implements IDateParserConfiguration
     readonly relativeWeekDayRegex: RegExp;
     readonly utilityConfiguration: IDateTimeUtilityConfiguration
     readonly dateTokenPrefix: string
+    readonly regionTitleRegex: RegExp;
```
tellarin, line 129: Similar to the Python comment. In C# this didn't happen either.

tellarin (changes requested):

Patterns/Chinese/Chinese-DateTime.yaml
ChopperMan33: > Can you explain this change and the DayRegexNumInChinese one?

This change fixed the issue #2381 . I changed their order, putted “三十” behind “三十一“ to avoid recognizing  ”五月三十一日“，”五月三十一“ as "五月三十”.

tellarin: Wouldn't it be better to simplify this a little? It seems to me that there's quite a bit of repetition.
For example for the case you said, why not use "五月三十(一?)" ? Similarly for quite a lot of cases.

ChopperMan33: Yes. I think I can simplify it.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
```diff
@@ -738,21 +738,15 @@ private int ConvertChineseYearToInteger(string yearChsStr)
...
-                string biasYearStr = yearChsStr.Substring(regionTitleMatch.Value.Length, yearChsStr.Length - regionTitleMatch.Length);
-                int biasYear = 1;
-                if (biasYearStr != "元")
-                {
-                    biasYear = Convert.ToInt32((double)(numberParser.Parse(integerExtractor.Extract(biasYearStr)[0]).Value ?? 0));
-                }
-
-                year = basicYear + biasYear - 1;
-                return year;
+            int dynastyYear = DateTimeFormatUtil.ParserDynastyYear(yearChsStr,
```
tellarin, line 741: Rename to ParseChineseDynastyYear.
Nitpick: Fix code indentation to match current code.

tellarin: Perhaps this method could be in some common Chinese config used by both Date and DatePeriod parsers? 
Then is could be named ParseDynastyYear.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
ChopperMan33: Thanks for your suggestions. I have renamed it.

tellarin (approved):

Issue #2384, opened 2020-12-02
[ZH DateTimeV2]Recognition of combined duration like "8天20小时"


ChopperMan33:
**Describe the bug**
Cannot correctly recognize combined duration like "8天20小时" in Chinese.

**Expected behavior**
```json
{
    "Input": "还剩8天20小时",
    "Context": {
      "ReferenceDateTime": "2017-03-22T00:00:00"
    },
    "Results": [
      {
        "Text": "8天20小时",
        "Type": "duration",
        "Value": {
          "Timex": "P8D",
          "FutureResolution": {
            "duration": "691200"
          },
          "PastResolution": {
            "duration": "691200"
          }
        },
        "Start": 2,
        "Length": 2
      }
    ]
  }
```
**Sample input/output**
```json
  {
    "Input": "还剩8天20小时",
    "Context": {
      "ReferenceDateTime": "2017-03-22T00:00:00"
    },
    "Results": [
      {
        "Text": "8天",
        "Type": "duration",
        "Value": {
          "Timex": "P8D",
          "FutureResolution": {
            "duration": "691200"
          },
          "PastResolution": {
            "duration": "691200"
          }
        },
        "Start": 2,
        "Length": 2
      },
      {
        "Text": "20小时",
        "Type": "duration",
        "Value": {
          "Timex": "PT20H",
          "FutureResolution": {
            "duration": "72000"
          },
          "PastResolution": {
            "duration": "72000"
          }
        },
        "Start": 4,
        "Length": 4
      }
    ]
  }
```

**Platform (please complete the following information):**
- Platform: [.NET, JavaScript, Python, Java]
- Environment: [nuget package, npm package, LUIS, console app, in browser]
- Version of package [v1.4.2]

PR #2385, opened 2020-12-02
[ZH DateTimeV2] Supported recognition of combined durations like "8天20小时"(#2384 )

ChopperMan33:
Fixed recognizing of combined duration like "8天20小时" in Chinese .NET (#2384).
Added test cases.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Globalization;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 using Microsoft.Recognizers.Definitions.Chinese;
@@ -30,6 +31,21 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
 
         private static readonly Regex HalfSuffixRegex = new Regex(DateTimeDefinitions.DurationHalfSuffixRegex, RegexFlags);
 
+        private static readonly Regex DurationUnitRegex = new Regex(DateTimeDefinitions.DurationUnitRegex, RegexFlags);
+
+        private static readonly Regex DurationConnectorRegex = new Regex(DateTimeDefinitions.DurationConnectorRegex, RegexFlags);
+
+        private static readonly Dictionary<string, string> UnitMap = DateTimeDefinitions.ParserConfigurationUnitMap.ToDictionary(k => k.Key, k => k.Value.Substring(0, 1) + k.Value.Substring(1).ToLower());
+
+        private static readonly Dictionary<string, long> UnitValueMap = DateTimeDefinitions.DurationUnitValueMap;
+
+        private readonly bool merge;
+
+        public ChineseDurationExtractorConfiguration(bool merge = true)
+        {
+            this.merge = merge;
+        }
+
         internal override ImmutableDictionary<Regex, DurationType> Regexes { get; }
 
         protected sealed override string ExtractType { get; } = Constants.SYS_DATETIME_DURATION; // "Duration";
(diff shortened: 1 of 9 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
```diff
@@ -52,9 +68,126 @@ public override List<ExtractResult> Extract(string source, DateObject referenceT
...
 
+            if (this.merge)
+            {
+                res = MergeMultipleDuration(source, res);
+            }
+
             return res;
         }
 
+        private List<ExtractResult> MergeMultipleDuration(string text, List<ExtractResult> extractorResults)
```
tellarin, line 79: Shouldn't merging of durations be at the parser level? Usually the extractors are lighter and parsers are the ones that rely in deeper semantics. What would be the drawback of moving the merge there?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
```diff
@@ -30,6 +31,21 @@ public class ChineseDurationExtractorConfiguration : ChineseBaseDateTimeExtracto
...
+
+        private static readonly Dictionary<string, string> UnitMap = DateTimeDefinitions.ParserConfigurationUnitMap.ToDictionary(k => k.Key, k => k.Value.Substring(0, 1) + k.Value.Substring(1).ToLower());
+
+        private static readonly Dictionary<string, long> UnitValueMap = DateTimeDefinitions.DurationUnitValueMap;
+
+        private readonly bool merge;
+
+        public ChineseDurationExtractorConfiguration(bool merge = true)
+        {
+            this.merge = merge;
```
tellarin, line 46: Why is this flag necessary?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDurationParserConfiguration.cs
```diff
@@ -82,6 +97,78 @@ public List<DateTimeParseResult> FilterResults(string query, List<DateTimeParseR
             return candidateResults;
         }
 
+        private DateTimeResolutionResult ParseMergedDuration(string text, DateObject referenceTime)
+        {
+            var ret = new DateTimeResolutionResult();
+            var durationExtractor = DurationExtractor;
+
+            // DurationExtractor without parameter will not extract merged duration
+            var ers = durationExtractor.Extract(text, referenceTime);
```
tellarin, line 106: This seems to be duplicating extraction work, which suggest it would be better to have the merge happen at the parser level.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDurationParserConfiguration.cs
```diff
@@ -33,33 +45,36 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
...
+                    return null;
+                }
 
-            dateTimeParseResult.Timex = "P" + (BaseDurationParser.IsLessThanDay(unitStr) ? "T" : string.Empty) + numStr + unitStr[0];
-            dateTimeParseResult.FutureValue = dateTimeParseResult.PastValue = double.Parse(numStr) * UnitValueMap[unitStr];
-            dateTimeParseResult.Success = true;
+                var unitStr = unitResult.Unit;
+                var numStr = unitResult.Number;
+
+                dateTimeParseResult.Timex = "P" + (BaseDurationParser.IsLessThanDay(unitStr) ? "T" : string.Empty) + numStr + unitStr[0];
```
tellarin, line 62: As you're touching this code, can we also try to refactor it so timex generation goes to a "centralized" module?

Specs/DateTime/Chinese/DurationExtractor.json
```diff
@@ -77,14 +77,86 @@
     ]
   },
   {
-    "Input": "还剩8天20时",
```
tellarin, line 80: Why was this originally incorrect?

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -5915,5 +5915,173 @@
         "End": 7
       }
     ]
+  },
+  {
+    "Input": "还剩8天20小时",
```
tellarin, line 5920: You don't need to add all the same cases to all three spec files. DateTimeModel can have the most, to show that the final behaviour is correct and that there is a reasonable coverage of different situations. The extractor/parser ones can have a few cases each, especially for corner cases or cases documenting some special behaviour.
Ideally the spec cases in the extractor/parser specs should form pairs though, to avoid cases were extraction looks good, but parsing may fail.
Basically there should be some intersection between the cases in DateTimeModel and *Extractor/*Parser pairs, but not total. This is just a suggestion to reduce manual duplication of cases.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
ChopperMan33: We have test cases in DurationExtractor.json for extractors which also need to return a complete text like "8天20小时“ as same as parser result. This part is the same in English.
 Refer to https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs#L41
Can we make the results text of extractor and parser different?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
ChopperMan33: As mentioned above, extractor returns a complete text like "8天20小时“ to parser. Parser calls extractor to separate it into "8天“ and "20小时“, then recognize both value separately. At that time, it uses this flag to avoid extractor from using the MergeMultipleDuration function.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDurationParserConfiguration.cs
ChopperMan33: OK, I will try to modify it.

Specs/DateTime/Chinese/DurationExtractor.json
ChopperMan33: "时“ is not in DurationSuffixList regex, it is not a duration unit. "20时“ means 20 o'clock. 个小时|小时|个钟头|钟头 means hours.
I think the previous case was a bit inaccurate, so I modified it.

Specs/DateTime/Chinese/DateTimeModel.json
ChopperMan33: Thank you, next time I will choose some representative cases to add to DateTimeModel.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
tellarin: There is no problem making the intermediary results distinct. The extractor/parser spec specify each step behaviour.
The model spec specifies the whole pipeline behaviour and final output.
But if the current implementation in the PR follow the existing English one, let's not modify this now.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDurationExtractorConfiguration.cs
tellarin: Is this the same behaviour as in English?

ChopperMan33: Yes. The code is in https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs#L19

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -5915,5 +5915,197 @@
         "End": 7
       }
     ]
+  },
+  {
+    "Input": "还剩8天20时",
+    "NotSupported": "javascript,python,java",
```
tellarin, line 5921: Shouldn't this be enabled in javascript and python? You've added code for it.

tellarin (approved):

Issue #2367, opened 2020-11-16
[EN DateTimeV2] "end of" by itself incorrectly extracted as entity


tellarin:
Cases like "We've reached the end of the line" lead to incorrect output tagging "end of" as a datetime entity.
This should be fixed, along with an analysis of similar situations.

LionbridgeCSII:
@tellarin, I am unable to reproduce this issue (neither in visual studio nor in simple console).

tellarin:
Interesting. Some recent commit may have partially addressed some cases. But if the input is just "end of" I still see it happening in the current /master.

LionbridgeCSII:
> Interesting. Some recent commit may have partially addressed some cases. But if the input is just "end of" or "end of the" I still see it happening in the current /master.

ah yes, it is true. I will check it after finishing #2371.

tellarin:
Please also add the 3 cases listed here as new specs, just to avoid regressions later. Thanks.

PR #2387, opened 2020-12-03
[EN DateTimeV2] Fixed "end of" by itself incorrectly extracted as entity (#2367)

LionbridgeCSII:
Fix to issue #2367.


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
```diff
@@ -307,19 +306,6 @@ public List<Token> SpecialTimeOfDate(string text, DateObject reference)
             return ret;
         }
 
-        // Special case for 'the end of today'
-        public List<Token> SpecialTimeOfDay(string text, DateObject reference)
-        {
-            var ret = new List<Token>();
-            var match = this.config.SpecificEndOfRegex.Match(text);
```
tellarin, line 314: I guess this regex is now references somewhere else?

tellarin (approved):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
LionbridgeCSII: yes, it is used in the method SpecialTimeOfDate which correctly extracts pattern like "the end of today".

Issue #2374, opened 2020-11-17
[DE DateTimeV2] Terms about years (last year, this year) are not recognized with values.


rostam:
**Describe the bug**
The expressions "letztes Jahr", "letzten Jahr", "letzte Jahr", "dieses Jahr", "diesen Jahr", "diesem Jahr", "diese Jahr"  produce wrong results and the expressions "vorletztes Jahr", "vorletzten Jahr", "vorletztem Jahr"  are not recognized at all. 

```
         {
            "entity": "... letzten jahr",
            "type": "builtin.datetimeV2.date",
            "startIndex": 26,
            "endIndex": 37,
            "resolution": {
                "values": [
                    {
                        "timex": "XXXX-00-00",
                        "type": "date",
                        "value": "not resolved"
                    }
                ]
            },
            "role": ""
        }
```

**To Reproduce**
1. Go to LUIS
2. Type any of the give expressions

**Expected behavior**
The results should be as follows,
- "letztes Jahr", "letzten Jahr", "letzte Jahr" -> last year
- "dieses Jahr", "diesen Jahr", "diesem Jahr", "diese Jahr" -> this year
- "vorletztes Jahr", "vorletzten Jahr", "vorletztem Jahr" -> two years ago

**Sample input/output**
As applicable, add examples of text input and json output to help explain the problem.

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: LUIS
 - Version of package: Last version

**Additional context**

tellarin:
@LionbridgeCSII, please add this one in the stack. Thanks.

PR #2388, opened 2020-12-03
[DE DateTimeV2] Fixed terms about years (last year, this year) not recognized correctly (#2374)

LionbridgeCSII:
Fix to issue #2374.
One test case moved from DateExtractor to DateTimeModel because it does not seem to represent a date: "Ich komme nächstes Jahr wieder." (I'll come again next year)

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Recognizers.Definitions.German
       public const bool CheckBothBeforeAfter = false;
       public const string TillRegex = @"(?<till>zu|bis\s*zum|zum|bis|bis\s*hin(\s*zum)?|--|-|—|——)";
       public const string RangeConnectorRegex = @"(?<and>und|--|-|—|——)";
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
```diff
@@ -99,7 +100,7 @@ public static class DateTimeDefinitions
       public static readonly string DateExtractor7 = $@"({DayRegex}\s*[\.]\s*{MonthNumRegex}[\.])";
       public static readonly string DateExtractor8 = $@"(?<=\b(am)\s+){DayRegex}[/\\\.]{MonthNumRegex}[/\\\.]({DateYearRegex})?\b";
       public static readonly string DateExtractor9 = $@"\b({DayRegex}\s*/\s*{MonthNumRegex}((\s+|\s*,\s*){DateYearRegex})?)\b";
-      public static readonly string DateExtractor10 = $@"\b({RelativeRegex}\s*jahr(\s*im)?({MonthNumRegex}|sommer|winter|frühling|herbst)?)\b";
```
tellarin, line 102: Where are such cases supported now?

tellarin (approved):

.NET/Microsoft.Recognizers.Definitions.Common/German/DateTimeDefinitions.cs
LionbridgeCSII: @tellarin, in English a pattern like "next year in summer" is extracted as 2 separate daterange entities. In German, it is processed in the same way by DatePeriod but it is also still wrongly extracted as a single entity by DateExtractor (but not parsed). 
I will remove this pattern from German DateExtractor to conform German to English.
Should I also modify both languages to extract "next year in summer" as a single daterange entity?

PR #2394, opened 2020-12-07
[Tools] Spec validation tool v1

ChopperMan33:
Build the basic framework of the format check tool.
Load JSON file path which supports simple wildcard character.
Check JSON valid and return the result.


.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
```
tellarin, line 3: Be careful to not make solution changes that are not essential to a PR. We should only change VS versions if really necessary.

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -812,6 +812,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Turkish", "Turkish", "{2F41
 		..\Specs\DateTime\Turkish\TimePeriodParser.json = ..\Specs\DateTime\Turkish\TimePeriodParser.json
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Tools", "Tools", "{F6B54840-FCA6-4824-A79B-B05B186B0DA8}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FormatCheckTool", "..\Tools\src\FormatCheckTool\FormatCheckTool.csproj", "{D8836462-EBA2-4DF2-B23D-17996776FF96}"
```
tellarin, line 817: I'm fine with adding the tool in the overall .NET solution, especially for CI. But I suggest also creating a standalone solution only with the tool in its directory. People changing the tool may not need to load everything.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+using Microsoft.Recognizers.Text;
+using Microsoft.Recognizers.Text.Choice;
+using Microsoft.Recognizers.Text.DateTime;
+using Microsoft.Recognizers.Text.Number;
+using Microsoft.Recognizers.Text.NumberWithUnit;
+using Microsoft.Recognizers.Text.Sequence;
+using Newtonsoft.Json;
+using Microsoft.Recognizers.Text.DataDrivenTests;
+
+namespace FormatCheckTool
```
tellarin, line 16: It's better to use a name space matching similar to that of the packages as this may in the future become its own DLL.
Maybe Microsoft.Recognizers.Text.Validation.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+{
+    class FormatCheckTool
```
tellarin, line 18: ValidationTool may be a better name as it's not only going to check for format issues.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            // Encoding for 'exotic' characters e.g. '€'
+            Encoding.RegisterProvider(CodePagesEncodingProvider.Instance);
+
+            ShowIntro();
+
+            while (true)
+            {
+                Console.WriteLine("Enter the specs JSON file path to be checked:");
+                var input = Console.ReadLine()?.Trim();
```
tellarin, line 32: For such tools, it's better to read arguments from the command line. This facilitates automation.
Let's also make sure that it can receive a specific file path, or a directory, from where the tool should load and validate each spec (one after the other).

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+                Console.WriteLine();
+
+                if (input?.ToLower(CultureInfo.InvariantCulture) == "exit")
```
tellarin, line 35: No need for this as the tool for now won't be interactive. I.e. one run per spec file or set of spec files.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+using Microsoft.Recognizers.Text.NumberWithUnit;
+using Microsoft.Recognizers.Text.Sequence;
+using Newtonsoft.Json;
+using Microsoft.Recognizers.Text.DataDrivenTests;
+
+namespace FormatCheckTool
+{
+    class FormatCheckTool
+    {
+        private static readonly String PleaseCheckStr = "Please Check it.";
```
tellarin, line 20: Messages to the log or to users should be more descriptive. This is too general and not informative.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+                Console.WriteLine();
+
+                if (input?.ToLower(CultureInfo.InvariantCulture) == "exit")
+                {
+                    break;
+                }
+
+                // Validate input
+                if (input?.Length > 0)
+                {
```
tellarin, line 42: This validation could be its own LoadSpecs method.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+                    {
+                        CheckPipeline(path);
+                    }
+
+                    Console.WriteLine();
+                }
+            }
+        }
+
+        private static void CheckPipeline(string filePath)
```
tellarin, line 65: Why not make the pipeline return a status?

tellarin: I'd prefer the method name to be RunPipeline or ExecutePipeline.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+                return;
+            }
+
+            bool stage3 = CheckModelResult(data.Item2);
+        }
+
+        private static Tuple<bool, List<TestModel>> LoadJsonFile(string filePath)
+        {
+            List<TestModel> specs = new List<TestModel>{ };
+            string flag = string.Empty;
```
tellarin, line 87: This is not really a flag, but a log message, right?

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+                {
+                    specs = (List<TestModel>)JsonConvert.DeserializeObject<IList<TestModel>>(rawData);
+                    if (!specs.Any())
+                    {
+                        flag = $"JSON File is Empty, {PleaseCheckStr}";
+                    }
+                }
+                catch (JsonReaderException jre)
+                {
+                    flag = $"JSON Reader Exception! {PleaseCheckStr}";
```
tellarin, line 107: For any exceptions, it would be better to log also some more details. At least the exception message/type and where it happened.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+            bool stage2 = CheckBasicContent(data.Item2);
+            if (!stage2)
+            {
+                return;
+            }
+
+            bool stage3 = CheckModelResult(data.Item2);
+        }
+
+        private static Tuple<bool, List<TestModel>> LoadJsonFile(string filePath)
```
tellarin, line 84: Wouldn't it be better to define an internal class to represent this return object?

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+
+        private static bool CheckBasicContent(List<TestModel> data)
+        {
+            List<string> res = new List<string> { };
+            foreach (TestModel spec in data)
+            {
+                res.Add(CheckOneBasicContentSpec(spec));
+            }
+
+            return ConsoleStageResult("Content Valid", string.Join("|", res), true);
```
tellarin, line 130: Why this join in results? It's better to keep a proper log.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        private static bool CheckModelResult(List<TestModel> data)
```
tellarin, line 133: Please add a comment here that this is jsut a stub for future implementation.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            string flag = string.Empty;
+            return ConsoleStageResult("Model Result", flag);
+        }
+
+        private static string CheckOneBasicContentSpec(TestModel spec)
```
tellarin, line 139: What do you mean by CheckOneBasicContentSpec?

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+            return flag;
+        }
+
+        private static string CheckOneBasicContentSpecResult(TestModel spec, object result)
+        {
+            var serializeResult = JsonConvert.SerializeObject(result);
+            var dynamicResult = JsonConvert.DeserializeObject<dynamic>(serializeResult);
+            if (dynamicResult["Text"] == null || dynamicResult["Start"] == null || (dynamicResult["End"] == null && dynamicResult["Length"] == null))
+            {
+                return "Result Parameter Loss";
```
tellarin, line 168: I don't get what this means. Error messages should be as informative as possible.

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+                    }
+                }
+            }
+
+            return flag;
+        }
+
+        private static string CheckOneBasicContentSpecResult(TestModel spec, object result)
+        {
+            var serializeResult = JsonConvert.SerializeObject(result);
```
tellarin, line 164: Why do we need to serialize and deserialize something here? Isn't the spec already deserialized at this stage?

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+            var startEndStr = spec.Input.Substring(start, end - start + 1);
+            if (dynamicResult["Text"] != startEndStr)
+            {
+                return "Index Error";
+            }
+
+            return string.Empty;
+        }
+
+        private static bool ConsoleStageResult(string stage, string flag, bool isList = false)
```
tellarin, line 182: This seems a bit complicated to me. What exactly is happening here that can't be represented in a list of log lines?

Tools/src/FormatCheckTool/FormatCheckTool.cs
```diff
@@ -0,0 +1,234 @@
...
+            Console.WriteLine($"{stage} Check {logStr}.");
+            return res;
+        }
+
+        /// <summary>
+        /// Introduction.
+        /// </summary>
+        private static void ShowIntro()
+        {
+            Console.WriteLine("Welcome to the Recognizers' JSON Format Check Tool console application!");
```
tellarin, line 222: We could summarize this intro to only mention which parameters the tool takes. And only show the intro if no parameters are passed to it.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.sln
ChopperMan33: I have revised all the problems below according to your suggestions. Could you help me to review it again? Thanks for your help.

tellarin (approved):
The returns from pipeline stages currently mix validation, logging, and visualization. This should be refactored to separate concerns in a future PR.
As this is for now a separate tool that already provides value to pre-validate new specs, let's merge as is and refactor as soon as possible.
Later we can turn this into a lib that can be called from a more complex spec processing tool.

Issue #640, opened 2018-06-21
DateTImeRanges don't use context as basis for establishing limits


NeilMacMullen:
First of all - _GREAT_ library - I'm using this extensively in a suite of command line tools that need to input time ranges and it's so much more intuitive than forcing the user to guess the correct syntax for datetime parsing etc.

That said, I think Recognizers.Text.DateTime could do a slightly better job with 'ambiguous; ranges.  For example, the input

> 2pm 1st jun 2000  until 2am on 2nd

Currently produces this resolution:

> timex=(2000-06-01T14,XXXX-XX-02T02,PT157788H) type=datetimerange start=2000-06-01 14:00:00 end=2018-06-02 02:00:00

whereas the end is almost certainly better interpreted as being in the same year as the start.  I do understand that this is ambiguous and there are contexts where a user may really intend that the end date here is relative to _now_  but currently there is no way to provide a policy to assist with interpretation.  Perhaps introducing the concept of a 'basis' date that defines the users expected context would be useful?

LionbridgeCSII:
@tellarin, should we look into this one?

tellarin:
@LionbridgeCSII, please take a stab at it after the newly opened issues. This will probably require bigger changes to solve in a more general way. But it would be great to at least estimate the effort.
The resolution code should also start to be refactored into its own module, piece by piece. Currently there's a bunch of duplication and mixed concerns.

PR #2395, opened 2020-12-09
[* DateTimeV2] Support for context for establishing limits in DatePeriod and DateTimeperiod in ExperimentaMode (#640)

LionbridgeCSII:
Fix to issue #640 where, in Experimental mode, ambiguous dates are recalculated using the other non-ambiguous date as reference (in DatePeriod and DateTimePeriod).
Test cases added to DateTimeModel and DateTimeModelExperimentalMode.

The use of context was already partially implemented in DatePeriod, but it needs to be improved because it produces inconsistent results in certain cases (e.g. "from 5/1/2015 till three days later").


tellarin:
@LionbridgeCSII, we'll review this soon. Is the behaviour in this PR now consistent between the default and experimental modes?

LionbridgeCSII:
> @LionbridgeCSII, we'll review this soon. Is the behaviour in this PR now consistent between the default and experimental modes?

@tellarin, I have not examined all possible scenarios yet, I wanted first to have a feedback on the approach (I am not sure, for example, if adding the new behaviour to experimental mode is appropriate or if it would be better to use another mode). 
But for the cases I have considered, the behaviour between the two modes seems to be consistent (taking into account that experimental mode adds 1 day to the ranges).

Also, I have not yet addressed the issue with the method ProcessDateEntityParsingResult in DatePeriod (which produces inconsistent results with certain ambiguous dates, e.g. "three days later").

Issue #2342, opened 2020-10-29
Age in english should not require the word "old" to match


chrimc62:
**Describe the bug**
If a bot asks "How old are you?" and you run against a LUIS model with a prebuilt age entity it should match something like "19 years".  Currently it will only match "19 years *old*".

**To Reproduce**
1. Create a LUIS model with prebuilt age and train it.
2. Test with an utterance like "19 years"

**Expected behavior**
Should be recognized as an age.

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: LUIS
 - Version of package ?

PR #2399, opened 2020-12-10
[EN Age] Fix to Age in English should not require the word "old" to match (#2342)

LionbridgeCSII:
Fix to issue #2342. 
Also added support for the pattern "I am 33." which is common and seems unambiguous when occurring at the end of sentence.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersWithUnitDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersWithUnitDefinitions.cs
@@ -23,11 +23,11 @@ namespace Microsoft.Recognizers.Definitions.English
     {
       public static readonly Dictionary<string, string> AgePrefixList = new Dictionary<string, string>
         {
-            { @"Age", @"Age|age" }
+            { @"Age", @"Age|age|age of|i am" }
         };
       public static readonly Dictionary<string, string> AgeSuffixList = new Dictionary<string, string>
         {
-            { @"Year", @"years old|year old|year-old|years-old|-year-old|-years-old|years of age|year of age|yo" },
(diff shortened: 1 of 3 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersWithUnitDefinitions.cs
```diff
@@ -23,11 +23,11 @@ public static class NumbersWithUnitDefinitions
     {
       public static readonly Dictionary<string, string> AgePrefixList = new Dictionary<string, string>
         {
-            { @"Age", @"Age|age" }
+            { @"Age", @"Age|age|age of|i am" }
         };
       public static readonly Dictionary<string, string> AgeSuffixList = new Dictionary<string, string>
         {
-            { @"Year", @"years old|year old|year-old|years-old|-year-old|-years-old|years of age|year of age|yo" },
+            { @"Year", @"years old|year old|year-old|years-old|-year-old|-years-old|years of age|year of age|yo|years" },
```
tellarin, line 30: This will lead to conflicts with durations in years. We'll need to consider carefully if we want this.

tellarin (changes requested):
Marking as "request changes" to avoid merging as more discussion is needed.

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersWithUnitDefinitions.cs
LionbridgeCSII: yes, I was thinking the same. 
It would be nice to be able to exclude it only when the Age Recognizer is called along with the DateTime Recognizer.

tellarin:
Closing due to introduced ambiguity.

Issue #2398, opened 2020-12-10
[Python] Enrich convenience of ModelResult for visualization and dict result


daviddwlee84:
When I get results from Recognizers-Text in Python, I would like to take a quick view of what I have extracted.
In addition, if we can easily get a dict result from it, this is useful that it using the same format of converting C# object through the Protobuf.
More specifically, I will make a pull request that implements `__repr__`, `__str__`, and `get_dict` methods for the `ModelResult` class.

PR #2402, opened 2020-12-10
[Python] Enrich convenience of ModelResult for visualization and dict result

daviddwlee84:
#2398 

* Implemented `__repr__`, `__str__`, and `get_dict` methods for the `ModelResult` class.
* Modify `README.md` for Python samples

### DIFF
--- a/Python/samples/README.md
+++ b/Python/samples/README.md
@@ -8,7 +8,7 @@ This sample demonstrate the combination of all Recognizers to extract possible v
 
 First, install its dependencies:
 
-```
+```bash
 cd .\samples\simple_console\
 pip install -r .\requirements.txt
 ```
@@ -33,7 +33,7 @@ So far, the available models are:
 
 ```Python
 # Use English for the Recognizers culture
-DEFAULT_CULTURE = Culture.English
+culture = Culture.English
 
 # Number recognizer - This function will find any number from the input
 # E.g "I have two apples" will return "2".
@@ -66,12 +66,21 @@ Recognizers.recognize_temperature(user_input, culture),
 # DateTime recognizer - This function will find any Date even if its write in colloquial language
 # E.g "I'll go back 8pm today" will return "2017-10-04 20:00:00"
 Recognizers.recognize_datetime(user_input, culture)
-````
+```
 
-All these models accept `user_input: str` as a string and returns a **List** of [ModelResult](../libraries/recognizers-text/recognizers_text/model.py#L10-L16):
+All these models accept `user_input: str`, `culture: str` and returns a **List** of [`ModelResult`](../libraries/recognizers-text/recognizers_text/model.py#L11-L48) object:
 
-````Python
-result = Recognizers.recognize_number("I have twenty apples")
+```Python
+results = Recognizers.recognize_number("I have twenty apples", culture)
+
+print(
+    'Returns:\n',
+    json.dumps(
+        results,
+        default=lambda o: o.__dict__,
+        indent='\t',
+        ensure_ascii=False)
+)
 
 # Returns:
 # [
@@ -82,7 +91,7 @@ result = Recognizers.recognize_number("I have twenty apples")
 # 			"value": "20"
 # 		},
 # 		"text": "twenty",
-# 		"typeName": "number"
+# 		"type_name": "number"
 # 	}
 # ]
-````
+```
--- a/Python/libraries/recognizers-text/recognizers_text/model.py
+++ b/Python/libraries/recognizers-text/recognizers_text/model.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from enum import Flag
-from typing import List, Dict, Generic, TypeVar, Callable, Optional
+from typing import List, Dict, Generic, TypeVar, Callable, Optional, Union, Any
 from collections import namedtuple
 
 from .culture import Culture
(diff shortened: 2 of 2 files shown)


Python/samples/README.md
```diff
@@ -82,7 +91,7 @@ result = Recognizers.recognize_number("I have twenty apples")
 # 			"value": "20"
 # 		},
 # 		"text": "twenty",
-# 		"typeName": "number"
+# 		"type_name": "number"
```
tellarin, line 94: If this is happening on the output, this is breaking behaviour. Schema and naming must be the same across platforms.

Python/samples/README.md
```diff
@@ -66,12 +66,21 @@ Recognizers.recognize_temperature(user_input, culture),
 # DateTime recognizer - This function will find any Date even if its write in colloquial language
 # E.g "I'll go back 8pm today" will return "2017-10-04 20:00:00"
 Recognizers.recognize_datetime(user_input, culture)
-````
+```
 
-All these models accept `user_input: str` as a string and returns a **List** of [ModelResult](../libraries/recognizers-text/recognizers_text/model.py#L10-L16):
+All these models accept `user_input: str`, `culture: str` and returns a **List** of [`ModelResult`](../libraries/recognizers-text/recognizers_text/model.py#L11-L48) object:
```
tellarin, line 71: Shouldn't this say that culture is optional and if not provided English is used as default?

tellarin (changes requested):

Python/samples/README.md
daviddwlee84: I guessed this (what `README.md` was showing) is the output of the "[sample script](https://github.com/daviddwlee84/Recognizers-Text/blob/master/Python/samples/simple_console/sample.py#L36)".
As you can see, this is basically the naming of the `ModelResult`'s attribute that I didn't change.
That is, this won't break any behavior, and instead, the old `README.md` is a little bit misleading.

Python/samples/README.md
daviddwlee84: Currently, somehow [the `culture` argument is not "optional"](https://github.com/microsoft/Recognizers-Text/blob/master/Python/libraries/recognizers-number/recognizers_number/number/number_recognizer.py#L149) (which I expected it is).
I think we should note that this is not optional in the `README.md` currently and give it default culture in the future. (maybe in another pull request?)

Python/samples/README.md
tellarin: But the output of the sample script should be the same as the canonical representation of each result object. It currently doesn't behave correctly?

daviddwlee84: I didn't change the `ModelResult`'s attribute. [It was named `type_name`](https://github.com/microsoft/Recognizers-Text/blob/master/Python/libraries/recognizers-text/recognizers_text/model.py#L16) since I used this package. Thus not sure if this should be `typeName` or not. And this is the result of how I got from running the sample script.

Python/samples/README.md
daviddwlee84: #2410 issue

Python/samples/README.md
tellarin: Sorry, I was not aware of the discrepancy before and didn't have time to look into the issue in detail.
Can you make the output uniform? The objects returned by .NET and Python following the same schema.
Thanks for bringing this up while the Python package is still in alpha!

daviddwlee84: No problem! You mean following [the naming of .NET](https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text/Models/ExtendedModelResult.cs) right?
But usually, in Python, people use PEP 8 more often that it said ["Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability."](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables) (to make the code more "Python" :sweat_smile:)
So I did it in a compromised way that the `get_dict()` method returns `dict` with the same naming as .NET return. (If we change the schema of Python, that might influence lots of users currently)
Not sure which is the best way, looking for your opinion. :smile:

tellarin: We also use PEP8 in the project, so we'll need to add something to supress the naming issue just at this place.
Your ```get_dict()``` implements exactly what the behaviour should be. The key thing here is not so much the platform specifc naming conventions. Code in principle could follow whatever standard it needs.
But the returned object should be the same (for example, generate the same JSON, with the same schema, if serialized) and compatible between platforms. Which it currently is not, due to the issue you found.
I agree making this change will affect some users, but as the package is in alpha, this is the time to do it. We better be consistent from the start.

daviddwlee84: I see. How about like change the [sample code](https://github.com/microsoft/Recognizers-Text/blob/master/Python/samples/simple_console/sample.py#L33-L38) into `print(result.get_dict())`. Or simply change `lambda o: o.__dict__` into `lambda o: o.get_dict()`
In this way, the attribute name of the Python object can still follow PEP 8, at the same time, the output can follow the platform convention.

tellarin: That's not enough, as serializing the result object will still have the problem. Same as deserializing the json specs into an Python object.

daviddwlee84: Yeah, that makes sense. Do we want to modify this in this pull request or open another issue? This might have huge changes, that also influence unittests.

* https://github.com/microsoft/Recognizers-Text/blob/master/Python/tests/test_runner_number.py#L19
* https://github.com/microsoft/Recognizers-Text/blob/master/Python/libraries/recognizers-choice/recognizers_choice/choice/models.py#L33
* ...

tellarin: So that's how the tests currently pass. Not comparing the same names. :)
I agree. Let's make this change in a new PR. 
I'll do a final check on this one to merge. Thanks!

tellarin (approved):

daviddwlee84:
It is possible to introduce [`dataclasses-json`](https://github.com/lidatong/dataclasses-json) to solve the Pascal Case (C#) vs. Snake Case (Python) issue.

For example:

```py
from dataclasses import dataclass
from dataclasses_json import LetterCase, dataclass_json


# changing casing at the class level
@dataclass_json(letter_case=LetterCase.PASCAL)
@dataclass
class Person:
    given_name: str
    family_name: str
    
Person('Alice', 'Liddell').to_json()  # '{"GivenName": "Alice", "FamilyName": "Liddell"}'
Person.from_json('{"GivenName": "Alice", "FamilyName": "Liddell"}')  # Person('Alice', 'Liddell')
```

That we can keep Python object to follow PEP8 while consistent in serializing and deserializing to json

tellarin:
@daviddwlee84, thanks for the suggestion! Would you be able to create a PR using it to the repo?

daviddwlee84:
Sure, but I am quite busy these days. I will contribute it once I have time. :)

PR #2404, opened 2020-12-11
Add files via upload

XiaoxiaoMa0815:
Two files"HolidayExtractor.json""MergedExtractorSkipFromTo.json" in this directory
The other "PercentModelPercentMode.json” should in "/Number/French".

### DIFF
--- a/Specs/DateTime/French/PercentModelPercentMode.json
+++ b/Specs/DateTime/French/PercentModelPercentMode.json
@@ -1,0 +1,157 @@
+[
+    {
+        "Input": "100%",
+        "NotSupported": "javascript, python",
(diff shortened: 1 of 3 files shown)


Specs/DateTime/French/PercentModelPercentMode.json
```diff
@@ -0,0 +1,157 @@
+[
+    {
+        "Input": "100%",
```
tellarin, line 3: Please focus on the files that end in Extractor, Parser, and Model. The only ones extra at first should be the ones with ExperimentalMode. All other modes and extra suffixes can wait as they are lower priority.

PR #2405, opened 2020-12-11
Update HolidayExtractor.json

XiaoxiaoMa0815:
Update from line 145

### DIFF
--- a/Specs/DateTime/French/HolidayExtractor.json
+++ b/Specs/DateTime/French/HolidayExtractor.json
@@ -143,3 +143,130 @@
     ]
   }
 ]
+[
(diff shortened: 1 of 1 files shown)


Specs/DateTime/French/HolidayExtractor.json
```diff
@@ -142,4 +142,131 @@
       }
     ]
   }
-]
\ No newline at end of file
+]
```
tellarin, line 145: Almost right. Remove the closing and reopening square brackets ( ] [ ) and add a comma in line 144 to show the list continues with the new cases.

PR #2406, opened 2020-12-11
Update HolidayExtractor.json

XiaoxiaoMa0815:


Specs/DateTime/French/HolidayExtractor.json
```diff
@@ -141,5 +141,15 @@
...
+      "Input": "J'y retournerai à Noël",
+      "Results": [
+        {
+          "Text": "Noël",
+          "Type": "date",
+          "Start": 16,
+          "Length": 9
+       }
+     ]
+   },
```
tellarin, line 155: There should be no "," here and the next line needs to finish the spec with ]. Like in the lines in red above.
Other than this, it seems you got the process.

tellarin: The build log shows where the error is:
WARNING: Invalid array passed in, extra trailing ','. (2906): [

tellarin:
@XiaoxiaoMa0815, when creating the real PR, please provide a description in the field above, ok?
Thanks.

Specs/DateTime/French/HolidayExtractor.json
```diff
@@ -141,5 +141,15 @@
         "Length": 14
       }
     ]
-  }
-]
\ No newline at end of file
+  },
+  {
+      "Input": "J'y retournerai à Noël",
+      "Results": [
```
tellarin, line 147: For all new cases, don't forget to add a new line 
"NotSupported": "dotnet, javascript, python, java",
This line should be below each new Input line.

PR #2407, opened 2020-12-11
[FR Percent] Create specs for PercentModelPercentMode

XiaoxiaoMa0815:


Specs/Number/French/PercentModelPercentMode
```diff
@@ -0,0 +1,157 @@
+[
+  {
+    "Input": "100%",
+    "NotSupported": "javascript, python",
```
tellarin, line 4: Please mark all new specs as
    "NotSupported": "dotnet, javascript, python, java",
 
Each spec case should have an annotation like this. This is to avoid breaking the build with new cases that are not yet supported.

tellarin (changes requested):

tellarin:
@XiaoxiaoMa0815, please name the PR titles following the template for this one that I have edited, ok?
Also, please add to the PR description the number of new (or edited/fixed) French cases in each PR.

Specs/Number/French/PercentModelPercentMode
tellarin: If this build doesn't fail, you can skip adding in this one. But please add them in all future PRs.

tellarin (approved):
Approving as-is as PR build has passed.

Issue #2375, opened 2020-11-17
[ZH DatetimeV2] Behavior change for month and day order in resolution


omarelhariry:
**Describe the bug**
There is a behavior change in the resolution for DateTime entity for chinese. The query "12/1" used to be detected as Dec 1st but now it is detected as Jan 12th.

**To Reproduce**
Create a test file that calls `DateTimeModel.Parse` method with a test case of "12/1".
Validate returned result.
Use 2 different versions, I used versions 1.3.2 and 1.4.0, (one lower and one higher than or equal to 1.4) the lower version detects Dec 1st while the higher version detects Jan 12th. 

**Expected behavior**
Resolution to be Dec 1st. "XXXX-12-01"

**Sample input/output**
Input:
"12/1"

Faulty output:
Jan 12th "XXXX-01-12"

Expected output:
Dec 1st "XXXX-12-01"

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: nuget package/LUIS
 - Version of package v1.3.2 and v1.4.0

**Additional context**

tellarin:
@omarelhariry, this was actually a correction of the designed behaviour. By design the default order if DD/MM, so 12/Jan is the "correct" output for "12/1" as a date. Unfortunately, in pre v1.4.0 there were some inconsistencies were some cases returned 12/Jan and others 1/Dec.
There is a feature in the pipeline to allow consumers to specify which format/order to use in such potentially ambiguous cases, but there's no ETA currently for when it will be out.

PR #2411, opened 2020-12-15
[ZH DateTimeV2] Date regression fix due to side-effect of previous Range change (#2375)

ChopperMan33:
Refer to discussion #2375, recognizing a purely numeric date in Chinese will follow these rules:
1. If the order/format is X/X/Y, treat it as D/M/Y. 
2. If the order/format is Y/X/X, treat it as Y/M/D.
3. If the order/format is X/X, treat it as M/D.

Added test cases.


Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -85,24 +85,23 @@ DateRegexList3: !nestedRegex
...
+  def: '{MonthNumRegex}\s*/\s*{DayRegex}'
+  references: [MonthNumRegex, DayRegex]
 # 23/7
 DateRegexList5: !nestedRegex
-  def: '{DayRegex}\s*/\s*{MonthNumRegex}((\s+|\s*,\s*){SimpleYearRegex})?'
-  references: [DayRegex, MonthNumRegex, SimpleYearRegex]
+  def: '{DayRegex}\s*/\s*{MonthNumRegex}'
+  references: [DayRegex, MonthNumRegex]
 # 3-23-2017
-# The final lookahead in DateExtractor6|7|8 avoids extracting as date "10/1-11" from an input like "10/1-11/2/2017"
```
ChopperMan33, line 95: "10/1-11/2/2017" is not a format used. This code used to implement it has caused cases like "10/1/2018-10/7/2018" to be incorrectly extracted.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -121,6 +121,7 @@ public static class Constants
 
         public const string DefaultLanguageFallback_MDY = "MDY";
```
tellarin, line 122: The comment is missing here.

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
```diff
@@ -55,11 +55,11 @@ public static class DateTimeDefinitions
...
-      public static readonly string DateRegexList4 = $@"{MonthNumRegex}\s*/\s*{DayRegex}((\s+|\s*,\s*){SimpleYearRegex})?";
-      public static readonly string DateRegexList5 = $@"{DayRegex}\s*/\s*{MonthNumRegex}((\s+|\s*,\s*){SimpleYearRegex})?";
-      public static readonly string DateRegexList6 = $@"{MonthNumRegex}\s*[/\\\-]\s*{DayRegex}\s*[/\\\-]\s*{SimpleYearRegex}(?!\s*[/\\\-\.]\s*\d+)";
-      public static readonly string DateRegexList7 = $@"{DayRegex}\s*[/\\\-\.]\s*{MonthNumRegex}\s*[/\\\-\.]\s*{YearNumRegex}(?!\s*[/\\\-\.]\s*\d+)";
-      public static readonly string DateRegexList8 = $@"{YearNumRegex}\s*[/\\\-\. ]\s*{MonthNumRegex}\s*[/\\\-\. ]\s*{DayRegex}(?!\s*[/\\\-\.]\s*\d+)";
+      public static readonly string DateRegexList4 = $@"{MonthNumRegex}\s*/\s*{DayRegex}";
+      public static readonly string DateRegexList5 = $@"{DayRegex}\s*/\s*{MonthNumRegex}";
+      public static readonly string DateRegexList6 = $@"{MonthNumRegex}\s*[/\\\-]\s*{DayRegex}\s*[/\\\-]\s*{SimpleYearRegex}";
+      public static readonly string DateRegexList7 = $@"{DayRegex}\s*[/\\\-\.]\s*{MonthNumRegex}\s*[/\\\-\.]\s*{SimpleYearRegex}";
```
tellarin, line 61: Why this change from YearNumRegex?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDateExtractorConfiguration.cs
```diff
@@ -75,38 +75,25 @@ public class ChineseDateExtractorConfiguration : AbstractYearExtractor, IDateTim
             // (2015年)?(农历)?十月二十(星期三)?
             new Regex(DateTimeDefinitions.DateRegexList3, RegexFlags),
 
-            DateTimeDefinitions.DefaultLanguageFallback == Constants.DefaultLanguageFallback_DMY ?
+            // 7/23
+            new Regex(DateTimeDefinitions.DateRegexList4, RegexFlags),
```
tellarin, line 79: Why is this no longer under the flag check?

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
```diff
@@ -55,11 +55,11 @@ public static class DateTimeDefinitions
       public static readonly string DateRegexList1 = $@"({LunarRegex}(\s*))?((({SimpleYearRegex}|{DateYearInChineseRegex})年)(\s*))?{MonthRegex}(\s*){DateDayRegexInChinese}((\s*|,|，){WeekDayRegex})?";
       public static readonly string DateRegexList2 = $@"((({SimpleYearRegex}|{DateYearInChineseRegex})年)(\s*))?({LunarRegex}(\s*))?{MonthRegex}(\s*){DateDayRegexInChinese}((\s*|,|，){WeekDayRegex})?";
       public static readonly string DateRegexList3 = $@"((({SimpleYearRegex}|{DateYearInChineseRegex})年)(\s*))?({LunarRegex}(\s*))?{MonthRegex}(\s*)({DayRegexNumInChinese}|{DayRegex})((\s*|,|，){WeekDayRegex})?";
-      public static readonly string DateRegexList4 = $@"{MonthNumRegex}\s*/\s*{DayRegex}((\s+|\s*,\s*){SimpleYearRegex})?";
```
tellarin, line 58: Had that commit that caused the side-effect also changed these? I thought it only changed regexes 5, 6, and 7. But maybe I misremember.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDateExtractorConfiguration.cs
```diff
@@ -75,38 +75,25 @@ public class ChineseDateExtractorConfiguration : AbstractYearExtractor, IDateTim
...
-                // 23/7
-                new Regex(DateTimeDefinitions.DateRegexList5, RegexFlags),
+                // 23-3-2017
+                new Regex(DateTimeDefinitions.DateRegexList7, RegexFlags),
 
-            DateTimeDefinitions.DefaultLanguageFallback == Constants.DefaultLanguageFallback_DMY ?
+            DateTimeDefinitions.DefaultLanguageFallback == Constants.DefaultLanguageFallback_MDY ?
 
-                // 23-3-2015
+                // 23-3-2017
```
tellarin, line 91: No need to change the year. :)

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/Constants.java
```diff
@@ -116,6 +116,7 @@
 
     public static final String DefaultLanguageFallback_MDY = "MDY";
```
tellarin, line 117: Same.

JavaScript/packages/recognizers-date-time/src/dateTime/constants.ts
```diff
@@ -35,6 +35,7 @@ export class Constants {
 
     static readonly DefaultLanguageFallback_MDY: string = 'MDY';
```
tellarin, line 36: Same

Patterns/Chinese/Chinese-DateTime.yaml
tellarin: Does it now extract cases like "10/1/2018-10/7/2018"  and "2018/1/10-2018/7/10" correctly?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
ChopperMan33: The YearNumRegex can only support years that contain 4 numbers like 1999 2020. Now SimpleYearRegex can support years that only contain 2 numbers like "3-7-07" -> ”2007-07-03“, which is same use in DateRegexList6.

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
ChopperMan33: Because the order of month and day in X/X(M/D) is different from X/X/Y(D/M/Y), I changed DateRegexList4/5 to only support X/X pattern. Separately extract X/X and X/X/Y pattern.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseDateExtractorConfiguration.cs
ChopperMan33: For now, DataRegexList4 only extract "X/X" pattern,  in Chinese, only use M/D in "X/X" pattern according to the rules.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
ChopperMan33: Sorry for that, I will add it immediately.

Patterns/Chinese/Chinese-DateTime.yaml
ChopperMan33: Yes. I will add them into specs.

Specs/DateTime/Chinese/DateParser.json
```diff
@@ -909,12 +909,12 @@
         "Text": "3-7-27",
```
tellarin, line 909: This one is interesting though. In the two digit year case, would 27-7-3 return 2027-07-03 too? It should as YMD should have precedence over DMY.

tellarin (changes requested):

tellarin (approved):

Issue #2400, opened 2020-12-10
[ES DateTimeV2] Date periods/ranges inconsistently recognized


tellarin:
Examples: 
- "Mostrar ventas toda la semana"
- "Me iré por unas horas"
- "Mostrar ventas en la primera semana"
- "2017-t1 a 2018-t1"

Some issues also occur in composite entities that include a period/range.

Specs already exists and are tagged with PERIODREF.

tellarin:
@LionbridgeCSII, can you take a look at this? If any spec has a behaviour differing from English, please comment here.

PR #2414, opened 2020-12-18
[ES DateTimeV2] Fix date periods/ranges inconsistently recognized (#2400)

LionbridgeCSII:
Fix to issue #2400.

The pattern "primera semana" is not currently supported, also in English "first week" is not recognized (but instead "week 1"/"semana 1" is extracted). Should I add support for the pattern?

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
@@ -1895,7 +1895,7 @@ namespace Microsoft.Recognizers.Text.DateTime
                 var swift = this.config.GetSwiftYear(orderStr);
                 if (swift < -1)
                 {
-                    return ret;
+                    swift = 0;
                 }
 
                 year = referenceDate.Year + swift;
--- a/Specs/DateTime/Spanish/DateTimeModel.json
+++ b/Specs/DateTime/Spanish/DateTimeModel.json
@@ -2751,8 +2751,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "Comment": "PERIODREF. Refine support for quarter, half, week, weekend, etc. Including relatives.",
-    "NotSupported": "dotnet, javascript, python, java",
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "un mes entero",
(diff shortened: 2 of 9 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -1895,7 +1895,7 @@ private DateTimeResolutionResult ParseHalfYear(string text, DateObject reference
                 var swift = this.config.GetSwiftYear(orderStr);
                 if (swift < -1)
                 {
-                    return ret;
+                    swift = 0;
```
tellarin, line 1898: Why is this necessary?

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -6714,52 +6702,18 @@
...
-              "timex": "2019-W01",
-              "type": "daterange",
-              "start": "2018-12-31",
-              "end": "2019-01-07"
-            }
-          ]
-        }
-      }
-    ]
+    "Comment": "Also in English, 'first week' is not extracted ('4th week of month/year' and 'week 3' are the supported patterns).",
```
tellarin, line 6705: That makes sense, but wouldn't it be better for "semana" to be extracted by itself in "primera semana" as a duration?
Same in English.

tellarin (approved):
Approving as-is. In case we decide to implement the extraction of "semana" from input like "1a semana", that can come in a follow up PR.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
LionbridgeCSII: otherwise the extraction is parsed only if it contains a year, but this is not always the case e.g. "2º semestre traerá desafíos" (the same behaviour was already implemented in ParseQuarter).

Specs/DateTime/Spanish/DateTimeModel.json
LionbridgeCSII: do you mean to extract a unit like "week" as duration in case it is preceded by an ordinal number?

tellarin: Yes. We can create an issue for it. It's lower priorirty than some of the other fixes.

tellarin: Do you see any potential problem with that extraction?

LionbridgeCSII: ok, I will open the issue. No I do not see any problem at the moment.

Issue #2401, opened 2020-12-10
[ES DateTimeV2] Handling of "in" vs "within" seems inconsistent with English implementation


tellarin:
There seem to be discrepancies in how Spanish is handling cases like "dentro de 5 minutos" vs "en 5 mins".

Specs exist and are tagged as INWITHIN.

tellarin:
@LionbridgeCSII

PR #2419, opened 2020-12-24
[ES DateTimeV2] Improved handling of "in" vs "within" (#2401)

LionbridgeCSII:
Conformed the behaviour of Spanish to English in handling "within" (dentro de) vs "in" (en) (#2401).

Updated a few cases in DatePeriod and DateTimePeriod to conform to the new convention:
- "Estare afuera en 3 años" -> "Estare afuera próximos 3 años"
- "Ya no estaré en dos semanas" -> "Ya no estaré la semana después de la próxima"
- "Ocurrirá en 4 semanas" -> "Ocurrirá 4 semanas en el futuro"
- "Voy a volver en 3 horas" -> "Voy a volver las próximas 3 horas"
- "Voy a volver en 5 h" -> "Voy a volver próximas 5 h"

The problem is "en" is ambiguous and can mean both "in" and "within", but since it is not possible to disambiguate it, it seems better to always treat it as "in" instead of having it parsed as "in" in certain cases and as "within" in others.
The original cases have been kept and added to DateTimeModel.


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19112,5 +19112,31 @@
       "ReferenceDateTime": "2011-07-02T00:00:00"
     },
     "Results": []
+  },
+  {
+    "Input": "I'll be back within 5 minutes",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "Comment": "INWITHIN. Refine support for _in_ vs. _within_ for consistency with English.",
```
tellarin, line 19121: This comment shouldn't be here, right?

Specs/DateTime/Spanish/DateTimePeriodParser.json
```diff
@@ -2891,5 +2856,55 @@
         "Length": 21
       }
     ]
+  },
```
tellarin, line 2859: It seems the source of most issues was a translation confusion between "in" vs "within".
In English the current behaviour is:
"in 3 min" -> "3 min from now" -> datetime
"within 3 min" -> "from now until 3 min later" -> range
Which it seems you managed to correct or adapt the mistranslated specs into the right output.
Could you add a couple more spec cases to the appropriate Parsers and DateTimeModel spec files that closely match the 4 cases above? Some may already exist.

tellarin (changes requested):
Thanks for the changes. This looks quite good.
I've just left a comment to add a couple more spec cases that closely match 4 cases that show the current interpretation distinction between "in"  and "within".

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: yes, removed.

Specs/DateTime/Spanish/DateTimePeriodParser.json
LionbridgeCSII: added cases to Parsers and Model. 
The case "from now until 3 min later" maybe requires additional fixes because the parsed Timex is "PT0H" instead of "PT3M"

tellarin: Got it. Please add it as the next in the task stack.

LionbridgeCSII: fixed

tellarin:
@LionbridgeCSII Sorry, this one fell through the cracks. Can you solve the conflicts here and check if PT needs the same fixes?
This can come after the issues you're currently working on.
Thanks.

LionbridgeCSII:
@tellarin, conflicts solved. 
Added WithinNextRegex to Portuguese yaml file and updated a few test cases to follow the same logic ("em" -> "in", "dentro de" -> within).

tellarin:
@LionbridgeCSII can you re-resolve conflicts? Now we can merge to prepare for the next release. Thanks.

LionbridgeCSII:
@tellarin, conflicts solved.

tellarin (approved):

PR #2421, opened 2020-12-27
[ZH DateTime V2] Fixed the output of invalid day like "2/29/2019" and "2/30"

ChopperMan33:
1. Modify the output of Feb 29th according to the reference date. 
2. Filter invalid futureresolution or pastresolution.
3. If they are both invalid, output "not resolved".
4. Added test cases.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
@@ -254,7 +254,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
                             {
                                 pastDate = pastDate.AddMonths(-1);
                             }
-                            else if (IsNonleapYearFeb29th(year, month - 1, day))
+                            else if (!DateObject.IsLeapYear(year) && IsFeb29th(year, month - 1, day))
                             {
                                 pastDate = pastDate.AddMonths(-2);
                             }
@@ -550,32 +550,49 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
                 ret.Timex = DateTimeFormatUtil.LuisDate(year, month, day);
             }
 
-            DateObject futureDate, pastDate;
-            if (IsNonleapYearFeb29th(year, month, day))
+            var futureDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
+            var pastDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
+            var futureYear = year;
+            var pastYear = year;
+            if (noYear)
             {
-                futureDate = DateObject.MinValue.SafeCreateFromValue(year + 1, month, day);
-                pastDate = DateObject.MinValue.SafeCreateFromValue(year - 1, month, day);
-                if (!futureDate.IsDefaultValue() && futureDate.AddYears(-1) >= referenceDate)
+                if (IsFeb29th(year, month, day))
                 {
-                    futureDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-                }
-                else if (!pastDate.IsDefaultValue() && pastDate.AddYears(+1) < referenceDate)
-                {
-                    pastDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-                }
-            }
-            else
-            {
-                futureDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-                pastDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-                if (noYear && futureDate < referenceDate && !futureDate.IsDefaultValue())
-                {
-                    futureDate = DateObject.MinValue.SafeCreateFromValue(year + 1, month, day);
-                }
+                    if (DateObject.IsLeapYear(year))
+                    {
+                        if (futureDate < referenceDate)
+                        {
+                            futureDate = DateObject.MinValue.SafeCreateFromValue(futureYear + 4, month, day);
+                        }
+                        else
+                        {
+                            pastDate = DateObject.MinValue.SafeCreateFromValue(pastYear - 4, month, day);
+                        }
+                    }
+                    else
+                    {
+                        while (futureDate.IsDefaultValue() && futureYear - year <= 4)
+                        {
+                            futureDate = DateObject.MinValue.SafeCreateFromValue(++futureYear, month, day);
+                        }
 
-                if (noYear && pastDate >= referenceDate && !pastDate.IsDefaultValue())
+                        while (pastDate.IsDefaultValue() && year - pastYear <= 4)
+                        {
+                            pastDate = DateObject.MinValue.SafeCreateFromValue(--pastYear, month, day);
+                        }
+                    }
+                }
+                else
                 {
-                    pastDate = DateObject.MinValue.SafeCreateFromValue(year - 1, month, day);
+                    if (futureDate < referenceDate && !futureDate.IsDefaultValue())
+                    {
+                        futureDate = DateObject.MinValue.SafeCreateFromValue(year + 1, month, day);
+                    }
+
+                    if (pastDate >= referenceDate && !pastDate.IsDefaultValue())
+                    {
+                        pastDate = DateObject.MinValue.SafeCreateFromValue(year - 1, month, day);
+                    }
                 }
             }
 
@@ -642,10 +659,10 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
             return DateObjectExtension.IsValidDate(year, month, day);
         }
 
-        // Judge the date is non-leap year Feb 29th
-        private static bool IsNonleapYearFeb29th(int year, int month, int day)
+        // Judge the date is Feb 29th
+        private static bool IsFeb29th(int year, int month, int day)
         {
-            return !DateObject.IsLeapYear(year) && month == 2 && day == 29;
+            return month == 2 && day == 29;
         }
 
         // Handle cases like "三天前"
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
@@ -550,10 +550,10 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
                 ret.Timex = DateTimeFormatUtil.LuisDate(year, month, day);
             }
 
-            var futurePastDateList = DateContext.GetFuturePastDate(noYear, referenceDate, year, month, day);
+            var futurePastDates = DateContext.GenerateDates(noYear, referenceDate, year, month, day);
 
-            ret.FutureValue = futurePastDateList[0];
-            ret.PastValue = futurePastDateList[1];
+            ret.FutureValue = futurePastDates.future;
+            ret.PastValue = futurePastDates.past;
             ret.Success = true;
 
             return ret;
@@ -615,7 +615,6 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
             return DateObjectExtension.IsValidDate(year, month, day);
         }
 
-        // Judge the date is Feb 29th
         private static bool IsFeb29th(int year, int month, int day)
         {
             return month == 2 && day == 29;
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
@@ -550,16 +550,33 @@ namespace Microsoft.Recognizers.Text.DateTime.Chinese
                 ret.Timex = DateTimeFormatUtil.LuisDate(year, month, day);
             }
 
-            var futureDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-            var pastDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-            if (noYear && futureDate < referenceDate)
+            DateObject futureDate, pastDate;
+            if (IsNonleapYearFeb29th(year, month, day))
             {
-                futureDate = futureDate.AddYears(+1);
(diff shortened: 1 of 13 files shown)


Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -6283,5 +6283,152 @@
...
+        "Text": "2月29日-3月1日",
+        "Start": 0,
+        "End": 9,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(XXXX-02-29,XXXX-03-01,P1D)",
+              "type": "daterange",
+              "value": "not resolved"
```
tellarin, line 6427: This is also incorrect. Such range should only be unresolved if the input specifies a specific year that is not a leap year.
For example, if the input was: 2019/02/29 - 2010/03/01
Also, can you add examples in DateTimeModel,json that have the fully specified date too?

tellarin: It would be good to add specs for English too, as we need to make sure this behaviour is correct across all languages.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
```diff
@@ -550,16 +550,33 @@ protected DateTimeResolutionResult Match2Date(Match match, DateObject referenceD
...
             }
 
-            var futureDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-            var pastDate = DateObject.MinValue.SafeCreateFromValue(year, month, day);
-            if (noYear && futureDate < referenceDate)
+            DateObject futureDate, pastDate;
+            if (IsNonleapYearFeb29th(year, month, day))
             {
-                futureDate = futureDate.AddYears(+1);
+                futureDate = DateObject.MinValue.SafeCreateFromValue(year + 1, month, day);
```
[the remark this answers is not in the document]
ChopperMan33: Yes. SafeCreateFromValue returns MinValue 0001-01-01 for invalid date. Then AddSingleDateTimeToResolution filters MinValues so that this resolution will be ignored.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/FullDateTimeParser.cs
```diff
@@ -10,6 +10,8 @@ public class FullDateTimeParser : IDateTimeParser
     {
         public const string ParserTypeName = "datetimeV2";
 
+        public static readonly string DateMinString = DateTimeFormatUtil.FormatDate(DateObject.MinValue);
```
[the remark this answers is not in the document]
ChopperMan33: OK. I will modify it.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/FullDateTimeParser.cs
```diff
@@ -20,7 +22,8 @@ public FullDateTimeParser(IFullDateTimeParserConfiguration configuration)
         public static void AddSingleDateTimeToResolution(Dictionary<string, string> resolutionDic, string type,
                                                          string mod, Dictionary<string, string> res)
         {
-            if (resolutionDic.ContainsKey(type))
+            if (resolutionDic.ContainsKey(type) &&
+                !resolutionDic[type].StartsWith(DateMinString, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
ChopperMan33: It is for filtering invalid date which returns DateMinString. I see same logic in English code.

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -6283,5 +6283,152 @@
...
+      "ReferenceDateTime": "2019-03-22T00:00:00"
+    },
+    "Results": [
+      {
+        "Text": "2月29日",
+        "TypeName": "datetimeV2.date",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "XXXX-02-29",
```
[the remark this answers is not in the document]
ChopperMan33: The code only checks 1y ahead and back now. So should it return 2016-2-29 and 2020-2-29 in this case?

Specs/DateTime/Chinese/DateTimeModel.json
ChopperMan33: OK. Got it. I can add it in other languages after I fix the Chinese one.

Specs/DateTime/Chinese/DateTimeModel.json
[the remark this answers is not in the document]
tellarin: I believe so. But returning only the closest is OK if returning both causes some issue or affects perf too much.
But definitely the code shouldn't check just one year in such cases.

Specs/DateTime/Chinese/DateParser.json
```diff
@@ -1289,5 +1289,97 @@
...
+      {
+        "Text": "2月29日",
+        "Type": "date",
+        "Value": {
+          "Timex": "XXXX-02-29",
+          "FutureResolution": {
+            "date": "2020-02-29"
+          },
+          "PastResolution": {
+            "date": "0001-01-01"
```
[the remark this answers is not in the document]
ChopperMan33: The Parser returns DateTimeResolutionResult, which contains FutureResolution and PastResolution two parameters. And we can't remove it in the Parser level.
So the parser result returns the default DateTime value "0001-01-01" for the invalid date, which will be removed in the Model level.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/DateContext.cs
```diff
@@ -20,6 +78,38 @@ public static DateObject SwiftDateObject(DateObject beginDate, DateObject endDat
             return beginDate;
         }
 
+        public (DateTimeParseResult pr1, DateTimeParseResult pr2) SyncDateEntityInFeb29th(DateTimeParseResult pr1, DateTimeParseResult pr2)
```
ChopperMan33, line 81: For case: 2.29-3.1 reference: 2019-03-22 
The pastresolution was (2016-2-29, 2019-3-1).
We can synchronize the year with this method.

Specs/DateTime/Chinese/DateTimeModel.json
```diff
@@ -6283,5 +6283,272 @@
...
+    "Results": [
+      {
+        "Text": "2019年2月29日-3月1日",
+        "Start": 0,
+        "End": 14,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(2019-02-29,2019-03-01,P737118D)",
```
ChopperMan33, line 6545: The invalid date 2019-02-29 will be parsed to 0001-01-01, so the period is "P737118D".

tellarin: There should be special logic here. Period can either be P1D or PXD.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/DateContext.cs
tellarin: I'm not sure what you mean by this. It seems the past resolution was correctly calculated?

tellarin: I see now. Your resolution to find a leap year for such cases is ok.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -964,27 +975,15 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
...
-                };
-                ret.PastResolution = new Dictionary<string, string>
-                {
-                    {
-                        DateTimeResolutionKey.Timex,
-                        $"({pr1.TimexStr},{pr2.TimexStr},P{(pastEnd - pastBegin).TotalDays}D)"
-                    },
-                };
+                ret.Comment = Constants.Comment_DoubleTimex;
+                ret.Timex += "|" + TimexUtility.GenerateDatePeriodTimex(pastBegin, pastEnd, DatePeriodTimexType.ByDay, pr1.TimexStr, pr2.TimexStr);
```
tellarin, line 984: It's better to be explicit on what's happening here. Create a new method called MergeTimeAlternatives in TimexUtility and pass both timexes to it.
And if you want to use a separator like '|' declare it as a constant.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -59,7 +59,7 @@ public static class Constants
         public const string Comment_WeekOf = "WeekOf";
         public const string Comment_MonthOf = "MonthOf";
 
-        public const string Comment_ContainFeb29th = "ContainFeb29th";
+        public const string Comment_DoubleTimex = "DoubleTimex";
```
tellarin, line 62: Please add a code comment here, like:
// Tag to mark cases where the specifc resolution timex depends on future or past values.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/FullDateTimeParser.cs
```diff
@@ -417,10 +417,9 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject referenceTime)
                 }
             }
 
-            if (!string.IsNullOrEmpty(comment) && comment.Equals(Constants.Comment_ContainFeb29th, StringComparison.Ordinal))
+            if (!string.IsNullOrEmpty(comment) && comment.Equals(Constants.Comment_DoubleTimex, StringComparison.Ordinal))
             {
-                ResolveFeb29th(res, Constants.ResolveToFuture, futureResolutionStr);
-                ResolveFeb29th(res, Constants.ResolveToPast, pastResolutionStr);
+                UpdateTimex(res, Constants.ResolveToFuture, Constants.ResolveToPast, timex);
```
tellarin, line 422: Maybe ProcessDoubleTimex or ProcessCompositeTimex?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/FullDateTimeParser.cs
```diff
@@ -417,10 +417,9 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject referenceTime)
                 }
             }
 
-            if (!string.IsNullOrEmpty(comment) && comment.Equals(Constants.Comment_ContainFeb29th, StringComparison.Ordinal))
+            if (!string.IsNullOrEmpty(comment) && comment.Equals(Constants.Comment_DoubleTimex, StringComparison.Ordinal))
```
tellarin, line 420: Minor comment. But 
`comment.Equals(Constants.Comment_DoubleTimex, StringComparison.Ordinal)`
could be a method HasDoubleTimex or HasCompositeTimex. Just to improve code readability.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/FullDateTimeParser.cs
```diff
@@ -489,15 +488,19 @@ public List<DateTimeParseResult> FilterResults(string query, List<DateTimeParseR
             return candidateResults;
         }
 
-        internal static void ResolveFeb29th(Dictionary<string, object> resolutionDic, string keyName, Dictionary<string, string> resolutionStr)
+        internal static void UpdateTimex(Dictionary<string, object> resolutionDic, string futureKey, string pastKey, string originTimex)
         {
-            if (!resolutionDic.ContainsKey(keyName) || !resolutionStr.ContainsKey(DateTimeResolutionKey.Timex))
+            string[] timexs = originTimex.Split('|');
```
tellarin, line 493: timexes :)
And separators must always be declared somewhere, and not used hardcoded in the code.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDatePeriodParserConfiguration.cs
```diff
@@ -979,9 +979,11 @@ private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject refe
 
             if (pr1.TimexStr.StartsWith(Constants.TimexFuzzyYear) && futureBegin.CompareTo(DateObject.MinValue.SafeCreateFromValue(futureBegin.Year, 2, 28)) <= 0 && futureEnd.CompareTo(DateObject.MinValue.SafeCreateFromValue(futureBegin.Year, 3, 1)) >= 0)
             {
-                // due to only solve the noYear cases, we only have to judge the period in one year contains Feb 29th;
+                // Handle cases like "2月28日到3月1日".
+                // There may be different timexes for FutureValue and PastValue due to the different validity of Feb 29th.
                 ret.Comment = Constants.Comment_DoubleTimex;
-                ret.Timex += "|" + TimexUtility.GenerateDatePeriodTimex(pastBegin, pastEnd, DatePeriodTimexType.ByDay, pr1.TimexStr, pr2.TimexStr);
+                var pastTimex = TimexUtility.GenerateDatePeriodTimex(pastBegin, pastEnd, DatePeriodTimexType.ByDay, pr1.TimexStr, pr2.TimexStr);
+                ret.Timex = TimexUtility.MergeTimeAlternatives(ret.Timex, pastTimex);
```
tellarin, line 986: There's a small typo here. Should be MergeTime**x**Alternatives.

.NET/Microsoft.Recognizers.Text.DateTime/Constants.cs
```diff
@@ -191,6 +192,8 @@ public static class Constants
 
         public const string InvalidDateString = "0001-01-01";
 
+        public const char TimexDelimiter = '|';
```
tellarin, line 195: Rename to CompositeTimexDelimiter as its not a regular timex delimiter, but only a delimiter when there are multiple timexes in the attribute.

tellarin (changes requested):
Minor changes left.

tellarin (approved):

PR #2422, opened 2020-12-29
Add new French specs for DateExtractor

XiaoxiaoMa0815:
143 new French specs go from line 981 to 1053

### DIFF
--- a/Specs/DateTime/French/DateExtractor.json
+++ b/Specs/DateTime/French/DateExtractor.json
@@ -978,3 +978,6751 @@
     ]
   }
 ]
+[
+  {
+    "Input": "je vais rentrer le 15",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "15.0",
+        "Type": "date",
+        "Start": 16,
+        "Length": 2
+      }
+    ]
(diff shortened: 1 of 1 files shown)


tellarin:
We'll review soon. But you can add at least both Extractor and Parser json specs in the same PR. They are likely to be very related.
The goal is not to make individual PRs. But group specs in groups that make sense together, without being too big. Maybe max 6 specs in a PR, if related (but this is flexible).
For example, Date Extractor/Parser and Holiday Extractor/Parser could be one PR as they are the same type.

tellarin:
@XiaoxiaoMa0815, the build is failing.

SpecInfo start...
Error decoding spec file:	C:\projects\recognizers-text\Specs\DateTime\French\DateExtractor.json
WARNING: Invalid JSON primitive: 
# Failed, including empty or duplicate inputs in Specs
Command exited with code 2

tellarin (changes requested):
Build is failing.

tellarin (changes requested):
I'll close this PR as the build is breaking due to invalid spec format and it seems PR #2424 is meant to replace it.
Please comment on this PR in case you want to re-open it.

PR #2424, opened 2020-12-30
[FR DateTimeV2] Add new French specs for DateExtractor/Parser

wumingxuanceline:
143 new specs for DateExtractor
127 new specs for DateParser

### DIFF
--- a/Specs/DateTime/French/DateExtractor.json
+++ b/Specs/DateTime/French/DateExtractor.json
@@ -976,5 +976,1690 @@
         "Length": 9
       }
     ]
+  },
+  {
+    "Input": "je vais rentrer le 15",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "15",
+        "Type": "date",
+        "Start": 16,
+        "Length": 2
+      }
+    ]
+  },
+  {
+    "Input": "je vais rentrer le 22 avril",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "22 avril",
+        "Type": "date",
+        "Start": 13,
+        "Length": 8
+      }
+    ]
+  },
+  {
+    "Input": "je vais rentrer le 1-jan.",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "1-jan",
+        "Type": "date",
+        "Start": 13,
+        "Length": 5
+      }
+    ]
+  },
+  {
+    "Input": "je vais rentrer le 01 jan.",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "01 jan",
+        "Type": "date",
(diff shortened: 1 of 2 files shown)


wumingxuanceline:
Hi Börje, we added news specs for these two files and had already checked them for many times, but one of them still failed because of "empty or duplicate inputs in Specs". Could you please have a check and point out the problem ? Thanks.

ChopperMan33:
Hi @wumingxuanceline ,
I find you add an extra characters `[`  in row 1451 https://github.com/microsoft/Recognizers-Text/commit/04a311aecd1daedda38ed68eb32ea9b721018570#diff-d3279030e99103578e94608f49dd2edabfa1343fb0323424d7ffbada14b79680R1451, which will make the json file duplicate.

wumingxuanceline:
> Hi @wumingxuanceline ,
> I find you add an extra character `[` in row 1451 [04a311a#diff-d3279030e99103578e94608f49dd2edabfa1343fb0323424d7ffbada14b79680R1451](https://github.com/microsoft/Recognizers-Text/commit/04a311aecd1daedda38ed68eb32ea9b721018570#diff-d3279030e99103578e94608f49dd2edabfa1343fb0323424d7ffbada14b79680R1451), which will make the json file duplicate.

ChopperMan33:
Hi @wumingxuanceline ,
You've already fixed the brackets problem. But there are 18 duplicate test cases like "je vais rentrer vendredi", 
"je vais rentrer le 22/04" and so on.
They caused the error. You can find them in https://ci.appveyor.com/project/tellarin/recognizers-text/builds/37054975.

wumingxuanceline:
Hi @ChopperMan33 ,
Thanks for your help. But these cases are not actually duplicate. In fact, they are different inputs but have the same french version, though we've tried to vary the expressions. Therefore, the rest files will also have the same problem. Is there any solution for the work ?

tellarin (changes requested):
Build is failing due to duplicate spec cases.

tellarin:
@wumingxuanceline, this means that they are now duplicate in French. 
There's no need to have more than one case with the exact same input sentence and reference parameters. If there's no other applicable variation for these cases (preserving the original case meaning in English) please remove the duplicates.
When removing duplicates, please update the PR description to mention what the original cases were in English and to which case they map in French.

Something like:
"I'll go back on Friday", and "I'll go back Friday" map to “je vais rentrer vendredi” in French.

wumingxuanceline:
> @wumingxuanceline, this means that they are now duplicate in French.
> There's no need to have more than one case with the exact same input sentence and reference parameters. If there's no other applicable variation for these cases (preserving the original case meaning in English) please remove the duplicates.
> When removing duplicates, please update the PR description to mention what the original cases were in English and to which case they map in French.
> 
> Something like:
> "I'll go back on Friday", and "I'll go back Friday" map to “je vais rentrer vendredi” in French.

wumingxuanceline:
Will start a new PR

Issue #2417, opened 2020-12-22
[* Duration] Extract duration units when preceded by ordinal numbers


LionbridgeCSII:
Duration units like "week", "month", "year"... should be recognized as duration entities when preceded by ordinal numbers.
For example: 
"the first week is dedicated to exercises"  --> "week" (duration)

tellarin:
@LionbridgeCSII, can you take a stab at this and see if there may be any side-effects?

LionbridgeCSII:
@tellarin, yes there are some interferences with DatePeriod and now some patterns are not correctly handled in certain languages. I am working on fixing these issues.

tellarin:
Please raise a flag if this becomes too unwieldy.

PR #2425, opened 2020-12-30
[* Duration] Extract duration units when preceded by ordinal numbers (#2417)

LionbridgeCSII:
Added support for duration units extraction when preceded by ordinal numbers (#2417).

Test cases added to EN and ES DateTimeModels.

Method not implemented in Turkish and Hindi (in both languages the ordinal "first" can be also interpreted as "before/previous" and this causes issues with certain patterns).


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -267,7 +267,7 @@ private bool ShouldSkipFromToMerge(ExtractResult er)
 
         private List<ExtractResult> FilterUnspecificDatePeriod(List<ExtractResult> ers)
         {
-            ers.RemoveAll(o => this.config.UnspecificDatePeriodRegex.IsMatch(o.Text));
+            ers.RemoveAll(o => o.Type == "daterange" && this.config.UnspecificDatePeriodRegex.IsMatch(o.Text));
```
tellarin, line 270: Can you explain the reason for this? What type of case falls here?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -223,7 +223,7 @@ private void AddTo(List<ExtractResult> dst, List<ExtractResult> src, string text
                 var firstIndex = -1;
                 for (var i = 0; i < dst.Count; i++)
                 {
-                    if (dst[i].IsOverlap(result))
+                    if (dst[i].IsOverlap(result) && !(dst[i].Text == result.Text && this.config.UnspecificDatePeriodRegex.IsMatch(result.Text)))
```
tellarin, line 226: Can you add a comment explaining this condition?

.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
```diff
@@ -60,12 +60,15 @@ public HindiDurationExtractorConfiguration(IDateTimeOptionsConfiguration config)
             : base(config)
         {
             CardinalExtractor = Number.Hindi.CardinalExtractor.GetInstance();
+            OrdinalExtractor = null;
```
tellarin, line 63: Why is this one null?

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDurationExtractorConfiguration.cs
```diff
@@ -60,12 +60,15 @@ public TurkishDurationExtractorConfiguration(IDateTimeOptionsConfiguration confi
             : base(config)
         {
             CardinalExtractor = Number.Turkish.CardinalExtractor.GetInstance();
+            OrdinalExtractor = null;
```
tellarin, line 63: Same question as in Hindi.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19104,13 +19104,86 @@
     "Context": {
       "ReferenceDateTime": "2011-07-02T00:00:00"
     },
-    "Results": []
+    "NotSupported": "java, javascript, python",
+    "Results": [
+      {
+        "Text": "week",
```
tellarin, line 19110: This should either return "first week" and be un-resolved, or it should still return [].

Specs/DateTime/Spanish/DatePeriodExtractor.json
```diff
@@ -3476,10 +3476,10 @@
     "NotSupportedByDesign": "javascript",
     "Results": [
       {
-        "Text": "año 247",
+        "Text": "el año 247",
```
tellarin, line 3479: Prefereably we shouldn't tag the article, if possible. It could be reasonable to keep it, if altering this behaviour would be too problematic.

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -6702,18 +6702,48 @@
     "Context": {
       "ReferenceDateTime": "2019-03-02T00:00:00"
     },
-    "Comment": "Also in English, 'first week' is not extracted ('4th week of month/year' and 'week 3' are the supported patterns).",
     "NotSupported": "javascript, python, java",
-    "Results": []
+    "Results": [
+      {
+        "Text": "semana",
```
tellarin, line 6708: Same comment as in EN.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
LionbridgeCSII: We want to filter unspecific date periods like "week", "month" only when they are extracted by DatePeriodExtractor, but not when they are extracted as Duration (otherwise many cases addressed in this PR would also be filtered since we are extracting units when preceded by ordinals, e.g. "first week" -> "week").

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
LionbridgeCSII: the condition avoids merging results that match UnspecificDatePeriodRegex. For example in "first week", "week" is extracted by DatePeriod and by Duration and we want to keep both results separate (the DatePeriod result will be later filtered). 
Comment added.

.NET/Microsoft.Recognizers.Text.DateTime/Hindi/Extractors/HindiDurationExtractorConfiguration.cs
LionbridgeCSII: In this way the pattern is not implemented in Hindi (and Turkish). As commented in the PR description, there are issues in these languages given by the fact that "first" can be interpreted both as ordinal and as "before/previous".

.NET/Microsoft.Recognizers.Text.DateTime/Turkish/Extractors/TurkishDurationExtractorConfiguration.cs
LionbridgeCSII: same as above

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: this is not clear: as for the issue description, I thought that only the unit "week" had to be extracted from "first week". Should we instead make it extract also the ordinal? (Returning [] is the previous behavior, so the second solution seems to imply to discard the whole PR).

Specs/DateTime/Spanish/DatePeriodExtractor.json
LionbridgeCSII: Ok removed. 
But then a problem occurs (for both English and Spanish) when the year is written in letters e.g. "in the year two thousand twelve". In such a case, two separate results are produced "the year" (matched by OneWordPeriodRegex) and "two thousand twelve"

PR #2432, opened 2021-01-07
[Tools] Spec validation tool v2

ChopperMan33:
1. Modify the error message.
2. Refactor runpipeline to separate validation and visualization.
3. Add the index automatic correction function, save the correct result to a new file.

### DIFF
--- a/Tools/src/ValidationTool/ValidationTool.cs
+++ b/Tools/src/ValidationTool/ValidationTool.cs
@@ -1,12 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.CommandLine;
+using System.CommandLine.Invocation;
 using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
 using System.Text.Json;
 using Microsoft.Recognizers.Text.DataDrivenTests;
-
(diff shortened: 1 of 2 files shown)


Tools/src/ValidationTool/ValidationTool.cs
```diff
@@ -100,125 +113,227 @@ private static string[] ListSpecFiles(string specsPath)
...
+            List<dynamic> originalSpecs = (List<dynamic>)JsonConvert.DeserializeObject<IList<dynamic>>(rawData);
+
+            // Replace matching results.
+            for (int i = 0; i < jsonSpecs.Count; ++i)
+            {
+                originalSpecs[i]["Results"] = jsonSpecs[i]["Results"];
+            }
+
+            string rawResults = JsonConvert.SerializeObject(originalSpecs, Formatting.Indented);
+            File.WriteAllText(path, rawResults);
```
tellarin, line 139: We should specify UTF-8 as the encoding here. The project assumes all specs are utf-8.

tellarin (changes requested):

tellarin (approved):

PR #2437, opened 2021-01-18
[.NET DateTime V2] Fixed the output of invalid day like "2/29/2019" and "2/30" in all languages

ChopperMan33:
In EN/FR/DE/HI/IT/JA/PT/ES/TR:
Modify the output of Feb 29th according to the reference date.
Support to recognize date period contains invalid date like 2/29 - 3/1.    
Generate future/past timexes separately if the case has no specific year. 
Added test cases.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
@@ -260,7 +260,7 @@ namespace Microsoft.Recognizers.Text.DateTime
 
         public static string MergeTimexAlternatives(string timex1, string timex2)
         {
-            if (timex1.Equals(timex2))
+            if (timex1.Equals(timex2, StringComparison.Ordinal))
             {
                 return timex1;
             }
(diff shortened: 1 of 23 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -260,9 +260,34 @@ public static DateObject OffsetDateObject(DateObject date, int offset, DatePerio
 
         public static string MergeTimexAlternatives(string timex1, string timex2)
         {
+            if (timex1.Equals(timex2))
```
tellarin, line 263: In such comparisons, always specify StringComparison.ordinal as it's not a language-dependant comparison.

tellarin (changes requested):

Specs/DateTime/Spanish/DateTimeModel.json
```diff
@@ -613,7 +613,7 @@
         "Resolution": {
           "values": [
             {
-              "timex": "(XXXX-02-02,XXXX-03-05,P32D)",
+              "timex": "(XXXX-02-02,XXXX-03-05,P31D)",
```
tellarin, line 616: Why is this now 31? 2020 is a leap year, so it should be 32, no?

tellarin (changes requested):

Specs/DateTime/Spanish/DateTimeModel.json
tellarin: Sorry. Didn't notice this is the timex for the unanchored past resolution.

tellarin (approved):

Issue #2423, opened 2020-12-29
[* DateTimeV2] Extend support for unambiguous date formats


tellarin:
Not all formats for unambiguous dates are currently supported (4-digit year, month as written or abbreviation).

Examples:
Input:"Sep-23-2020" 
Input:"September-2020-23" 
Input:"2020/23/Sep"

These should be supported in all languages that use month names.

tellarin:
@LionbridgeCSII, can you take a look at this one? Only for fully unambiguous complete dates.

tellarin:
@LionbridgeCSII, do you have an ETA on this one? Thanks.

tellarin:
@LionbridgeCSII?

LionbridgeCSII:
@tellarin, I have been told to stop working until the contract for the new year is finalized. 
But I still have a few hours left from last year, so if I manage to finish this using that time, I will push it (maybe later today or tomorrow).

PR #2438, opened 2021-01-19
[* DateTimeV2] Extended support for unambiguous date formats (#2423)

LionbridgeCSII:
Added support in various languages for dates formed by all possible permutations of DayNumber, MonthName, FourDigitYear (#2423).


.NET/Microsoft.Recognizers.Definitions.Common/Dutch/DateTimeDefinitions.cs
```diff
@@ -128,6 +128,7 @@ public static class DateTimeDefinitions
       public const string DatePreposition = @"\b(op(\s+de)?)";
       public static readonly string DateExtractorYearTermRegex = $@"(\s+|\s*,\s*){DateYearRegex}";
       public static readonly string DateExtractor1 = $@"\b({WeekDayRegex}\s*[,-]?\s*)?(({MonthRegex}(\.)?\s*[/\\.,-]?\s*{DayRegex})|(\({MonthRegex}\s*[-.]\s*{DayRegex}\)))(\s*\(\s*{WeekDayRegex}\s*\))?({DateExtractorYearTermRegex}\b)?";
+      public static readonly string DateExtractor2 = $@"\b({WeekDayRegex}\s*[,-]?\s*)?(?<!\d[.,])(({MonthRegex}[\.]?\s*[/\\.,-]?)\s*({DayRegex}\s*[/\\.,-]?\s*{BaseDateTime.FourDigitYearRegex}|{BaseDateTime.FourDigitYearRegex}\s*[/\\.,-]?\s*{DayRegex})|({DayRegex}\s*[/\\.,-]?)\s*({MonthRegex}[\.]?\s*[/\\.,-]?\s*{BaseDateTime.FourDigitYearRegex}|{BaseDateTime.FourDigitYearRegex}\s*[/\\.,-]?\s*{MonthRegex}[\.]?)|({BaseDateTime.FourDigitYearRegex}\s*[/\\.,-]?)\s*({DayRegex}\s*[/\\.,-]?\s*{MonthRegex}[\.]?|{MonthRegex}[\.]?\s*[/\\.,-]?\s*{DayRegex}))\b";
```
tellarin, line 131: Do we really need to add a new regex for all extra cases? I'm a bit worried about performance impact as it's one more regex to execute and it looks long and complex.
Could you run some experiments with different input (repeated dates will hit the cache) and try to measure the impact? Running English only is ok.

tellarin: For ease of reading and  maintenance, I'd also prefer if this new regex has a consistent name across cultures and if it's name is more descriptive.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/Dutch/DateTimeDefinitions.cs
LionbridgeCSII: > Do we really need to add a new regex for all extra cases?

Not necessarely, I followed this approach because it is more readable and easier to port to other cultures. 
I will run the experiments and in case try to merge the unsupported patterns with the existing regexes.

LionbridgeCSII: The impact of the new regex is noticeable only when considering long input strings and in this case it appears to be more performant to merge the new patterns in the existing regexes. 
So, I removed the new regex and modified the existing regexes (only in English and Spanish for now).

tellarin (changes requested):
Looks reasonable. Can you add to the other languages (including EnglishOthers)?
Can you also provide some numbers for the previous perf test you ran? It would be interesting to quantify the previous impact.

LionbridgeCSII:
> Looks reasonable. Can you add to the other languages (including EnglishOthers)?
> Can you also provide some numbers for the previous perf test you ran? It would be interesting to quantify the previous impact.

ok, I will add the other languages too.

For the performance test, I replaced all the inputs in DateExtractor with strings of around 700 words. The mean execution times of the test for the three cases (without new patterns, new patterns implemented in new regex, new patterns merged in existing regexes) are respectively: 16.7s, 20.5s, 17.4s.

LionbridgeCSII:
added patterns (and test cases) to other languages.

Specs/DateTime/Dutch/DateExtractor.json
```diff
@@ -2585,5 +2585,77 @@
     "Comment": "1/4 shouldn't recognized as date here",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": []
+  },
+  {
+    "Input": "Ik ga terug op Sep-23-2020.",
+    "NotSupportedByDesign": "javascript,python,java",
```
tellarin, line 2591: Why disable tests in the other platforms? If the changes are YAML regex only, they should work, no?

Specs/DateTime/Dutch/DateExtractor.json
```diff
@@ -2585,5 +2585,77 @@
...
+      {
+        "Text": "Sep-23-2020",
+        "Type": "date",
+        "Start": 15,
+        "Length": 11
+      }
+    ]
+  },
+  {
+    "Input": "Ik ga terug op September 2020 23.",
```
tellarin, line 2602: I don't think we need to support space-separated cases like these... It will probably lead to false positive merges if there are number mentions around the date. Let's support only dates with proper separators.
Also, "2020/September/23", with the full month name, should be a supported format. Are there spec cases for it?

tellarin (changes requested):

Specs/DateTime/Dutch/DateExtractor.json
LionbridgeCSII: re-enabled but only in English, French and Spanish (the other languages are implemented only in .NET)

Specs/DateTime/Dutch/DateExtractor.json
LionbridgeCSII: agreed, removed patterns with space separators. 
All patterns support both cases, full month name and abbreviation (since these are defined in the same regex MonthRegex).
For this particular pattern, I put in the spec the case with abbreviation "2020/Sep/23", but there are other cases that show support for the full month name (e.g. "September-2020-23").

iMicknl: To chime in, I haven't seen this case in Dutch at all. `23 september 2020` would make sense, or in a notation like

23/sept/2020, 2020/sept/23 or something similar.

LionbridgeCSII: I have not added this kind of pattern (23 september 2020) to the specs because it was already covered and the Dutch specs contain it already (for example the sixth case in DateExtractor "Ik ga terug op 12 januari 2016"). The other two cases have been added in the current PR ("Ik ga terug op 23/Sep/2020" and "Ik ga terug op 2020-Sep-23")

Specs/DateTime/French/DateTimeModel.json
```diff
@@ -3206,6 +3206,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
+    "NotSupported": "java, python",
```
tellarin, line 3209: When adding a commit message like this, please specify in which platforms you're disabling them.

.NET/Microsoft.Recognizers.Definitions.Common/Dutch/DateTimeDefinitions.cs
```diff
@@ -126,9 +126,9 @@ public static class DateTimeDefinitions
       public static readonly string RelativeWeekDayRegex = $@"\b({WrittenNumRegex}\s+{WeekDayRegex}\s+(vanaf\s+nu|later))\b";
       public static readonly string SpecialDate = $@"(?=\b(op\s+)(de\s+)?){DayRegex}\b";
       public const string DatePreposition = @"\b(op(\s+de)?)";
-      public static readonly string DateExtractorYearTermRegex = $@"(\s+|\s*,\s*){DateYearRegex}";
+      public static readonly string DateExtractorYearTermRegex = $@"(\s+|\s*[,./-]\s*){DateYearRegex}";
```
tellarin, line 129: Can you rebase this PR on /master after #2502 merges to make sure changes are tracked from the right pushes and there are no side-effects?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/Dutch/DateTimeDefinitions.cs
tellarin: It's now merged.

LionbridgeCSII: rebased and pushed

tellarin (approved):

PR #2383, opened 2020-11-26
[ES DateTimeV2] Fixed mention formats not supported (#2371)

LionbridgeCSII:
Fix to issue #2371. 

According to a native speaker, cases with "9 a." / "9 p." are not common and have been updated to "9 a.m." / "9 p.m."
Also, two test cases have been added to DateTimeModel to conform the originals to the English counterparts:
- "Volveré el 1.º de julio, 17 veces." -> "Volveré el 1.º de julio, 17.° veces." (I'll go back July 1st, 17th times.)
- "Cortana, ¿puedes preparar algo para el día 21?" -> "Cortana, ¿puedes preparar algo para el día 21 de este mes?" (Cortana, can you please set something up around the 21st this month.)

PR #2439, opened 2021-01-19
[Java] Unify fallback bindLookbehind method for all JDKs

### LINKED ISSUE #1786: [Java] isRestrictedJavaVersion() returns wrong value
**Describe the bug**
[isRestrictedJavaVersion](https://github.com/microsoft/Recognizers-Text/blob/b211ef589d40a99bbf445858d4c76641b0bbe52c/Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java#L399) check if `specVersion` (version on local) **is greater than or equal to** `targetVersion` (version required to Recognizers-Text), if I understood [Java API BigDecimal](https://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html) correctly.

As a result, even if Java >= 11 is installed, `result` variable always get `true`, and it will print out warning message, and fallback method will always be applied.

**To Reproduce**
0. Be sure that your local machine runs with Java version > 8
1. Go to 'Java/samples/simple-console'
2. Type 'mvn exec:java'
3. Type anything
4. See 'WARN: Look-behind groups with no maximum length not supported. Java version <= 8.'

**Expected behavior**
Not seeing such warning message

**Sample input/output**
(Described above)

 - Platform: [Java]
 - Environment: [OpenJDK Runtime Environment Zulu11.33+15-CA (build 11.0.4+11-LTS)]
 - Version of package [master branch]

**Additional context**
I'm digging with the issue #1775 and if I fix the problem described above then more tests fails on `Zulu 11 SDK`. Strange. However, if I understand correctly, then current behavior of isRestrictedJavaVersion() is somewhat misleading.

### LINKED ISSUE #2253: [Java] Test fail during mvn install
Using the instructions in https://github.com/microsoft/Recognizers-Text/blob/master/Java/README.md, I get the following during the "mvn clean install" step.

``````
Failed tests:   test[DateTimeRecognizer - English - DateTimeModel - "Must deliver sixteen (16) units before the thirty-first day of the month." - [2019-08-12T00:00:00]](com.microsoft.recognizers.text.tests.datetime.DateTimeTest): Does not match text on Input: "Must deliver sixteen (16) units before the thirty-first day of the month." expected:<[before the thirty-first day of ]the month> but was:<[]the month>
  test[DateTimeRecognizer - English - DateTimeModel - "We will come back before the twenty-sixth day of the month" - [2020-07-01T12:00:00]](com.microsoft.recognizers.text.tests.datetime.DateTimeTest): Does not match text on Input: "We will come back before the twenty-sixth day of the month" expected:<[before the twenty-sixth day of ]the month> but was:<[]the month>
  test[DateTimeRecognizer - English - DateTimeModel - "We will come back after the 15th day of next month" - [2020-07-01T12:00:00]](com.microsoft.recognizers.text.tests.datetime.DateTimeTest): Does not match text on Input: "We will come back after the 15th day of next month" expected:<[after the 15th day of ]next month> but was:<[]next month>
  test[DateTimeRecognizer - English - DateTimeModel - "We will come back after the 15th day of June" - [2020-07-01T12:00:00]](com.microsoft.recognizers.text.tests.datetime.DateTimeTest): Does not match text on Input: "We will come back after the 15th day of June" expected:<[after the 15th day of ]june> but was:<[]june>

Tests run: 11329, Failures: 4, Errors: 0, Skipped: 4065
``````

As a side note, it would be helpful to include the POM dependency values for the sample in the README:

```
    <dependency>
      <groupId>com.microsoft.recognizers.text</groupId>
      <artifactId>recognizers-text</artifactId>
      <version>1.0-SNAPSHOT</version>
    </dependency>

    <dependency>
      <groupId>com.microsoft.recognizers.text.number</groupId>
      <artifactId>recognizers-text-number</artifactId>
      <version>1.0-SNAPSHOT</version>
    </dependency>
```

### PR DESCRIPTION
**Pull request by Batta32**
Fixes #1786
Fixes #2253 

## Description
The fallback method [bindLookbehinds](https://github.com/microsoft/Recognizers-Text/blob/b211ef589d40a99bbf445858d4c76641b0bbe52c/Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java#L191) can be re-used for all the JDKs.

## Specific Changes
- Unify the execution of default fallback method [bindLoolbehinds](https://github.com/microsoft/Recognizers-Text/blob/b211ef589d40a99bbf445858d4c76641b0bbe52c/Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java#L191) for all the JDKs
- Remove [isRestrictedJavaVersion](https://github.com/microsoft/Recognizers-Text/blob/b211ef589d40a99bbf445858d4c76641b0bbe52c/Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java#L392) method as it's no longer needed
- Use [getSafeLookbehindRegExp](https://github.com/microsoft/Recognizers-Text/blob/b211ef589d40a99bbf445858d4c76641b0bbe52c/Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java#L179) method when it's needed, depending the RegExp to be compiled that contains look-behinds or not.

## Testing
All the tests supported by Java are correctly passing using Java 8, Java 11 and Java 15 as we checked the functionality and implementation.
- Tests in total: 12,671
- Tests passed: 7,920
- Tests ignored: 4,751

image

### CODE REVIEW

**Comment by tellarin** [comment]
@Batta32, we can't disable those spec cases. So it seems in some situations, even with the newer JDK versions we still need some regex processing. Can you try to indentify what the core issue is? Is it a difference in how Java handles look-behinds? 
We may need to have different fallback regex methods. :(

**Review by tellarin — CHANGES_REQUESTED** [review]
The changes seem to make sense, but we can't merge them as they introduce breaks.
We need to figure out if we need different fallback for different Java versions.

**Comment by Batta32** [comment]
@tellarin - thanks for your feedback. We will be reviewing how Java handles look-behinds using different SDKs version and also reviewing the fallback regex method.

**Comment by Batta32** [comment]
Hi @tellarin, we successfully unified the fallback method to be default for all the JDKs. We confirmed that the unit tests are correctly passing with these changes!

File: Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java [hunk 1]
```diff
@@ -177,14 +174,8 @@ public static Pattern getSafeLookbehindRegExp(String source) {
     }
 
     public static Pattern getSafeLookbehindRegExp(String source, int flags) {
-
         String result = source;
-
-        // Java pre 1.9 doesn't support unbounded lookbehind lengths
-        if (unboundedLookBehindNotSupported) {
```
tellarin, line 184: Great to see that calling this fallback method from those two regexes passes all tests in Java 8.
But we should not use the fallback in versions where the declared regex works as is. Using this fallback may lead to perf issues and to side effects in the future.
Can you re-enable the version check that you folks had fixed before? This should properly fix the issue.
Also, why didn't this PR enable the tests Lionbridge disabled in that other PR (mentioned on this issue/PR discussion)? Whatever fix comes from this, should apply to those test cases as well.

tellarin (changes requested):

Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/RegExpUtility.java
tellarin: PR #2383. Spec cases marked as: "Comment": "Java does not correctly handle lookbehinds."
should be enabled and the comment removed.

Batta32: @tellarin, after re-enabling the version check, the unit tests are failing using Java 11. That's why we unified the fallback method for all the JDKs.

We will be working on adding re-enabled specs which were disabled in the PR #2383 in order to add them in this PR.

_5 specs are failing using Java 11_
image

Batta32: @tellarin, we confirmed that re-enabling the mentioned specs of #2383 and incorporating the JDK's validation, the specs are failing no matter which version of Java is used (e.g. Java 8, Java 11).

Also, we found the same problem in JS rollbacking the changes of PR #1069. In that PR, NodeJS version 10.3.0 starts supporting Lookbehind Assertions so all the machinery introduced in RegExpUtility to deal with them was unnecesary.

We tried to apply the same changes, however, a lot of specs were failing. This means that Java doesn't support those spec cases.

Finally, we think that it's necessary to unify the fallback method and the specs are correctly disabled currently. Are you okay with this?

tellarin: This is interesting. The current build passes, so the unbouded look-behind is supported in Java (I guess the build is Java 15). We should check in which Java version this unbounded issue was resolved and that version check to trigger the fallback should reflect that.
Are the other cases from #2383 failing due to the same issue? Or a different look-behind issue?

Batta32: Hi @tellarin, the current build passed as it's using the fallback method by default for all the JDKs in this PR.

We **summarized** all the scenarios in the following list:
1. Using `master` branch
    * ❌Java 8: 62 tests are broken (all for look-behind's issue)
    * ✔Java 11: all tests passed
    * ✔Java 15: all tests passed
2. Using `master` branch re-enabling the disabled tests of PR #2383
    * ❌Java 8: 66 tests are broken (62 for look-behind's issue, 4 assertionerror)
    * ❌Java 11: 4 tests are broken (all for assertionerror), the four are the re-enabled
    * ❌Java 15: 4 tests are broken (all for assertionerror), the four are the re-enabled
3. Branch of PR #2439 with JDKs validation
    * ❌Java 8: 62 tests are broken (all for look-behind's issue)
    * ❌Java 11: 5 tests are broken (all for assertionerror)
    * ❌Java 15: 5 tests are broken (all for assertionerror)
4. Branch of PR #2439 with JDKs validation and re-enabling the disabled tests of PR #2383
    * ❌Java 8: 66 tests are broken (62 for look-behind's issue, 4 assertionerror)
    * ❌Java 11: 9 tests are broken (all for assertionerror), including the re-enabled
    * ❌Java 15: 9 tests are broken (all for assertionerror), including the re-enabled
5. Branch of PR #2439 without JDKs validation (unifying the fallback method)
    * ✔Java 8: all tests passed
    * ✔Java 11: all tests passed
    * ✔Java 15: all tests passed
5. Branch of PR #2439 without JDKs validation (unifying the fallback method) and re-enabling the disabled tests of PR #2383
    * ❌Java 8: 4 tests are broken (all for assertionerror), the four are the re-enabled
    * ❌Java 11: 4 tests are broken (all for assertionerror), the four are the re-enabled
    * ❌Java 15: 4 tests are broken (all for assertionerror), the four are the re-enabled

We came up with 2 **conclusions**:
1. As mentioned above, we think that Java doesn't support lookbehind fully and it's necessary to use the fallback method no matter which version of Java is used 
2. The disabled specs are correct to be disabled

Let us know what do you think about this 😊

tellarin: So it seems the disabled cases in #2383 fail anyways. Any clues on what's the root cause of those 4 failures? The main goal of improving this fallback behaviour is to have a solution that also works for them.
If this is a complex item it could come in its own PR. 
Also, according to your description above, if we have a separate PR for that, the right solution for this one (fixing the current failing tests and the java version check, wihtout re-enabling these four spec cases) if to include your fix to the JDK version check so the fallback only applies to Java 8 and call the method for those two NumberRange regexes. 
Your list item 1) shows that's the only situation where things were failing.

Batta32: @tellarin, thanks for your comment. Your understanding is **right**, we think that we can **merge** this PR, then do the correct JDK validation and fix the broken specs, including the disabled ones of PR #2383.

> I see. So Java 11 and 15 were only working because the fallback was wrongly triggered, right? I.e. the "unbounded lookbehind" issue afects all versions?

Exactly, Java 11 and Java 15 were only working because the fallback method was wrongly triggered. With the correct validation, it will fail some specs (5 for assertionerror)

> And also those 4 cases from PR #2383 fail due to other lookbehind issue Java?

The problem is not a lookbehind issue but an assertion error. Currently, we don't know the root cause of the problem and we think that it would be too complex to address it right now.

tellarin (approved):

PR #2447, opened 2021-01-22
[Java | FR DateTimeV2] Correct regex definition and add ambiguity filter

ChopperMan33:
1. Filter ambiguity.
2. Correct "NotSupported" tags in French DateTimeModel since the full stop is included in the extracted text (even if the extractor returns the correct match).


tellarin (approved):

tellarin:
@ChopperMan33, build still fails due to one not disabled “.” case. Also, please fix the spec comments to mention both Python and Java.

Issue #1528, opened 2019-04-16
[TimexExpression] System.Format exception when parsing Timex Durations with Decimal values.


jpb18:
**Describe the bug**
When using the TimexProperty class to aid in the parsing of an Timex Expression issued by LUIS, we get a System.FormatException if we input a decimal value with a dot. Bug is replicable on the test DataTypes_Parsing_DurationWeeksFloatingPoint:

![image](https://user-images.githubusercontent.com/11457062/56202020-0a60da80-603a-11e9-9f42-2e9c3715134c.png)

The machine where we're running the code is using a Portuguese culture.

**To Reproduce**
Run the DataTypes_Parsing_DurationWeeksFloatingPoint test with a Portuguese culture.

Alternatively, feed the constructor of TimexProperties with the parameters "(2019-04-05,XXXX-04-11,P5.54701493625231D)", using a Portuguese culture.

**Expected behavior**
Using the timex  "(2019-04-05,XXXX-04-11,P5.54701493625231D)", we should get this (explanation of how we got the result in additional context):
![image](https://user-images.githubusercontent.com/11457062/56203483-4184bb00-603d-11e9-9452-cba953716456.png)

**Sample input/output**
If input is  "(2019-04-05,XXXX-04-11,P5.54701493625231D)", there's the following exception:
![image](https://user-images.githubusercontent.com/11457062/56203056-4eed7580-603c-11e9-8413-4fde193aed90.png)

**Platform (please complete the following information):**
 - .Net Framework 4.6.2
 - Portuguese Culture

**Additional context**
We've managed to fix this issue by making our own fork of the library and setting the CultureInfo for the Parse methods in the TimexProperty.AssignDateDuration and TimexProperty.AssignTimeDuration as CultureInfo.InvariantCulture. Here's the commit link: [commit](https://github.com/thecodeshift/Recognizers-Text/commit/c65a4da8546bf25d1790105868661db73795f3f5 ) If this solution is acceptable, we can issue a pull request so it can be included here,

johnataylor:
@tellarin can you comment on this?
is this is legal TIMEX expression. it seems very odd to have P5.54701493625231D I've not seen one like that.

If this is illegal then format exception seems like a reasonable exception.

jpb18:
@johnataylor ISO 8601 states that the duration segment can have a decimal value, and the separator can either be a dot or a comma.

PR #2452, opened 2021-01-26
[TimexExpression] Fix parsing Timex in different cultures in TimexResolover (#1528)

ChopperMan33:
Fix the culture problem of parsing Timex Durations with Decimal values. #1528
All timexes should use only '.' as canonical separator and parsing should work in any culture.
Add unit test to make sure it works in different culture settings.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Globalization;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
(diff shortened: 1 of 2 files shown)


.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
```diff
@@ -305,6 +306,42 @@ public void DataTypes_Resolver_DateRange_Last_Three_Weeks()
             Assert.AreEqual("2019-05-01", resolution.Values[0].End);
         }
 
+        [TestMethod]
+        public void DataTypes_Resolver_DateRange_Demaical_Period_PT()
+        {
+            var sourceLanguage = CultureInfo.InstalledUICulture;
```
tellarin, line 312: Isn't it enough to only read and set `CurrentCulture` instead of also `InstalledUICulture` and `CurrentUICulture`?

tellarin (changes requested):

tellarin (approved):

PR #2454, opened 2021-01-27
[Choice] Add support for emoji skin tone modifiers across cultures

ChopperMan33:
Add support for emoji skin tone modifiers in .NET, Java, JavaScript across cultures.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Arabic/ChoiceDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Arabic/ChoiceDefinitions.cs
@@ -23,7 +23,8 @@ namespace Microsoft.Recognizers.Definitions.Arabic
     {
       public const string LangMarker = @"Ara";
       public const string TokenizerRegex = @"[^\u0621-\u064A0-9]";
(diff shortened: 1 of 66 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/Arabic/ChoiceDefinitions.cs
```diff
@@ -23,7 +23,8 @@ public static class ChoiceDefinitions
     {
       public const string LangMarker = @"Ara";
       public const string TokenizerRegex = @"[^\u0621-\u064A0-9]";
-      public const string TrueRegex = @"\b(صحيح|نعم|حسنا|موافق|متأكد|بالتأكيد|أتفق|اتفق|بالطبع|هيا)\b|(\uD83D\uDC4D|\uD83D\uDC4C|\u0001f44c)";
-      public const string FalseRegex = @"\b(لست متاحا|لا أستطيع|ليس تماما|أختلف|نختلف|اوافق|لست أوافق|لا|ليس صحيح|ليس|ليست|غير موافق)\b|(\uD83D\uDC4E|\u270B|\uD83D\uDD90|\u0001F44E|\u0001F590)";
+      public const string SkinToneRegex = @"(\uD83C\uDFFB|\uD83C\uDFFC|\uD83C\uDFFD|\uD83C\uDFFE|\uD83C\uDFFF)";
```
tellarin, line 26: These timexes could be simplified to avoid repetition and it would also help a bit with perf. But impact doesn't seem too big to block merging.

tellarin (approved):

PR #2461, opened 2021-02-02
[Java][Number] Decrease end property of ModelResult by 1

### LINKED ISSUE #2254: [Java | EN Ordinal] Parse returns a different ModelResult.end value than .NET (off by one)
Calling: 
```model.parse("the first one please")```

Returns a ModelResult with:
```
result.text = "first"
result.start = 4
result.end = 9
result.typeName = "ordinal"
```

In dotnet, result.end is 8.

### PR DESCRIPTION
**Pull request by Batta32**
Fixes #2254

## Description
There is a disparity on the generation of the `ModelResult`, specifically the `end` property, which needs to be decreased by 1. See [C#](https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.Number/Models/AbstractNumberModel.cs#L79) and [Java](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java#L57). Also, [JavaScript](https://github.com/microsoft/Recognizers-Text/blob/e56e3a7fc82aeeb3979ce2c3c61d20a25f4261d3/JavaScript/packages/recognizers-number/src/number/models.ts#L94) follows the behavior of C#.

## Specific Changes
- Decrease the `end` property of `ModelResult` by 1

## Testing
These changes were validated using the PR #2439.

_Parity of ModelResult_
image

_Test passing correctly using Java 8/11/15_
image

### DIFF
--- a/Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java
+++ b/Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java
@@ -54,7 +54,7 @@ public abstract class AbstractNumberModel implements IModel {
             return new ModelResult(
                     o.getText(),
                     o.getStart(),
-                o.getStart() + o.getLength(),
+                o.getStart() + o.getLength() - 1,
                 getModelTypeName(),
                 sortedMap
             );                
--- a/Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java
+++ b/Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java
@@ -51,6 +51,7 @@ public abstract class AbstractNumberModel implements IModel {
             SortedMap<String, Object> sortedMap = new TreeMap<String, Object>();
             sortedMap.put(ResolutionKey.Value, o.getResolutionStr());
 
+            // We decreased the end property by 1 in order to keep parity with other platforms (C#/JS).
             return new ModelResult(
                     o.getText(),
                     o.getStart(),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java [hunk 1]
```diff
@@ -54,7 +54,7 @@ protected AbstractNumberModel(IParser parser, IExtractor extractor) {
             return new ModelResult(
                     o.getText(),
                     o.getStart(),
-                o.getStart() + o.getLength(),
+                o.getStart() + o.getLength() - 1,
```
tellarin, line 57: Can you add a code comment before to return to specify this is to keep behaviour consistent with the other platforms?
Also, how did the *Model.json unit tests pass if this was incorrect?
Can you fix whatever issue there is there in this PR too? Thanks!

tellarin (changes requested):

Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/models/AbstractNumberModel.java
Batta32: Hi @tellarin, we already added the code comment before to `return` as requested.

Also, we noticed that the `start` and `end` were not validated and that's why the tests were passing with/without these changes. We updated the `NumberTest` following C# and JavaScript in order to validate the necessary properties divided in each type of Recognizer.

We confirmed that all the tests are correctly passing 😊.

tellarin: Thanks! I'll review soon. Are the other models correct? DateTime, Units, ...?

Batta32: @tellarin, all the models are validating the `typeName`, `text` and `Resolution` properties of the specs.

If you are okay, we can create an issue improving this behavior for all the models apart from this PR/issue in order to have this PR reviewed/merged as soon as possible 😊.

tellarin: Please create the issue and a new PR. 😊

tellarin (approved):

PR #2463, opened 2021-02-05
[FR DateTimeV2] Add new French specs for DurationParser

wumingxuanceline:
Add 56 new French specs


tellarin (changes requested):
Same problem. Many unrelated files.

tellarin:
@wumingxuanceline, do you think you can update this PR and create the next ones soon?
Please always check the build results after creating a PR too. Thanks.

tellarin:
@wumingxuanceline, @XiaoxiaoMa0815, do you think you can update this PR and create the next ones soon? It's been a while.
Unfortunately currently it is not enough to just run the ValidationTool on the specs as the build process runs some additional validation. Please always check the build results after creating a PR too. The log shows why the build fails.

wumingxuanceline:
> @wumingxuanceline, @XiaoxiaoMa0815, do you think you can update this PR and create the next ones soon? It's been a while.
> Unfortunately currently it is not enough to just run the ValidationTool on the specs as the build process runs some additional validation. Please always check the build results after creating a PR too. The log shows why the build fails.

Hi! Sorry for replying late. We targeted the problem but its seems there is conflict between the fail log and json. Would you please take a check?
d58875c69e31948087568c398a4e8f7
0f1b240fc7d6cc909f1ef4a6e3a5eac

tellarin (approved):

PR #2465, opened 2021-02-05
[Java] Add Sequence Recognizer

Batta32:
## Description
The Sequence Recognizer is missing in Java. This PR contains the migrated recognizer from C# to Java (only code).

This PR contains the English parsers/extractors.

## Specific Changes
- Add main structure of the `recognizers-sequence`
- Add recognizer as submodule of `recognizers-text`
- Migrate English Extractors & Parsers
- Migrate root files
- Integrate Sequence in Simple console
- Overload [trimStart](https://github.com/southworks/Recognizers-Text/blob/master/Java/libraries/recognizers-text/src/main/java/com/microsoft/recognizers/text/utilities/StringUtility.java#L12) of `StringUtility`
- Catch infinite values in [NumberFormatUtility](https://github.com/southworks/Recognizers-Text/blob/master/Java/libraries/recognizers-text-number/src/main/java/com/microsoft/recognizers/text/number/parsers/NumberFormatUtility.java#L33)

## Testing
We created PR #2466 which contains the unit tests passing correctly (524/524) following the C# & JS unit tests.

_Migrated structure of Sequence_
image

image

image

PR #2466, opened 2021-02-05
[Java | * Sequence] Add Sequence Recognizer Unit Tests

Batta32:
Depends on #2465

## Description
The Sequence Recognizer is missing in Java. This PR contains the migrated recognizer from C# to Java (only tests).

## Specific Changes
- Add `SequenceRecognizerCacheTest` unit tests
- Add `SequenceRecognizerInitializationTest` unit tests
- Add `SequenceTest` unit tests
- Skip unsupported tests comparing them against C#/JS

## Testing
All the tests are correctly passing 845/845. These tests were validated using JDK 8/11/15.

_Unit tests passing correctly_
image

image

_Main structure of the migrated tests_
image


tellarin (changes requested):
Same issue as #2465.

Specs/Sequence/Portuguese/IpAddressModel.json
```diff
@@ -1,6 +1,7 @@
 [
   {
     "Input": "O IP do meu computador é 1.1.1.1",
+    "NotSupported":"java",
```
tellarin, line 4: Is this because of the commented out regex? If so, then we should not register the IP recognizer yet. Please disable it on the previous PR until we have a solution.

Specs/Sequence/Korean/PhoneNumberModel.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "Input": "Tel: (+31)12-3456789.",
-    "NotSupported":"javascript",
+    "NotSupported":"javascript,java",
```
tellarin, line 4: Why is this not supported? Can you add comments here or in the PR description for the reason to disable all these specs?

tellarin (changes requested):

Specs/Sequence/Portuguese/IpAddressModel.json
Batta32: We updated the specs that are supported in Java Sequence. This PR contains the extractor of English and some specs are not supported because those regex are not configured (e.g. some specs of Chinese PhoneNumber model).

The current number of specs was improved to 845 passing correctly and 289 ignored.

Specs/Sequence/Korean/PhoneNumberModel.json
Batta32: There is no reason to be `notsupported`, we updated the list of specs 😊

Specs/Sequence/Chinese/IpAddressModel.json
```diff
@@ -1,381 +1,363 @@
...
-        {
-          "Text": "1.1.1.1",
-          "TypeName": "ip",
-          "Resolution": {
-            "value": "1.1.1.1",
-            "type": "ipv4"
-          }
+  {
+    "Input": "我电脑IP是1.1.1.1",
+    "NotSupported": "java",
```
tellarin, line 4: Why are these still disabled in Java only?

tellarin: BTW, please avoid changing line endings in spec files unless intentional. It makes review a bit harder.
I recommend reverting them.

tellarin (changes requested):

Specs/Sequence/Chinese/IpAddressModel.json
Batta32: Hi @tellarin, we reverted the changes and disabled the ones that are not supported by Java.

The non-supported specs are because the Regex doesn't match with the text in Java. For example the input `我电脑IP是1.1.1.1` should be matched with the Regex `\b(1\d{2}|2[0-4]\d|25[0-5]|0?[1-9]\d|0{0,2}\d)((\.(1\d{2}|2[0-4]\d|25[0-5]|0?[1-9]\d|0{0,2}\d)){3})\b` but as you can see in the images it doesn't match using Java engine.

To validate this we used a JavaScript engine and Java engine with:
* [Freeformatter](https://www.freeformatter.com/java-regex-tester.html#ad-output): this is the Java RegEx tester
* [RegExr](https://regexr.com/): settings as JavaScript RegEx engine

Java doesn't match the text with the Regex
image

JavaScript matches the text with the Regex
image

tellarin (approved):
Approving as-is, so we have some unit tests in place for v0 of the Sequence recognizers in Java.

Issue #2453, opened 2021-01-26
Hour was caught from float number


OlhaAntonenko:
**Describe the bug**
Hour number was caught from float number, when merged hour with timerange (some kinds like: morning, in the evening...).

**Sample input/output**
`
model = DateTimeRecognizer(Culture.English).get_datetime_model()
model.parse('My temperature was 37.1 in the morning')
`
Output:
`{
  "text": "1 in the morning",
  "type_name": "datetimeV2.time",
  "resolution": {
    "values": [
      {
        "timex": "T01",
        "type": "time",
        "value": "01:00:00"
      }
    ]
  },
  "start": 22,
  "end": 37
}`

**Expected behavior**
`{
  "resolution": {
    "values": [
      {
        "timex": "TMO",
        "type": "timerange",
        "start": "08:00:00",
        "end": "12:00:00"
      }
    ]
  },
  "text": "in the morning",
  "type_name": "datetimeV2.timerange",
  "start": 15,
  "end": 28,
}`

**Platform (please complete the following information):**
 - Platform: Python
 - Environment: python 3.8.0
 - Version of package 1.0.2a2

PR #2502, opened 2021-02-17
[EN DateTimeV2] Fixed hour incorrectly extracted from float number (#2453)

LionbridgeCSII:
Fix to issue #2453. 


Patterns/English/English-DateTime.yaml
```diff
@@ -374,13 +374,13 @@ TimeRegexWithDotConnector: !nestedRegex
   def: ({BaseDateTime.HourRegex}(\s*\.\s*){BaseDateTime.MinuteRegex})
   references: [ BaseDateTime.HourRegex, BaseDateTime.MinuteRegex ]
 TimeRegex1: !nestedRegex
-  def: \b({TimePrefix}\s+)?({WrittenTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex})(\s*|[.]){DescRegex}
+  def: \b({TimePrefix}\s+)?({WrittenTimeRegex}|{HourNumRegex}|(?<!\d[,.]){BaseDateTime.HourRegex})(\s*|[.]){DescRegex}
   references: [ TimePrefix, WrittenTimeRegex, HourNumRegex, BaseDateTime.HourRegex, DescRegex ]
```
tellarin, line 378: We need a solution that works across all cultures. Shouldn’t perhaps an approach like this be applied at the BaseDateTime level?
Also, ideally we should also have at least a workaround for Java, if possible.

tellarin (changes requested):

Patterns/English/English-DateTime.yaml
LionbridgeCSII: ok, moved implementation to base code (following the solution already used in DateExtractor) and re-enabled test case in Java.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
```diff
@@ -222,7 +222,14 @@ public List<Token> TimeOfTodayAfter(string text, DateObject reference)
 
             var matches = this.config.SimpleTimeOfTodayAfterRegex.Matches(text);
             foreach (Match match in matches)
-            {
+            {
```
tellarin, line 225: This is not exactly what I meant. I meant modifying the regexes like BaseDateTime.HourRegex in the BaseDateTime.YAML file, if possible.
I worry that this new implementation might cause a perf hit. Can you measure the performance impact of it vs changing the YAML file?
Also, could you add some spec test cases in non-English languages to show that the fix works in them too?
Thanks.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
LionbridgeCSII: ah I see, ok I will do and will check if it is possible modifying BaseDateTime.YAML.

tellarin:
Can you make sure this fix also covers cases like "Labs were notable for a BUN/Cr of 30/4.85, troponin 0.182"?
Dates should not be extracted with mixed separators (except some corner cases with space and comma for year only).

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
LionbridgeCSII: moved solution to Base-DateTime.yaml (but kept themodifications to base code in Java because of the problems with negative lookbehinds).

LionbridgeCSII: noticed also that FilterAmbiguity is not implemented in JavaScript MergedExtractor (this causes problems for example with the input "Mi temperatura era 37.1 por la mañana" in Spanish where "Mi" is extracted as weekday (abbreviation of "miercoles")). 
Should I open an issue?

LionbridgeCSII:
> Can you make sure this fix also covers cases like "Labs were notable for a BUN/Cr of 30/4.85, troponin 0.182"?
> Dates should not be extracted with mixed separators (except some corner cases with space and comma for year only).

ok, added filter to exclude dates with mixed separators to ValidateMatch in DateExtractor.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
tellarin: Please open an issue for the ambiguity filter in JS and mention this spec case there. Thanks.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -124,12 +124,12 @@ public static class DateTimeDefinitions
...
-      public static readonly string DateExtractor6 = $@"(?<={DatePreposition}\s+)({StrictRelativeRegex}\s+)?({WeekDayRegex}\s+)?{MonthNumRegex}[\-\.]{DayRegex}(?![%])\b";
+      public static readonly string DateExtractor6 = $@"(?<={DatePreposition}\s+)({StrictRelativeRegex}\s+)?({WeekDayRegex}\s+)?{MonthNumRegex}[\-\.]{DayRegex}(?![%])(?![,.]\d)\b";
       public static readonly string DateExtractor7L = $@"\b({WeekDayRegex}\s+)?{MonthNumRegex}\s*/\s*{DayRegex}{DateExtractorYearTermRegex}(?![%])\b";
-      public static readonly string DateExtractor7S = $@"\b({WeekDayRegex}\s+)?{MonthNumRegex}\s*/\s*{DayRegex}(?![%])\b";
-      public static readonly string DateExtractor8 = $@"(?<={DatePreposition}\s+)({StrictRelativeRegex}\s+)?({WeekDayRegex}\s+)?{DayRegex}[\\\-]{MonthNumRegex}(?![%])\b";
+      public static readonly string DateExtractor7S = $@"\b({WeekDayRegex}\s+)?{MonthNumRegex}\s*/\s*{DayRegex}(?![%])(?![,.]\d)\b";
+      public static readonly string DateExtractor8 = $@"(?<={DatePreposition}\s+)({StrictRelativeRegex}\s+)?({WeekDayRegex}\s+)?{DayRegex}[\\\-]{MonthNumRegex}(?![%])(?![,.]\d)\b";
```
tellarin, line 130: Couldn't "(?![,.]\d)" be also a regex in the base YAML then referenced elsewhere? This would make regexes more readable and help make sure all languages use it correctly.

tellarin: BTW, why only English is using it now?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
tellarin: And this check should now be removed from .NET, right?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
```diff
@@ -90,6 +90,13 @@ private List<Token> BasicRegexMatch(string text)
 
                 foreach (Match match in matches)
                 {
+                    // Check that the extracted time is not part of a decimal number (e.g. 123.24)
```
tellarin, line 93: Shouldn't this have been removed after the YAML changes?

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseTimeExtractor.java
```diff
@@ -88,6 +88,13 @@ public BaseTimeExtractor(ITimeExtractorConfiguration config) {
 
             Match[] matches = RegExpUtility.getMatches(regex, text);
             for (Match match : matches) {
+                
+                // @TODO Remove when lookbehinds are handled correctly
```
tellarin, line 92: Can you make this a method?

tellarin (changes requested):

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -19448,5 +19448,37 @@
...
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Labs were notable for a BUN/Cr of 30/4.85, troponin 0.182",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupported": "java, javascript, python",
```
tellarin, line 19481: Can the filter be easily added in the other platforms?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDateTimeExtractor.cs
LionbridgeCSII: yes sorry, I forgot to update the .NET files.

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: I will check if it is possible

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseTimeExtractor.java
LionbridgeCSII: mmm, but the if statement returns a "continue" when matched and as far as I know it is not possible to return an instruction from a method in Java. So, I could put the current if condition in a method which returns a boolean and then add a new if statement that "continue" when the method returns "true". Is that what you mean?

tellarin: Yes, as I think it would make the code a little more readable and it then could be reused in other filtering or in lambdas.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
tellarin: @LionbridgeCSII

LionbridgeCSII: moved regex to base YAML and added it to other languages.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimeExtractor.cs
LionbridgeCSII: yes, removed now.

Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/extractors/BaseTimeExtractor.java
LionbridgeCSII: done.

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: added it to Java and Python, in Javascript the method validateMatch is not implemented.

Python/libraries/recognizers-date-time/recognizers_date_time/date_time/base_date.py
```diff
@@ -302,6 +302,20 @@ def validate_match(self, match: Match, text: str):
                     sub_text = text[text.index(year_group) + len(year_group):].strip()
                     sub_text = self.trim_start_range_connector_symbols(sub_text)
                     is_valid_match = self.starts_with_basic_date(sub_text)
+                    
+            # Expressions with mixed separators are not considered valid dates e.g. "30/4.85" (unless one is a comma "30/4, 2016")
```
tellarin, line 306: This should be a method, like in .NET and Java, right?

Patterns/Spanish/Spanish-DateTime.yaml
```diff
@@ -394,7 +394,7 @@ NowRegex: !simpleRegex
 SuffixRegex: !simpleRegex
   def: ^\s*(((y|a|en|por)\s+la|al)\s+)?(mañana|madrugada|medio\s*d[ií]a|(?<!(m[áa]s\s+))tarde|noche)\b
 TimeOfDayRegex: !nestedRegex
-  def: \b(?<timeOfDay>(({LaterEarlyRegex}\s+)((del?|en|por)(\s+(el|los?|las?))?\s+)?|(en|por)\s+(el|l[ao]s?)\s+)?(mañana|madrugada|pasado\s+(el\s+)?medio\s?d[ií]a|(?<!((m[áa]s|tan)\s+))tarde|noche))\b
+  def: \b((?<timeOfDay>(({LaterEarlyRegex}\s+)((del?|en|por)(\s+(el|los?|las?))?\s+)?)?(mañana|madrugada|pasado\s+(el\s+)?medio\s?d[ií]a|(?<!((m[áa]s|tan)\s+))tarde|noche))|(en|por)\s+las?\s+mañana)\b
```
tellarin, line 397: Why is this causing a problem in Java? Also, now it's not fully correct, right? As cases like "por la madrugada" and others without later/early won't be captured anymore.

Python/libraries/recognizers-date-time/recognizers_date_time/date_time/base_date.py
LionbridgeCSII: also in Java and .NET, I added this check as a further step in the already existing method ValidateMatch

Patterns/Spanish/Spanish-DateTime.yaml
LionbridgeCSII: I have not fully understood the reason for the failure, I can investigate the problem in more details if needed. 
Now the regex is in its original form, a part for the additional pattern "(en|por)\s+las?\s+mañana". 
For a case like "por la madrugada", it will extract only "madrugada". 
The additional pattern "por la mañana" is needed because isolated extractions of "mañana" are filtered out due to the ambiguity in the meaning "morning/tomorrow".

tellarin (approved):

tellarin:
@LionbridgeCSII, can you resolve the conflicts here so we can merge? Thanks.

LionbridgeCSII:
> @LionbridgeCSII, can you resolve the conflicts here so we can merge? Thanks.

@tellarin, confilcts solved.

tellarin (approved):

Issue #2473, opened 2021-02-09
[ES DateTimeV2] "anoche" not properly handled in resolution


cahann:
EN: last night at 8  Returns datetime with yesterday's date
PT: ontem à noite às 8 horas  (same)
ES: anoche a las 8 en punto  Returns datetime with today's date

PR #2503, opened 2021-02-18
[ES DateTimeV2] Fixed "anoche" not properly handled in resolution (#2473)

LionbridgeCSII:
Fix to issue #2473. 


.NET/Microsoft.Recognizers.Text.DateTime/Spanish/Parsers/SpanishDateTimeParserConfiguration.cs
```diff
@@ -1,176 +1,187 @@
...
-        }
-
-        public bool ContainsAmbiguousToken(string text, string matchedText)
-        {
-            // @TODO move hardcoded values to resources file
-            return text.Contains("esta mañana") && matchedText.Contains("mañana");
-        }
-    }
-}
+﻿using System;
```
tellarin, line 1: Please be careful to not modify line endings unless intentional.

tellarin (approved):

PR #2506, opened 2021-02-22
[JA Number] Improvement for extractor timeout in long number input

iofu728:
## Reproduce
```json
input: "金融市場で運用する資金運用事業を合わせた財投総額も過去最大の同二・七％減の四十九兆九千五百九十二億円となっている。"
```
When use c# version 1.5.0, the `NumberWithUnitRecognizer.RecognizeCurrency` will get stuck endlessly.

## Fixing detail
The problem caused by `NumbersAggressiveRegex` Regex in https://github.com/microsoft/Recognizers-Text/blob/master/Patterns/Japanese/Japanese-Numbers.yaml#L174 and the problem only happend in c# version, in python this regex is normal.
```yaml
PercentageRegex: !simpleRegex
  def: .+(?=パ\s*ー\s*セ\s*ン\s*ト)|.*(?=[％%])
NumbersAggressiveRegex: !nestedRegex
  def: (({AllIntRegex})(?!({AllIntRegex}*([、.]{ZeroToNineIntegerRegex}+)*|{AllFloatRegex})*\s*{PercentageRegex}*))
```
It's mainly because the `PercentageRegex` and `{PercentageRegex}*` increased the computational complexity.


Patterns/Japanese/Japanese-Numbers.yaml
```diff
@@ -171,7 +171,7 @@ NumbersWithAllowListRegex: !nestedRegex
   def: '(?<!(離は))({NegativeNumberTermsRegex}?({NotSingleRegex}|{SingleRegex})(?!({AllIntRegex}*([、.]{ZeroToNineIntegerRegex}+)*|{AllFloatRegex})*\s*{PercentageRegex}+))(?!(\s*{AllMultiplierLookupRegex}))'
   references: [NegativeNumberTermsRegex, NotSingleRegex, SingleRegex, AllIntRegex, ZeroToNineIntegerRegex, AllFloatRegex, PercentageRegex, AllMultiplierLookupRegex]
 NumbersAggressiveRegex: !nestedRegex
-  def: (({AllIntRegex})(?!({AllIntRegex}*([、.]{ZeroToNineIntegerRegex}+)*|{AllFloatRegex})*\s*{PercentageRegex}*))
+  def: (({AllIntRegex})(?!({AllIntRegex}*([、.]{ZeroToNineIntegerRegex}+)*|{AllFloatRegex})*\s*{PercentageRegex}+))
```
tellarin, line 174: This is likely to still be quite slow, right? How long does the equivalent case in ZH take to run?
Can you also add a corresponding Chinese spec case in this PR?

tellarin (changes requested):

tellarin (approved):
Approving as-is as it partially mitigates the original issue.

PR #2507, opened 2021-02-23
[JA DateTimeV2] Holiday extractor refinements

LionbridgeCSII:
All test cases pass. 

The extraction of the case "明日清明だけどどこ行く？" has been modified because it was wrongly pointing to "明日" (tomorrow) instead of the holiday "清明" (Qingming).

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Japanese.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Japanese.cs
@@ -76,14 +76,6 @@ namespace Microsoft.Recognizers.Text.DateTime.Tests
 
         [NetCoreTestDataSource]
         [TestMethod]
-        public void TimeZoneExtractor(TestModel testSpec)
-        {
-            ExtractorInitialize(Extractors);
-            TestDateTimeExtractor(testSpec);
-        }
-
-        [NetCoreTestDataSource]
-        [TestMethod]
         public void DurationExtractor(TestModel testSpec)
         {
             ExtractorInitialize(Extractors);
@@ -106,11 +98,11 @@ namespace Microsoft.Recognizers.Text.DateTime.Tests
             TestDateTimeExtractor(testSpec);
         }
 
-        [NetCoreTestDataSource]
+        /*[NetCoreTestDataSource]
         [TestMethod]
         public void DateTimeModel(TestModel testSpec)
         {
             TestDateTime(testSpec);
-        }
+        }*/
     }
 }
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Japanese.cs
```diff
@@ -28,29 +28,81 @@ public void DateExtractor(TestModel testSpec)
...
+        [TestMethod]
+        public void SetExtractor(TestModel testSpec)
+        {
+            ExtractorInitialize(Extractors);
+            TestDateTimeExtractor(testSpec);
         }
 
         [NetCoreTestDataSource]
         [TestMethod]
-        public new void DateTimeParser(TestModel testSpec)
```
tellarin, line 49: Why remove the other methods? The test code should have them all, for all types.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DataDrivenTests/DateTime/TestDateTime_Japanese.cs
LionbridgeCSII: ok, reintroduced (I thought that since only extraction support is required for the language, the parser tests were not needed).

tellarin: Full support is required for all languages, so parsers should never be disabled. Even if extraction is prioritized first in some cases, the whole E2E flow needs to work and the parser tests also updated. 
There is a special spec attribute for parser test cases in the json files that are not yet supported - 'ExtractOnly'.

Specs/DateTime/Japanese/DateTimeModel.json
```diff
@@ -15147,7 +15147,7 @@
     "Context": {
       "ReferenceDateTime": "2017-03-22T00:00:00"
     },
-    "NotSupported": "javascript, python, java",
+    "NotSupported": "dotnet, javascript, python, java",
```
tellarin, line 15150: The DateTimeModel is not currently registered to run, so should we really disable these?
I'd pefer to disable them only if they don't run when tried.
If you have tried them, then it's ok to keep the new tag.

LionbridgeCSII: yes I tried, they need to be disabled otherwise they will give an error (Could not find Model).

tellarin: I see. Thanks.

tellarin (approved):

PR #2508, opened 2021-02-25
[DateTime] Update SpanishDateTime resource

Batta32:
## Description
The [SpanishDateTimeParserConfiguration](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/spanish/parsers/SpanishDateTimeParserConfiguration.java) doesn't recognize the [SpanishDateTime.LastNightTimeRegex](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/spanish/parsers/SpanishDateTimeParserConfiguration.java#L84) as the resource is not updated. The compilation fails without updating the resource.

## Specific Changes
Update the [SpanishDateTime](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/resources/SpanishDateTime.java) resource with the latest changes.

## Testing
_Java build correctly passing_
image

### DIFF
--- a/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/resources/SpanishDateTime.java
+++ b/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/resources/SpanishDateTime.java
@@ -369,6 +369,16 @@ public class SpanishDateTime {
 
     public static final String PmTimeRegex = "(?<pm>(esta|(por|de|a|en)\\s+la)\\s+(tarde|noche))";
 
+    public static final String NightTimeRegex = "(noche)";
+
+    public static final String LastNightTimeRegex = "(anoche)";
+
+    public static final String NowTimeRegex = "(ahora|mismo|momento)";
+
+    public static final String RecentlyTimeRegex = "(mente)";
+
+    public static final String AsapTimeRegex = "(posible|pueda[ns]?|podamos)";
+
     public static final String LessThanOneHour = "(?<lth>((\\s+y\\s+)?cuarto|(\\s*)menos cuarto|(\\s+y\\s+)media|{BaseDateTime.DeltaMinuteRegex}(\\s+(minutos?|mins?))|{DeltaMinuteNumRegex}(\\s+(minutos?|mins?))))"
             .replace("{BaseDateTime.DeltaMinuteRegex}", BaseDateTime.DeltaMinuteRegex)
             .replace("{DeltaMinuteNumRegex}", DeltaMinuteNumRegex);


tellarin (changes requested):
This is not the proper way to handle resource updates. The build process should automatically update such resource code files during the build process (via the YAML files).
Perhaps there is a silent failure somewhere in the resource generation process?

Batta32:
Hi @tellarin, these changes were added after executing the [build.cmd](https://github.com/microsoft/Recognizers-Text/blob/master/Java/build.cmd) of Java, they are **not manually updated**.

The compilation of [datetime](https://github.com/microsoft/Recognizers-Text/tree/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime) is **failing** as the property `LastNightTimeRegex` is **not present** in [SpanishDateTime](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/resources/SpanishDateTime.java) and it's already used in [SpanishDateTimeParserConfiguration](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/spanish/parsers/SpanishDateTimeParserConfiguration.java#L84). However, the `build.cmd` won't fail because, as you mentioned, these changes are generated but they are not present in the repository.

To reproduce the issue using `master` branch:
1. Open [Java](https://github.com/microsoft/Recognizers-Text/tree/master/Java) Recognizers-Text in 
2. Build [recognizers-text-date-time](https://github.com/microsoft/Recognizers-Text/tree/master/Java/libraries/recognizers-text-date-time) library
3. See the compilation error

_Compilation error in Datetime recognizers of Java_
image

tellarin:
@Batta32, I meant to say that if the build process generates such files correctly, than there can never be a compilation failure.
The correct fix is to try to identify if the resource code is not generated at the right time and fix the build process.
Perhaps the resource gen step is happening after the relevant project is being built?
Adding the generated code file would just be a mitigation that would hide the root problem until the next YAML update.

Batta32:
@tellarin - doing the [manual build](https://github.com/microsoft/Recognizers-Text/blob/master/Java/README.md#manual-build-and-installation-on-local-maven-folder-m2) (executing `mvn clean install`) will fail because of this. As you said, the `build.cmd` will generate such files correctly as they are consuming the updated YAMLs.

C# is working as PR #2503 introduced the changes in the [DateTimeDefinitions](https://github.com/microsoft/Recognizers-Text/pull/2503/files#diff-58f903028057f6dfa68b4633a436015420ef3273a66073da1e54ddaa085c35d9), however [Java](https://github.com/microsoft/Recognizers-Text/blob/master/Java/libraries/recognizers-text-date-time/src/main/java/com/microsoft/recognizers/text/datetime/resources/SpanishDateTime.java)/[JavaScript](https://github.com/microsoft/Recognizers-Text/blob/master/JavaScript/packages/recognizers-date-time/src/resources/spanishDateTime.ts)/[Python](https://github.com/microsoft/Recognizers-Text/blob/master/Python/libraries/recognizers-date-time/recognizers_date_time/resources/spanish_date_time.py) don't contain those last changes in the _autogenerated resources_. 

Just to know, which should be the correct way to update the autogenerated files in order to have the `mvn clean install` working?

tellarin:
@Batta32, think we're talking past each other.
The original design is for the build process to auto-re-generate any resource code from YAMLs if the YAML is newer than the corresponding code in the repo.
One reason for this is exactly to avoid such problems as you are seeing, as many times contributors change the YAML file in one platform, but not in the others. But the build process is supposed to catch if such new changes cause build issues, so for that it needs to generate the necessary code updates.
.NET implements it like this (only re-gen if YAML is newer than .cs code), while JS and Python re-gen the resources on every build.
If build.cmd for Java re-gens the resources on every build and the build doesn't fail, couldn't "mvn clean install" do the same?
Otherwise the build process will never validate that "mvn clean install" will work correctly. That's what I mean when I say that adding the file may mitigate the break you mention, but is not solving the root issue and it will likely break again as soon as someone updates some YAML file.

Batta32:
Awesome @tellarin! Thanks for the explanation. I'll close this PR then 😊.

Issue #2488, opened 2021-02-09
[FR DateTimeV2] Resolving a weekday + time-range construction as an invalid range


cahann:
When the date portion of this expression is a weekday, the value returned is a range where both endpoints are the current datetime:
•	FR: "Jeudi avant 17h" returns (2021-01-15T02:13:32,2021-01-15T02:13:32,PT17H)
•	EN: "Thursday before 5 p.m." returns a reasonable value: XXXX-WXX-4T17/"before".
Note: The same behavior occurs when a relative date like "demain" ("tomorrow") is substituted for the weekday.
When the date portion is a calendar date, the returned value is a reasonable datetimerange:
•	FR: "23/12 avant 17h" returns (XXXX-XX-23T12,XXXX-XX-23T17,PT5H)

tellarin:
@LionbridgeCSII, can you prioritise this one and #2489? I’ve also marked some issues as ‘ongoing’ to avoid publication of effort.
BTW, before spending too much time in a new issue fix, please comment on the issue you are going to tackle so others are aware and don’t work on the same issue.

LionbridgeCSII:
@tellarin, ok will do.

Issue #445, opened 2018-04-06
[FR DateTimeV2] Hours duration wrongly recognized


thomasryck:
Ex:
If I say: " Quelles sont les salles disponibles aujourd'hui pendant 2 heures ?" LUIS returns that "pendant 2 heures" is a time whereas it is supposed to be a duration . This issue only appears when you say "during X hours". This issue does not exist with other time units as minutes, seconds, days, months ...

tellarin:
@matthewshim-ms could you take a look at this?

PR #2510, opened 2021-03-02
[FR DateTimeV2] Fixed weekday + time-range construction producing an invalid range (#2488)

LionbridgeCSII:
Fix to issue #2488. As for issue #445, the problem is "heures" (hours) is ambiguous in French and can indicate both a duration "dans 2 heures" (in 2 hours) or a time "il est 2 heures" (it is 2 o'clock).
Solved by rearranging the parser methods in BaseDateTimePeriodParser so that Duration patterns are parsed after DateWithTimePeriod patterns. This does not seem to cause issues because Duration patterns are still parsed correctly e.g. "Je serai de retour dans les 2h" ("I'll be back within 2h").
An alternative solution could be to introduce a FilterAmbiguity method in BaseDurationExtractor analogous to the one in BaseTimeExtractor.

Test cases added to English and French DateTimeModels.


tellarin:
@LionbridgeCSII can you resolve the conflicts?

LionbridgeCSII:
@tellarin, conflicts solved.

tellarin (approved):

PR #2520, opened 2021-03-09
[.NET] Korean DateTime code skeleton + Holiday support

LionbridgeCSII:
- Base code skeleton for Korean DateTime branching from English.
- Added support for flag IgnoreResolution in DateTime tests in TestBase following Number implementation.
- Support for Korean Holiday (only extraction), all test cases pass.
- Added new test cases in HolidayExtractor corresponding to Korean holidays.


tellarin:
Sorry, shouldn't it branch from Chinese as they have similar root and to encourage reuse between Chinese/Japanese/Korean (CJK) like Numbers?

LionbridgeCSII:
> Sorry, shouldn't it branch from Chinese as they have similar root and to encourage reuse between Chinese/Japanese/Korean (CJK) like Numbers?

Since we are not completely aware of the differences between English and Chinese implementations at the moment, we decided to consider English first because following the standard implementation seems preferable (if possible). In case it is not, we will switch to Chinese implementation. 
Or is it better in this case to start directly from Chinese?

tellarin:
I'd say for Korean/Japanese it's probably better to start from Chinese, but for re-use we'd need to introduce customization points there. Sorry for not raising this earlier.
The idea is to not duplicate everything though. The base classes can be the same base for CJK and then follow the same idea as in other languages and have specific config classes for KO/JA.
Let us know if you identify any issues/blockers/etc and we can discuss the best path forward.

LionbridgeCSII:
> The idea is to not duplicate everything though. The base classes can be the same base for CJK and then follow the same idea as in other languages and have specific config classes for KO/JA.

Ok, so just to be sure to get the picture right, we are going to create base CJK extractors and parsers (e.g. BaseCJKDateExtractor) to contain the common methods (Extract, Parse,...) that are now defined in the Chinese configuration files and we will modify the configuration files to contain only the config classes for CH/KO/JA. Correct?

tellarin:
That's the sugested approach.

LionbridgeCSII:
@tellarin, a blocking issue we noticed is that the Chinese implementation uses NumberWithUnit which is not implemented for Korean, so we will have to add support for it before proceding.

tellarin (changes requested):
This PR should be abandoned in favour of #2566, right?

LionbridgeCSII:
> This PR should be abandoned in favour of #2566, right?

yes

tellarin:
Closing as superseded by #2566.

PR #2523, opened 2021-03-12
[FR Number|DateTime] Additonal French Specs in improving language coverage

wumingxuanceline:
Updates all specs in one PR


wumingxuanceline:
@tellarin 
Hi Börje, we've updated all files and all checks have passed. Please feel free to let us know for any questions. Thanks.

Specs/DateTime/French/DateTimeModel.json
```diff
@@ -3202,210 +3202,17153 @@
     ]
   },
   {
-    "Input": "Je retournerai 4:45 du matin",
```
tellarin, line 3205: No previously existing case should be removed. Unless it is clearly incorrect and this should be noted in the PR description.
If a duplicate case was inserted, please remove the new one, not the original. This faciliates merging as multiple people may be working over the same files.

Specs/DateTime/French/DateTimeModel.json
```diff
@@ -3202,210 +3202,17153 @@
...
+    "Results": [
+      {
+        "Text": "d'avril à juin 2017",
+        "Start": 19,
+        "End": 38,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(2017-04-0
```
tellarin, line 3347: Same and so on.

Specs/DateTime/French/DateTimeModelCalendarMode
```diff
@@ -0,0 +1,951 @@
+[
```
tellarin, line 1: All spec files must have names ending in json. Can you please correct this one and the others?

tellarin (changes requested):
There are some small issues, specially in the DateTimeModel. json specs.
Are all these cases translation? Will there be a new PR with new French additions of forms not existing in English?

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+        "Resolution": {
+          "value": "(,100]"
+        },
+        "Start": 38,
+        "End": 45
+      }
+    ]
+  },
+  {
+    "Input": "Sa taille est en dessuous de 170.",
```
tellarin, line 383: Typo? Shouldn't it be "dessous"?

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "en dessuous de 170",
```
tellarin, line 387: Why should the "en" be included?

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+        "End": 30
+      }
+    ]
+  },
+  {
+    "Input": "Moins d’un millier de pandas géants vivent encore dans la nature.",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "Moins d’un millier de",
```
tellarin, line 417: "de" should not be tagged.

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+        "End": 42
+      }
+    ]
+  },
+  {
+    "Input": "La gamme de nombres est de 1000 à 5000",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "1000 à 5000",
```
tellarin, line 816: "de 1000 à 5000"

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+      }
+    ]
+  },
+  {
+    "Input": "inférieur ou supérieur à 30",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": []
+  },
+  {
+    "Input": "égal ou supérieur à 30",
```
tellarin, line 967: Why are these next three empty?

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+        "End": 26,
+        "TypeName": "numberrange",
+        "Resolution": {
+          "value": "[20000,20000]"
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Le nombre est passé de 200 à 300 en 2008",
```
tellarin, line 992: Why is this empty?

Specs/Number/French/NumberRangeModel
```diff
@@ -0,0 +1,1359 @@
...
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Plus de 40 et moins de 10 sont deux gammes",
+    "NotSupported": "dotnet, javascript, python, java",
+    "Results": [
+      {
+        "Text": "Plus de 40",
```
tellarin, line 1139: All model-level outputs are lowercase.

tellarin: Please review all.

Specs/Number/French/NumberRangeModel
wumingxuanceline: The whole phrase "en dessous de" = under

Specs/Number/French/NumberRangeModel
wumingxuanceline: French translation should be consistent with original English data. In this case, if there is an "of" in an English output text, the French translation will also contain "de", vice versa.

Specs/Number/French/NumberRangeModel
wumingxuanceline: According to the rules, French translation should be consistent with original English data and we should try to avoid preposition. Therefore, for example, if an English output text is "from 1000 to 5000'', the translation will be "de 1000 à 5000", but if the English one is "1000 to 5000'' or "1000-5000'', the French output will also be "1000 à 5000".

Specs/Number/French/NumberRangeModel
wumingxuanceline: Because original English data are empty.

Specs/Number/French/NumberRangeModel
wumingxuanceline: Same.

tellarin (approved):
I've fixed the DateTimeModel.json file, so the build passes and no previous test spec is lost.
There are still some issues, especially in the tagged "Text" attributes. But I'll merge as-is and triage bugs later.

PR #2525, opened 2021-03-16
[.NET] Arabic DateTime code skeleton + Holiday support

LionbridgeCSII:
- Base code skeleton for Arabic DateTime branching from English.
- Added support for flag IgnoreResolution in DateTime tests in TestBase following Number implementation.
- Support for Arabic Holiday (only extraction), all test cases pass.
- Added new test cases in HolidayExtractor corresponding to Arabic holidays.

We modified ArabicNumberParserConfiguration to replace the culture "ar-*" with "ar-sa" instead of "ar-ms" (we get the error "ar-ms is an invalid culture identifier" otherwise).


tellarin:
> We modified ArabicNumberParserConfiguration to replace the culture "ar-*" with "ar-sa" instead of "ar-ms" (we get the error "ar-ms is an invalid culture identifier" otherwise).

Any specific reason to not keep it as "ar-*"? Even if the current implementation will be standard only, the goal is to try to have a single Arabic model for all variants, if possible.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Extractors/ArabicDateTimeAltExtractorConfiguration.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Recognizers.Definitions.Arabic;
+
+namespace Microsoft.Recognizers.Text.DateTime.Arabic
+{
+    public class ArabicDateTimeAltExtractorConfiguration : BaseDateTimeOptionsConfiguration, IDateTimeAltExtractorConfiguration
```
tellarin, line 7: This sub-type has lower priority. Feel free to only work on it once all others are in place.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Parsers/ArabicDateTimeAltParserConfiguration.cs
```diff
@@ -0,0 +1,27 @@
+﻿namespace Microsoft.Recognizers.Text.DateTime.Arabic
```
tellarin, line 1: Same as for the Extractor.

tellarin:
> > We modified ArabicNumberParserConfiguration to replace the culture "ar-*" with "ar-sa" instead of "ar-ms" (we get the error "ar-ms is an invalid culture identifier" otherwise).
> 
> Any specific reason to not keep it as "ar-*"? Even if the current implementation will be standard only, the goal is to try to have a single Arabic model for all variants, if possible.

I see what you mean. Internally representing it as "ar-sa", but model would accept "ar-*" as input. No worries then.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Extractors/ArabicDateTimeAltExtractorConfiguration.cs
LionbridgeCSII: you mean we should add tests and specs also for DateTimeModelComplexCalendar and DateTimeModelExtendedTypes? (They seem to be the ones using BaseDateTimeAltExtractor)
If so, what about the other models (CalendarMode, ExperimentalMode, SplitDateAndTime), should we add them too?

tellarin: Ideally yes, but they are even lower priority because they don't introduce new types. Thanks for catching this.

LionbridgeCSII:
@tellarin, is there some change required for this PR, or can it be merged as it is? (apart for the additional tests and specs which however, being low-priority, could also be added in a later stage, right?)

tellarin:
> @tellarin, is there some change required for this PR, or can it be merged as it is? (apart for the additional tests and specs which however, being low-priority, could also be added in a later stage, right?)

These were just some preliminary comments. We'll review and merge soon.

tellarin (approved):

tellarin:
@LionbridgeCSII, could you please add these 4 cases as specs in the appropriate files in a future PR to make sure they're also properly supported? 
ارتفاع أسعار الذهب في مصر اليوم 6 مايو.. العيار يقفز 4 جنيهات | بوابة أخبار اليوم الإلكترونية 
لخميس، 06 مايو 2021 - 10:50 ص
يونايتد يسعى لإنهاء مهمته أمام روما | يلاكورة | 2021 
الخميس 6 مايو 202101:08 ص

Issue #2178, opened 2020-06-17
[* Sequence] Recognize quoted strings as entities


tomlm:
Quotes are used in language as a way of mention in another work of the title of a short or subsidiary work.

_such as a chapter or an episode: "Doctor K" is the eleventh episode of Power Rangers RPM._

Recognizing quotes as entities would be super useful in language recognition scenarios.  For example:

* _change the meeting title to "getting help"_

recognizing that "getting help" is the title and not a request for help is super useful.

It turns out that while quotes are easy to recognize, there are a lot of little variations across the languages.  Luckily, wikipedia has already enumerated them, if you look at the **Summary Table** in this article:
https://en.wikipedia.org/wiki/Quotation_mark

That table boils down to these patterns:
"…"
'…'
“…”
‘…’
„…“
«…»
‹…›
«…»‎
”…”
’…’
„…”
‘…’
「…」
‚…‘
»…«
›…‹
«…»
«…»
“…”
”…”‎
’…’‎
『…』
《…》
“…”
‚…‘
’…‘
〈…〉
‹…›‎

Which can be matched with this regex:
```
(?<=")(.*)(?=")|(?<=')(.*)(?=')|(?<=“)(.*)(?=”)|(?<=‘)(.*)(?=’)|(?<=„)(.*)(?=“)|(?<=«)(.*)(?=»)|(?<=‹)(.*)(?=›)|(?<=«)(.*)(?=‎)|(?<=”)(.*)(?=”)|(?<=’)(.*)(?=’)|(?<=„)(.*)(?=”)|(?<=「)(.*)(?=」)|(?<=‚)(.*)(?=‘)|(?<=»)(.*)(?=«)|(?<=›)(.*)(?=‹)|(?<=”)(.*)(?=‎)|(?<=’)(.*)(?=‎)|(?<=『)(.*)(?=』)|(?<=《)(.*)(?=》)|(?<=’)(.*)(?=‘)|(?<=〈)(.*)(?=〉)|(?<=‹)(.*)(?=‎)
```

There is a slight problem with this regex, which is that it is too greedy when dealing with multiple quoted strings.

```this is a "test" of "quoted"```

Will return **test" of "quoted** as the match instead of **test** and **quoted**

tomlm:
BTW I have an library implemented here:

[...]

PR #2528, opened 2021-03-22
Porting quoted text recognizer into project

coldplaying42:
Add quoted text recognizer support.
v0 of a implementation for #2178.


.NET/Microsoft.Recognizers.Text.Sequence/Chinese/Extractors/ChineseQuotedTextExtractor.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Recognizers.Text.Sequence.Chinese
+{
+    public class ChineseQuotedTextExtractor : BaseQuotedTextExtractor
+    {
+        private static Dictionary<Regex, string> regexes = new Dictionary<Regex, string>
+            {
+                {
+                    new Regex(@"(“([^“”]+)”)"),
```
tellarin, line 12: These regex declarations should be in a YAML file for Chinese. Can you refactor these before adding new languages?
Should this with double-quotes be in Base though as it would fir multiple languages?

.NET/Microsoft.Recognizers.Text.Sequence/English/Extractors/EnglishQuotedTextExtractor.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Recognizers.Text.Sequence.English
+{
+    public class EnglishQuotedTextExtractor : BaseQuotedTextExtractor
+    {
+        private static Dictionary<Regex, string> regexes = new Dictionary<Regex, string>
+            {
+                {
+                    new Regex(@"(‘([^‘’]+)’)"),
```
tellarin, line 12: Same comments as for Chinese.

.NET/Microsoft.Recognizers.Text.Sequence/SequenceRecognizer.cs
```diff
@@ -105,6 +110,18 @@ public IModel GetHashtagModel(string culture = null, bool fallbackToDefaultCultu
             return GetModel<HashtagModel>(Culture.English, fallbackToDefaultCulture);
         }
 
+        public IModel GetQuotedTextModel(string culture = null, bool fallbackToDefaultCulture = true)
+        {
+            if (culture != null && (
+           culture.ToLowerInvariant().StartsWith("zh-", StringComparison.Ordinal) ||
```
tellarin, line 116: This existed in other types as there were only two models. But in your case there will be a bunch, right? Or also only two main ones that cover all cultures?

tellarin, line 116: Minor: Fix code indentation.

Specs/Sequence/Chinese/QuotedTextModel.json
```diff
@@ -0,0 +1,28 @@
+[
+	{
```
tellarin, line 2: Fix indentation. All specs use whitespace and not tabs.

Specs/Sequence/English/QuotedTextModel.json
```diff
@@ -0,0 +1,54 @@
+[
+	{
+		"Input": "This “number” is `123` or `456`",
```
tellarin, line 3: What happens for cases like these below?
- Between 2" and 5".
- It should say: "I finished the race under 1'46"8 yesterday."
We should have well defined behaviour for both and add specs to show they work right.

Specs/Sequence/Chinese/QuotedTextModel.json
```diff
@@ -0,0 +1,28 @@
...
+		"Input": " 这个“数字”是『123』或者『456』",
+		"Results": [
```
tellarin, line 4: As this new type only exists in .NET, you should add one annotation per spec case to disable tests in the other platforms.

tellarin (changes requested):

Specs/Sequence/English/QuotedTextModel.json
coldplaying42: I have discussed with Chris, and we think currently these corner cases could be ignored.

tellarin: OK, let's keep the spec and mark as NotSupported.

Specs/Sequence/Chinese/QuotedTextModel.json
coldplaying42: I find that sometime when NotSupportedByDesign is added in the QuotedTextModel json file, the tests will fail.

tellarin: This should not happen. I'll take a look. Sorry for the delay.

coldplaying42:
I have already implemented quoted text recognizer for 12 languages (without Arabic, Swedish and Bulgarian). Now, there are two issues:
1)	For each language, the maximum number of quote types is 9. I use QuotedTextRegex1, QuotedTextRegex2, …, QuotedTextRegex9 to represent them in the yaml file. If the number of quote types is fewer than 9, the QuotedTextRegex9 maybe look like (123456) (the dummy regex). I am not sure if there is a better way to implement this?
2)	For some languages, if NotSupportedByDesign is added in the QuotedTextModel.json, the tests will fail. I still could not find the reason why it happens.

Specs/Sequence/Chinese/QuotedTextModel.json
coldplaying42: Thanks 👍

tellarin (approved):
Current design has two limitations:
- Nested or false positives of non-quote use of quotation chars
- Encoding of some code chars in input and output
But I'm OK with merging this as v0 once the build passes.

PR #2529, opened 2021-03-22
[.NET DateTime] Added base CJK Extractors and Parsers

LionbridgeCSII:
- Moved Extract and Parse methods from DateTime Chinese configuration files to new BaseCJK[Type]Extractor and BaseCJK[Type]Parser files.
- Updated Chinese configuration files according to the new implementation (Japanese and Korean will follow if the approach is approved).

In MergedParser, I was planning to move common methods (used in both standard and CJK implementations) to a MergedParserUtil, but the implementations are slightly different and produce different results, for example:
- In standard implementation, the modifiers "after" and "since" convert the "value" key to "start" and the modifiers "before" and "until" convert the "value" key to "end".
- In CJK, only the modifiers "after" and "before" modify the "value" key while "since" and "until" keep it unchanged.

Is this a wanted feature? If not, should I continue to refactor the common methods and update the Chinese Model specs accordingly?


LionbridgeCSII:
@tellarin, could you have a look at this one? We need to know if the approach is correct in order to continue the refactoring and Korean development. Thanks!

tellarin (commented):
I'm about halfway through the review. So far all looks good.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Extractors/ChineseTimePeriodExtractorConfiguration.cs
```diff
@@ -1,59 +1,58 @@
-﻿using System.Collections.Generic;
```
tellarin, line 1: Be careful with unintentional line ending changes. They just make reviewing harder.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
```diff
@@ -1,769 +1,155 @@
...
-            }
+            NumberParser = config.NumberParser;
 
-            var numConfig = new BaseNumberOptionsConfiguration(config.Culture, numOptions);
+            DateExtractor = config.DateExtractor;
+            DurationExtractor = config.DurationExtractor;
 
-            integerExtractor = new IntegerExtractor(numConfig);
-            ordinalExtractor = new OrdinalExtractor(numConfig);
+            DateRegexList = new ChineseDateExtractorConfiguration(this).DateRegexList;
```
tellarin, line 26: This seems a bit strange.

tellarin:
> * In CJK, only the modifiers "after" and "before" modify the "value" key while "since" and "until" keep it unchanged.

All four modifiers should behave the same. Can you open an issue for this?

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseMergedDateTimeParserConfiguration.cs
```diff
@@ -1,135 +0,0 @@
-﻿using System;
-using System.Text.RegularExpressions;
-using Microsoft.Recognizers.Definitions.Chinese;
-
-using DateObject = System.DateTime;
-
-namespace Microsoft.Recognizers.Text.DateTime.Chinese
```
tellarin, line 7: What happened to this class?

tellarin: Became ChineseMergedParserConfiguration.cs?

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -68,9 +68,10 @@ protected override void InitializeConfiguration()
             RegisterModel<DateTimeModel>(
                 Culture.Chinese,
                 options => new DateTimeModel(
-                    new FullDateTimeParser(
```
tellarin, line 71: FullDateTimeParser was not modified in this PR? What the key difference between it and the new BaseCJKMergedDateTimeParser?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateParser.cs
```diff
@@ -0,0 +1,750 @@
...
+                {
+                    num = Convert.ToInt32((double)(this.config.NumberParser.Parse(er[0]).Value ?? 0));
+                }
+            }
+
+            return num;
+        }
+
+        // convert Chinese Year to Integer
+        private int ConvertChineseYearToInteger(string yearChsStr)
```
tellarin, line 702: Minor: Such methods should be renamed to "CJK" instead of "Chinese".

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+
+            var yearStr = match.Groups["year"].Value;
+            var writtenYearStr = match.Groups["fullyear"].Value;
+
+            if (!string.IsNullOrEmpty(yearStr) && !yearStr.Equals(writtenYearStr, StringComparison.Ordinal))
+            {
+                year = int.Parse(yearStr, CultureInfo.InvariantCulture);
+                if (year < 100 && year >= Constants.MinTwoDigitYearPastNum)
+                {
+                    year += 1900;
```
tellarin, line 211: Minor: Reference existing constants.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+                }
+                else if (year >= 0 && year < Constants.MaxTwoDigitYearFutureNum)
+                {
+                    year += 2000;
```
tellarin, line 215: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+
+            return ret;
+        }
+
+        public List<DateTimeParseResult> FilterResults(string query, List<DateTimeParseResult> candidateResults)
+        {
+            return candidateResults;
+        }
+
+        public int GetYearFromText(Match match)
```
tellarin, line 199: Is this implementation exactly the same as in the regular DatePeriodParser?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+                            year += 2000;
+                        }
+                    }
+                }
+            }
+
+            return year;
+        }
+
+        private static DateObject ComputeDate(int cardinal, int weekday, int month, int year)
```
tellarin, line 289: This should be in a DateUtils.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+        }
+
+        // parse entities that made up by two time points
+        private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject referenceDate)
+        {
+            var ret = new DateTimeResolutionResult();
+            var er = this.config.DateExtractor.Extract(text, referenceDate);
+            if (er.Count < 2)
+            {
+                er = this.config.DateExtractor.Extract("on " + text, referenceDate);
```
tellarin, line 1047: "on "? Was this already in the Chinese code before?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+            }
+
+            ret.FutureValue = ret.PastValue = new Tuple<DateObject, DateObject>(beginDay, endDay);
+            ret.Success = true;
+
+            return ret;
+        }
+
+        // parse entities that made up by two time points
+        private DateTimeResolutionResult MergeTwoTimePoints(string text, DateObject referenceDate)
```
tellarin, line 1041: Is there something here that's CJK specific?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+
+            ret.FutureValue = new Tuple<DateObject, DateObject>(futureBegin, futureEnd);
+            ret.PastValue = new Tuple<DateObject, DateObject>(pastBegin, pastEnd);
+            ret.Success = true;
+
+            return ret;
+        }
+
+        // handle like "前两年" "前三个月"
+        private DateTimeResolutionResult ParseNumberWithUnit(string text, DateObject referenceDate)
```
tellarin, line 1115: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+                return ret;
+            }
+
+            var cardinalStr = match.Groups["cardinal"].Value;
+            var monthStr = match.Groups["month"].Value;
+            var noYear = false;
+            int year;
+
+            int cardinal;
+            if (cardinalStr.Equals("最后一", StringComparison.Ordinal))
```
tellarin, line 1310: Would a followup PR move these to YAML and update the relevant Chinese config and config interface?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -0,0 +1,1622 @@
...
+                }
+                else if (this.config.IsNextYear(yearRel))
+                {
+                    year++;
+                }
+            }
+
+            if (year < 100 && year >= this.config.TwoNumYear)
+            {
+                year += 1900;
```
tellarin, line 1493: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateTimePeriodParser.cs
```diff
@@ -0,0 +1,598 @@
...
+        public static readonly string ParserName = Constants.SYS_DATETIME_DATETIMEPERIOD;
+
+        private readonly ICJKDateTimePeriodParserConfiguration config;
+
+        public BaseCJKDateTimePeriodParser(ICJKDateTimePeriodParserConfiguration configuration)
+        {
+            config = configuration;
+        }
+
+        public static string BuildTimex(TimeResult timeResult)
```
tellarin, line 24: This should move to TimexUtil and merge with something equivalent already there.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateTimePeriodParser.cs
```diff
@@ -0,0 +1,598 @@
...
+
+            if (timeResult.Second >= 0)
+            {
+                build.Append(":" + timeResult.Second.ToString("D2", CultureInfo.InvariantCulture));
+            }
+
+            return build.ToString();
+        }
+
+        public static TimeResult DateObject2TimeResult(DateObject dateTime)
```
tellarin, line 45: Also move to util?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKMergedDateTimeParser_with_mergedparserutil.c#
```diff
@@ -0,0 +1,218 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Microsoft.Recognizers.Text.Utilities;
+using DateObject = System.DateTime;
+
+namespace Microsoft.Recognizers.Text.DateTime
+{
+    public class BaseCJKMergedDateTimeParser : IDateTimeParser
```
tellarin, line 10: What is this?

tellarin (changes requested):
I've tried to focus on the structure as there are some pieces that still need refactoring to move Chinese-specific items to the relevant configs. Looks mostly OK, but I didn't have the time to pull the code into an IDE and really check finer details.
I've left some questions and mostly small feedback.
Thanks for the great effort!

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseDateParserConfiguration.cs
LionbridgeCSII: it is done at the same way in English and the other languages.

.NET/Microsoft.Recognizers.Text.DateTime/Chinese/Parsers/ChineseMergedDateTimeParserConfiguration.cs
LionbridgeCSII: yes, I changed the name to be in line with the other languages.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: it is a bit different from the standard version, for example it does not implement the following steps:
- ParseNowAsDate
- // Propagate the possible future relative context from the first entity to the second one in the range.
- // Check if weekPrefix is already included in the extractions otherwise include it
- // Expressions like "today", "tomorrow",... should keep their original year

I am not sure if these steps are required in the CJK implementation, and since the corresponding regexes are not defined in the Chinese yaml file, I have kept the two versions separate for now. Possibly we can merge them (and move them to some utils file) if it turns out to be necessary as we develop Japanese and Korean.

tellarin: Can you add a `@TODO` code comment here as a reminder of that?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: this method exists only in CJK implementation.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: yes, in this PR we focused only on the refactoring of the Base CJK Extractor/Parser without touching the various methods used by each entity. The idea is that, as we proceed in the Japanese/Korean development, we will inspects these methods more closely and move any Chinese specific reference to the config files.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateTimePeriodParser.cs
LionbridgeCSII: it turned out this method was not used anywhere, so I deleted it.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateTimePeriodParser.cs
LionbridgeCSII: also this one was not used, so removed.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
LionbridgeCSII: They should be the same, I introduced BaseCJKMergedDateTimeParser instead of keep using FullDateTimeParser to be in line with the standard implementation. 
I think FullDateTimeParser can be removed once the refactoring of Japanese and Korean is completed.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateParser.cs
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: yes, added a @TODO to use the method defined in AbstractYearExtractor.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: moved to DateContext

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: yes, replaced with TokenBeforeDate as in standard implementation

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKMergedDateTimeParser_with_mergedparserutil.c#
LionbridgeCSII: sorry, this was included by mistake, removed now. 
(it is the version of BaseCJKMergedDateTimeParser with common methods moved to Utils, I will use this solution in a follow up PR since you confirmed Chinese should behave like English regarding the modifiers "after", "since", "before" and "until")

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27130.2026
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30128.74
```
tellarin, line 4: Be careful not to change the solution file unless necessary.

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -869,7 +869,6 @@ Global
 		{F1005971-0341-4635-AC1B-B272681D0834}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{F1005971-0341-4635-AC1B-B272681D0834}.Release|Any CPU.Build.0 = Release|Any CPU
 		{E6BAD17F-2FB0-4A3C-B897-BA9331DEB88A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{E6BAD17F-2FB0-4A3C-B897-BA9331DEB88A}.Debug|Any CPU.Build.0 = Debug|Any CPU
```
tellarin, line 872: What was this?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateParser.cs
```diff
@@ -0,0 +1,750 @@
...
+            if (this.config.MonthOfYear.ContainsKey(monthStr) && this.config.DayOfMonth.ContainsKey(dayStr))
+            {
+                month = this.config.MonthOfYear[monthStr] > 12 ? this.config.MonthOfYear[monthStr] % 12 : this.config.MonthOfYear[monthStr];
+                day = this.config.DayOfMonth[dayStr] > 31 ? this.config.DayOfMonth[dayStr] % 31 : this.config.DayOfMonth[dayStr];
+                if (!string.IsNullOrEmpty(yearStr))
+                {
+                    year = int.Parse(yearStr, CultureInfo.InvariantCulture);
+                    if (year < 100 && year >= Constants.MinTwoDigitYearPastNum)
+                    {
+                        year += 1900;
```
tellarin, line 513: Minor: Same constants.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.sln
LionbridgeCSII: yes I just noticed, I reverted these changes

.NET/Microsoft.Recognizers.Text.sln
LionbridgeCSII: reverted also this

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateParser.cs
LionbridgeCSII: done

tellarin (approved):
Approved. But I'll hold merge for around a day to see if I can run some more local tests.

Issue #2493, opened 2021-02-09
[* DateTimeV2] Resolving 2-digit year mentions


cahann:
When faced with a 2-digit year, LUIS returns a timex with a century value of 20 if year is less than 30, 19 if the year is greater than 30, and it omits the year entirely if it's in the 30s.
input    =>  date string  => timex
12/12/29 => 12/12/29 => 2029-12-12
12/12/30 => 12/12      => XXXX-12-12
12/12/39 => 12/12      => XXXX-12-12
12/12/40 => 12/12/40 => 1940-12-12
Presumably this threshold will change over time.
We're wondering why the year is thrown out entirely (instead of returning both 1930-12-12 and 2030-12-12)

tellarin:
This is due to an old design decision. The goal is to allow users to specifify the thresholds if they want to override the default and to also return the alternatives. But this is currently a breaking change.
The 30s case seems to be a bug though. We'll double check.

LionbridgeCSII:
checking this

PR #2531, opened 2021-03-24
[* DateTimeV2] Resolving 2-digit year mentions (#2493)

LionbridgeCSII:
Fixed extraction of dates with two-digit years in the range [30,40) (e.g. 12/12/35) (#2493).
As suggested, these ambiguous patterns are resolved to both values 1935-12-12 and 2035-12-12.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
@@ -852,11 +852,11 @@ namespace Microsoft.Recognizers.Text.DateTime
                     year = int.Parse(yearStr, CultureInfo.InvariantCulture);
                     if (year < 100 && year >= Constants.MinTwoDigitYearPastNum)
                     {
-                        year += 1900;
+                        year += Constants.BASE_YEAR_PAST_CENTURY;
                     }
                     else if (year >= 0 && year < Constants.MaxTwoDigitYearFutureNum)
                     {
-                        year += 2000;
+                        year += Constants.BASE_YEAR_CURRENT_CENTURY;
                     }
                     else if (year >= Constants.MaxTwoDigitYearFutureNum && year < Constants.MinTwoDigitYearPastNum)
                     {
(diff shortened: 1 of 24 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -893,6 +899,17 @@ private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDat
...
             ret.Success = true;
 
+            // Ambiguous two-digit years are assigned values in both centuries (e.g. 35 -> 1935, 2035)
+            if (ambiguousCentury)
+            {
+                ret.PastValue = futurePastDates.past.AddYears(1900);
+                ret.FutureValue = futurePastDates.future.AddYears(2000);
+                if (ret.Timex.StartsWith("00"))
+                {
+                    ret.Timex = "XX" + ret.Timex.Substring(2);
```
tellarin, line 909: Timex creation should come from a method. Perhaps in TimexUtils. Pass the 2-year digit and the ambiguous flag and get the timex back.
Do we need this check for "00"? Won't all cases that fall here always be ambiguous? If we need the check, move it inside the method.

Patterns/English/English-DateTime.yaml
```diff
@@ -92,7 +92,7 @@ OfPrepositionRegex: !simpleRegex
   def: (\bof\b)
 # Exclude cases that include the "Am/Pm" suffix
 TwoDigitYearRegex: !nestedRegex
-  def: \b(?<![$])(?<year>([0-24-9]\d))(?!(\s*((\:\d)|{AmDescRegex}|{PmDescRegex}|\.\d)))\b
+  def: \b(?<![$])(?<year>([0-9]\d))(?!(\s*((\:\d)|{AmDescRegex}|{PmDescRegex}|\.\d)))\b
```
tellarin, line 95: Doesn't this issue affect also other languages?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
```diff
@@ -893,6 +899,17 @@ private DateTimeResolutionResult Match2Date(Match match, DateObject referenceDat
             ret.PastValue = futurePastDates.past;
             ret.Success = true;
 
+            // Ambiguous two-digit years are assigned values in both centuries (e.g. 35 -> 1935, 2035)
+            if (ambiguousCentury)
+            {
+                ret.PastValue = futurePastDates.past.AddYears(1900);
```
tellarin, line 905: I'd prefer if 1900 and 2000 are constants in Constants.cs 
Maybe BASE_YEAR_PAST_CENTURY and BASE_YEAR_CURRENT_CENTURY?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
LionbridgeCSII: moved Timex modification to TimexUtility.

Patterns/English/English-DateTime.yaml
LionbridgeCSII: yes, modified regex in other languages too.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDateParser.cs
LionbridgeCSII: done

tellarin (approved):

PR #2533, opened 2021-03-25
[.NET] Dutch DateTime Set support

LionbridgeCSII:
In SetExtractor, all test cases pass.
In SetParser, 2 cases fail: 
"Laten we vrijdag om de week afspreken" and "Laten we vrijdag om de andere week afspreken" which correspond to "Every other Friday" in English. The different structure of the Dutch expression (literally "Friday every other week") prevents these cases to be parsed in the current implementation.

A few cases have been modified in SetExtractor and the original spec moved to DateTimeModel:
- "Ik zal vertrekken om 09:00 's ochtends" -> "Ik ga weg iedere ochtend om 9 uur" (I leave every morning at 9am)
- "Ik zal vertrekken 's ochtends om 9" -> "Ik ga weg om 9 uur iedere ochtend" (I leave at 9am every morning)
In the above two cases, the original sentences do not represent Set entities but Time.
- "Ik zal elke maandag vertrekken om 09:00" -> "Ik zal elke maandag om 09:00 vertrekken" (I will leave every Monday at 9am)
In the original sentence the verb splits Set and Time that consequently cannot be extracted together as a single entity.

In DateTimeModel: 166 pass, 199 fail.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
@@ -32,6 +32,9 @@ namespace Microsoft.Recognizers.Text.DateTime.Dutch
         public static readonly Regex SetEachRegex =
             new Regex(DateTimeDefinitions.SetEachRegex, RegexFlags);
 
+        public static readonly Regex BeforeEachDayRegex =
+           new Regex(DateTimeDefinitions.BeforeEachDayRegex, RegexFlags);
+
         private const RegexOptions RegexFlags = RegexOptions.Singleline | RegexOptions.ExplicitCapture;
 
         public DutchSetExtractorConfiguration(IDateTimeOptionsConfiguration config)
@@ -60,7 +63,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Dutch
 
         public IDateTimeExtractor DateTimePeriodExtractor { get; }
 
-        bool ISetExtractorConfiguration.CheckBothBeforeAfter => DateTimeDefinitions.CheckBothBeforeAfter;
+        bool ISetExtractorConfiguration.CheckBothBeforeAfter => true;
 
         Regex ISetExtractorConfiguration.LastRegex => SetLastRegex;
 
@@ -72,7 +75,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Dutch
 
         Regex ISetExtractorConfiguration.EachDayRegex => EachDayRegex;
 
-        Regex ISetExtractorConfiguration.BeforeEachDayRegex => null;
+        Regex ISetExtractorConfiguration.BeforeEachDayRegex => BeforeEachDayRegex;
 
         Regex ISetExtractorConfiguration.SetWeekDayRegex => SetWeekDayRegex;
 
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
@@ -63,6 +63,8 @@ namespace Microsoft.Recognizers.Text.DateTime.Dutch
 
         public IDateTimeExtractor DateTimePeriodExtractor { get; }
 
+        // CheckBothBeforeAfter normally gets its value from DateTimeDefinitions.CheckBothBeforeAfter which however for Dutch is false.
+        // It only needs to be true in SetExtractor.
         bool ISetExtractorConfiguration.CheckBothBeforeAfter => true;
 
         Regex ISetExtractorConfiguration.LastRegex => SetLastRegex;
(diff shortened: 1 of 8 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
```diff
@@ -60,7 +63,7 @@ public DutchSetExtractorConfiguration(IDateTimeOptionsConfiguration config)
 
         public IDateTimeExtractor DateTimePeriodExtractor { get; }
 
-        bool ISetExtractorConfiguration.CheckBothBeforeAfter => DateTimeDefinitions.CheckBothBeforeAfter;
+        bool ISetExtractorConfiguration.CheckBothBeforeAfter => true;
```
tellarin, line 66: Why not read the config property?

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -534,18 +534,20 @@ DurationUnitRegex: !nestedRegex
...
+  def: \b(?<each>(iedere|elke|eenmaal per)\s*$)
 SetEachRegex: !simpleRegex
-  def: \b(?<each>(iedere|elke)\s*)
+  def: \b(?<each>(iedere|elke|om de)\s*(?<other>\s+andere)?\s*(week)?)
 SetLastRegex: !simpleRegex
   def: (?<last>volgende?|komende|aankomende|aanstaande|deze|huidige|aanstaande|vorige?|verleden|laatste)
 EachDayRegex: !simpleRegex
-  def: ^\s*(elke)\s*dag\b
+  def: ^\s*(iedere|elke)\s*dag\b
+BeforeEachDayRegex: !simpleRegex
```
tellarin, line 549: Can you add a comment in the YAML here to explain this one?

tellarin (changes requested):
Looks mostly food to me. @iMicknl, could you take a look?

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
LionbridgeCSII: CheckBothBeforeAfter is false in the YAML, we need to set it to true only in SetExtractor.

Patterns/Dutch/Dutch-DateTime.yaml
```diff
@@ -1364,4 +1366,16 @@ YearToDateTerms: !list
...
+DayTypeRegex: !simpleRegex
+  def: ^(dag(elijks)?)$
+WeekTypeRegex: !simpleRegex
+  def: ^(wekelijks|week)$
+BiWeekTypeRegex: !simpleRegex
+  def: ^(tweewekelijks)$
+MonthTypeRegex: !simpleRegex
+  def: ^(maand(elijks)?)$
+QuarterTypeRegex: !simpleRegex
+  def: ^(kwartaal|driemaandelijkse)$
```
iMicknl, line 1378:
```suggestion
  def: ^(kwartaal|driemaandelijks)$
```

Form would be `driemaandelijks` instead of `driemaandelijkse`. (Same form as DayTypeRegex, WeekTypeRegex etc.).

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
iMicknl: Not sure if other cases require it as well, but if this is the case for now, would be good to have a comment here + in the YAML config. Otherwise it will become very messy in the future when we try to use the flag in YAML.

Patterns/Dutch/Dutch-DateTime.yaml
iMicknl: In the end, best way forward would be to add the optional 'e' suffix to all forms (`wekelijks(e)?`), since this is a common form used in a sentence. But not sure if this is something that should be part of this PR.

Example: wekelijkse boodschappen. (weekly groceries)

iMicknl (commented):
Looks good to me! Just a small remark. Many of the (manually) translated testcases are indeed following the original English form a bit too much sometimes, thus changing some of them is not an issue at all.

.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Extractors/DutchSetExtractorConfiguration.cs
LionbridgeCSII: ok, comment added.

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: added.

Patterns/Dutch/Dutch-DateTime.yaml
LionbridgeCSII: added optional 'e' suffix to these patterns.

iMicknl: @tellarin do these new additions be covered by specs as well? If so; hereby some example sentences that could make sense:

` Lees de beste artikelen op het gebied van Contoso via onze wekelijkse nieuwsbrief!`
(Read the best articles on the field of Contoso via our weekly newsletter)

`Laten we een driemaandelijkse vergadering houden. `
(Let's have a quarterly meeting.)

tellarin: @LionbridgeCSII, can you add these cases suggested by @iMicknl?

LionbridgeCSII: added

Issue #2527, opened 2021-03-17
[EN DateTimeV2] False positives detecting abbreviated month name as DateTime incorrectly


zoinkydoink:
Following text seem to pick up December when it has nothing to do with month, if it has other chars around the short form, it shouldnt detect it as a month.

> $DEC farming is coming to $CUB Finance!

LionbridgeCSII:
Looking into this.

Issue #2509, opened 2021-02-26
[JavaScript | * DateTimeV2] Add missing filter methods


LionbridgeCSII:
**Is your feature request related to a problem? Please describe.**

Javascript is missing the implementation of some filter methods. 
In particular: FilterAmbiguity in MergedExtractor and ValidateMatch in DateExtractor.
This causes issues with certain patterns, for example:
- the Spanish input "Mi temperatura era 37.1 por la mañana" returns the wrong extraction "Mi" (as abbreviation of "miercoles") because the result is not filtered by FilterAmbiguity.
- the input "Labs were notable for a BUN/Cr of 30/4.85, troponin 0.182" returns the wrong extraction "30/4.85" as date because the result is not excluded by ValidateMatch.

**Describe the solution you'd like**

Add the methods FilterAmbiguity and ValidateMatch in Javascript MergedExtractor and DateExtractor as implemented in the other platforms.

LionbridgeCSII:
looking this one

LionbridgeCSII:
@tellarin, has this issue been fixed?

tellarin:
Thanks for the catch. Closed by mistake.

PR #2534, opened 2021-03-25
[EN DateTimeV2] Fixed false positives detecting abbreviated month name (#2527)

LionbridgeCSII:
Fix to issue #2527.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
@@ -62,7 +62,8 @@ namespace Microsoft.Recognizers.Definitions.English
       public const string WeekDayRegex = @"\b(?<weekday>(?:sun|mon|tues?|thurs?|fri)(day)?|thu|wedn(esday)?|weds?|sat(urday)?)s?\b";
       public const string SingleWeekDayRegex = @"\b(?<weekday>(?<!(easter|palm)\s+)sunday|(?<!easter\s+)saturday|(?<!(easter|cyber)\s+)monday|mon|(?<!black\s+)friday|fri|(?:tues?|thurs?)(day)?|thu|wedn(esday)?|weds?|((?<=on\s+)(sat|sun)))\b";
       public static readonly string RelativeMonthRegex = $@"(?<relmonth>((day\s+)?of\s+)?{RelativeRegex}\s+month)\b";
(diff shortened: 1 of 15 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -62,7 +62,8 @@ public static class DateTimeDefinitions
       public const string WeekDayRegex = @"\b(?<weekday>(?:sun|mon|tues?|thurs?|fri)(day)?|thu|wedn(esday)?|weds?|sat(urday)?)s?\b";
       public const string SingleWeekDayRegex = @"\b(?<weekday>(?<!(easter|palm)\s+)sunday|(?<!easter\s+)saturday|(?<!(easter|cyber)\s+)monday|mon|(?<!black\s+)friday|fri|(?:tues?|thurs?)(day)?|thu|wedn(esday)?|weds?|((?<=on\s+)(sat|sun)))\b";
       public static readonly string RelativeMonthRegex = $@"(?<relmonth>((day\s+)?of\s+)?{RelativeRegex}\s+month)\b";
-      public const string WrittenMonthRegex = @"(((the\s+)?month of\s+)?(?<month>apr(il)?|aug(ust)?|dec(ember)?|feb(ruary)?|jan(uary)?|july?|june?|mar(ch)?|may|nov(ember)?|oct(ober)?|sept(ember)?|sept?))";
+      public const string MonthRegex = @"(?<month>april|august|december|february|january|july|june|march|november|october|september|(?<![$%£&!?@#]|\p{L})(apr|aug|dec|feb|jan|jul|jun|mar|may|nov|oct|sept?)(?![$%£&@#]|\p{L}))";
```
tellarin, line 65: Wouldn't perf be better if this is handled in the ambiguity filter?
Also, shouldn't this be applied to all languages?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
LionbridgeCSII: ah probably yes since the wrong extractions should be less frequent that the correct ones. 
So I will move it to the ambiguity filter and consider the other languages too.

LionbridgeCSII: moved fix to ambiguity filter and applied to other languages. 
But the new approach is not supported in JavaScript (FilterAmbiguity not implemented, #2509).

tellarin (approved):

tellarin:
@LionbridgeCSII, can you solve the conflicts, so we can merge this? Tks.

LionbridgeCSII:
> @LionbridgeCSII, can you solve the conflicts, so we can merge this? Tks.

@tellarin, conflicts solved.

PR #2538, opened 2021-03-31
[.NET] Korean NumberWithUnit partial support

LionbridgeCSII:
AgeModel: 8 pass, 1 only extraction, 7 fail.
CurrencyModel: 50 pass, 16 only extraction, 42 fail.
DimensionModel: 41 pass, 11 fail.
TemperatureModel: 8 pass, 24 only extraction.

The not resolved and not supported cases are due mostly to problems in the Number Recognizer which is not fully developed and fails to extract or parse certain numbers (especially large ones). Other issues are given by the Korean way of handling units (e.g. splitting them around the number like in "시속 188 마일" literally "per hour 188 miles").


.NET/Microsoft.Recognizers.Text.NumberWithUnit/NumberWithUnitRecognizer.cs
```diff
@@ -551,6 +551,62 @@ protected override void InitializeConfiguration()
                     },
                 }));
 
+            RegisterModel<CurrencyModel>(
+                Culture.Korean,
+                (options) => new CurrencyModel(new Dictionary<IExtractor, IParser>
```
tellarin, line 556: Minor: Please follow alphabetic order per culture.

Patterns/Korean/Korean-NumbersWithUnit.yaml
```diff
@@ -0,0 +1,781 @@
...
+  entries:
+    - 살|년|세|일|주|월|개월
+AgeSuffixList: !dictionary
+  types: [string, string]
+  entries:
+    Year: 살|년|세
+    Month: 월|개월
+    Week: 주
+    Day: 일
+#ChineseNumberWithUnitExtractorConfiguration
```
tellarin, line 14: Minor: Update comment.

Patterns/Korean/Korean-NumbersWithUnit.yaml
```diff
@@ -0,0 +1,781 @@
...
+    Dominican peso: 도미니칸 페소 #多米尼加比索
+    Mexican peso: 멕시코 페소 #墨西哥比索
+    Philippine peso: 필리핀 페소 #菲律宾比索
+    Uruguayan peso: 우루과이 페소 #乌拉圭比索
+    Peso: 페소 #比索
+    Centavo: 仙|菲辅币
+#Pound
+    Alderney pound: 奥尔德尼镑
+    British pound: 영국 파운드 #英镑
+    Guernsey pound: 건지섬 파운드 #https://ko.wikipedia.org/wiki/%EA%B1%B4%EC%A7%80%EC%84%AC_%ED%8C%8C%EC%9A%B4%EB%93%9C #根西镑
```
tellarin, line 267: What do these commented out cases at the end mean?

Patterns/Korean/Korean-NumbersWithUnit.yaml
```diff
@@ -0,0 +1,781 @@
...
+    - 란드
+    - 엔
+    - 센트
+    - 프랑
+    - 휠던
+#DimensionExtractorConfiguration
+DimensionSuffixList: !dictionary
+  types: [string, string]
+  entries:
+# China 公制
```
tellarin, line 623: ???

Patterns/Korean/Korean-NumbersWithUnit.yaml
```diff
@@ -0,0 +1,781 @@
...
+    Jersey pound: 저지섬 파운드 #https://ko.wikipedia.org/wiki/%EC%A0%80%EC%A7%80%EC%84%AC_%ED%8C%8C%EC%9A%B4%EB%93%9C #泽西岛镑
+    Lebanese pound: 레바논 파운드 #黎巴嫩镑
+    South Sudanese pound: 남수단 파운드 #南苏丹镑
+    Sudanese pound: 수단 파운드 #苏丹镑
+    Syrian pound: 시리아 파운드 #叙利亚镑
+    Pound: 파운드
+    Pence: 펜스
+    Shilling: 先令
+#Dollar
+    United States dollar: US 달러|미국 달러 #美元|美金|美圆
```
tellarin, line 282: In no informal case people use foreign words for currencies or other units?
If informally they show up in text or speech, we should cover them.

Specs/NumberWithUnit/Korean/AgeModel.json
```diff
@@ -1,7 +1,8 @@
 [
   {
     "Input": "그녀가 다섯 살 때, 그녀는 자전거 타는 것을 배웠다.",
-    "NotSupportedByDesign": "dotNet,javascript,python,java,",
+    "NotSupportedByDesign": "javascript,python,java",
+    "IgnoreResolution": "true",
```
tellarin, line 5: For such cases, can you add a "Comment" attribute with the reason? For example, if it's a number failure or some other case?

Specs/NumberWithUnit/Korean/AgeModel.json
```diff
@@ -57,7 +59,8 @@
   },
   {
     "Input": "중국의 만리장성은 500년 이상 되었고, 5,000마일 이상 뻗어 있다.",
-    "NotSupportedByDesign": "dotNet,javascript,python,java,",
+    "NotSupportedByDesign": "javascript,python,java",
+    "NotSupported": "dotnet",
```
tellarin, line 63: Same.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.NumberWithUnit/NumberWithUnitRecognizer.cs
LionbridgeCSII: reordered models

Patterns/Korean/Korean-NumbersWithUnit.yaml
LionbridgeCSII: done

Patterns/Korean/Korean-NumbersWithUnit.yaml
LionbridgeCSII: removed

Patterns/Korean/Korean-NumbersWithUnit.yaml
LionbridgeCSII: removed

Patterns/Korean/Korean-NumbersWithUnit.yaml
LionbridgeCSII: it seems they are not really used. Symbols ($,£,€...) are used but only as prefixes not as suffixes.

Specs/NumberWithUnit/Korean/AgeModel.json
LionbridgeCSII: comments added.

Specs/NumberWithUnit/Korean/AgeModel.json
LionbridgeCSII: added.

LionbridgeCSII:
After reviewing AgeModel more carefully, we decided to keep only the unambiguous age units to avoids conflicts with duration and date, and we modified a couple of inputs to take this into account:
"십 년" (ten years) -> "십 세" (ten years old)
"500년" (500 years) -> "500살" (500 years old)

tellarin:
> After reviewing AgeModel more carefully, we decided to keep only the unambiguous age units to avoid conflicts with duration and date, and we modified a couple of inputs to take this into account:
> "십 년" (ten years) -> "십 세" (ten years old)
> "500년" (500 years) -> "500살" (500 years old)

In this case, perhaps it would be good to keep the ambiguous specs and leave them marked as NotSupportedByDesign with a Comment for the reason. Having such cases in the spec set will be very helpful in future iterations of the project.

LionbridgeCSII:
> In this case, perhaps it would be good to keep the ambiguous specs and leave them marked as NotSupportedByDesign with a Comment for the reason. Having such cases in the spec set will be very helpful in future iterations of the project.

Ok, reintroduced with comment.

tellarin (approved):

Issue #2494, opened 2021-02-09
[DE Unit/Currency] Not recognizing hyphen-connected unit expressions


cahann:
Word sequences are often hyphenated in German, which can prevent currency expressions from being recognized:
•	DE: "12-Pfund-Tasse Kaffee" ("12-pound-cup of coffee") returns only a number

LionbridgeCSII:
looking into this

PR #2544, opened 2021-04-01
[DE Unit/Currency] Support for hyphen-connected unit expressions (#2494)

LionbridgeCSII:
Fix to issue #2494 in .NET. 
Solution applied also in Java, but an additional problem appears there with non-spaced connectors (i.e. "12 - Pfund" is recognized correctly but "12-Pfund" is not).
Test cases added to German DimensionModel and CurrencyModel.

### DIFF
--- a/Specs/NumberWithUnit/German/DimensionModel.json
+++ b/Specs/NumberWithUnit/German/DimensionModel.json
@@ -363,5 +363,41 @@
     "Input": "2:00 pm",
     "NotSupported": "java",
     "Results": []
+  },
+  {
+    "Input": "12 Pfund Tasse Kaffee",
+    "NotSupportedByDesign": "python",
+    "NotSupported": "javascript",
+    "Results": [
+      {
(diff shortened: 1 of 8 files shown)


Specs/NumberWithUnit/German/DimensionModel.json
```diff
@@ -363,5 +363,41 @@
     "Input": "2:00 pm",
     "NotSupported": "java",
     "Results": []
+  },
+  {
+    "Input": "12 Pfund Tasse Kaffee",
```
tellarin, line 368: Can you also add a case for "12 - Pfund"?

tellarin (changes requested):

Specs/NumberWithUnit/German/DimensionModel.json
LionbridgeCSII: added

tellarin (approved):

Issue #2479, opened 2021-02-09
[* Currency] Support for compound Currency entities in languages other than English


cahann:
"15 dollars and 15 cents" is a single currency entity in EN, but the PT and ES counterparts are captured as a pair of currency entities.
EN: 15 dollars and 15 cents  Returns 15.15 Dollar
PT: 15 dólares e 15 centavos  Returns 15 Dólar, 15 Centavo
ES: 15 dólares y 15 centavos  Returns 15 Dólar, 15 Centavo
DE: "25 Dollar 50" => 25 "Dollar"

tellarin:
Currently only English recognizes compound Currency entities. All other languages recognize two entities.
This is not considered a bug, but an extension, which can be implemented after higher priority bugs are fixed.

cahann:
@tellarin how do we track the enhancements? Should i create other tasks?

tellarin:
We can track it here. But currently most enhancements have low priority due to resource limitations.

LionbridgeCSII:
looking into this

PR #2551, opened 2021-04-13
[* Currency] Added support for compound Currency entities in Portuguese and Spanish (#2479)

LionbridgeCSII:
Partial fix to issue #2479. 
Test cases added to PT and ES CurrencyModels.


tellarin (approved):

tellarin:
@LionbridgeCSII as this partially fixes the referenced issue. Can you look into fixing the reported German case too?

LionbridgeCSII:
> @LionbridgeCSII as this partially fixes the referenced issue. Can you look into fixing the reported German case too?

ah, I missed the German case, yes I will address it too.

tellarin:
If you can propagate the fix to French, Dutch, Italian, and Swedish that would be great too! The bug was open as "*". ☺

PR #2575, opened 2021-04-30
[.NET] Japanese CJK configuration + DateExtractor refinements

LionbridgeCSII:
- Refactored Japanese DateTime configuration files to use common CJK implementaion
(Some of the base methods have been modified to take into account the differences in Japanese implementation).
- Deleted files (because not used in the new implementation): JapaneseBaseDateTimeExtractorConfiguration, JapaneseDateTimeParser, JapaneseMergedDateTimeParserConfiguration, FullDateTimeParser, IFullDateTimeParserConfiguration
- Refined Japanese DateExtractor coverage: 213 test cases pass, 1 fail.
Failing case: "2週先の月曜日のご都合はいかがでしょうか" (How about the convenience of Monday two weeks ahead?), the pattern "2週先の月曜日" (Monday two weeks ahead) does not seem to be supported even in English.

Modified test cases:
- Added support for digit variants that look the same as normal digits. New test cases were added for them.
- Test cases for "元旦" and "大晦日" (New year's day and New Year's eve) were removed from DateExtractor and inserted in DateTimeModel. These are supported by HolidayExtractor.
- The test case "旧暦の２０１５年お正月の１日は春節です。" (EN: the first day of the New Year 2015 in the old calendar is Spring Festival) has been moved to DateTimeModel. "旧暦の２０１５年お正月の１日" (EN: the first day of the New Year 2015 in the old calendar) should be extracted by HolidayExtractor.
- The test case "次の十二日" (EN: next 12 days) was moved  to DateTimeModel as a range.
- Added support for Era names. They goes back until "1789 寛政 Kansei". Test cases referring to years in this format (平成13年, 平成二年, 平成元年, 昭和13年, 昭和二年, 昭和元年, 大正13年, 大正13年, 大正二年, 大正元年, 明治13年, 明治二年,明治元年) were copied to DateTimeModel. The original inputs were modified to indicate a proper date in DateExtractor.
- "すみません、６月１５日の朝出発の航空便を予約しました。" (EN: I'm sorry, I booked a flight departing on the morning of June 15th.) was moved to DateTimeModel because it represents a DateTime and not a Date.
- Support for expressions like "this month", "last year"... has been moved to DatePeriod and the relative test cases moved to DatePeriodExtractor ("今月","こんげつ","来月","らいげつ","再来月","昨月","さくげつ","先月","せんげつ","先々月","せんせんげつ","前月","ぜんげつ","今年","ことし","来年","らいねん","再来年","さらいねん","昨年","さくねん","きょねん")
- The resolution of the test case "08月から12月まで" (August to December) in DatePeriodParser has been modified to be in line with similar cases in the Chinese spec.


.NET/Microsoft.Recognizers.Definitions.Common/Japanese/DateTimeDefinitions.cs
```diff
@@ -22,91 +22,95 @@ namespace Microsoft.Recognizers.Definitions.Japanese
...
-      public const string DateDayRegexInJapanese = @"(?<day>初一|三十|一日|十一日|二十一日|三十一日|二日|三日|四日|五日|六日|七日|八日|九日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|一日|十一日|十日|二十一日|二十日|三十一日|三十日|二日|三日|四日|五日|六日|七日|八日|九日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|十日|二十日|三十日|10日|11日|12日|13日|14日|15日|16日|17日|18日|19日|1日|20日|21日|22日|23日|24日|25日|26日|27日|28日|29日|2日|30日|31日|3日|4日|5日|6日|7日|8日|9日|一号|十一号|二十一号|三十一号|二号|三号|四号|五号|六号|七号|八号|九号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|一号|十一号|十号|二十一号|二十号|三十一号|三十号|二号|三号|四号|五号|六号|七号|八号|九号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|十号|二十号|三十号|10号|11号|12号|13号|14号|15号|16号|17号|18号|19号|1号|20号|21号|22号|23号|24号|25号|26号|27号|28号|29号|2号|30号|31号|3号|4号|5号|6号|7号|8号|9号)";
-      public const string DayRegexNumInJapanese = @"(?<day>一|十一|二十一|三十一|二|三|四|五|六|七|八|九|十二|十三|十四|十五|十六|十七|十八|十九|二十二|二十三|二十四|二十五|二十六|二十七|二十八|二十九|一|十一|十|二十一|二十|三十一|三十|二|三|四|五|六|七|八|九|十二|十三|十四|十五|十六|十七|十八|十九|二十二|二十三|二十四|二十五|二十六|二十七|二十八|二十九|十|二十|廿|卅)";
+      public const string DayRegexForPeriod = @"(?<day>[3３][01０１]|[012０１２]?[\d０１２３４５６７８９]|(三十一?|(一|二)?十?[一二三四五六七八九]))(\s*日)?(?=\b|t|まで|から)?";
```
tellarin, line 29: The extra encoding digits should not be needed. All input in the recognizers is normalized before extraction/parsing.

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Extractors/JapaneseTimeExtractorConfiguration.cs
```diff
@@ -6,14 +6,56 @@
...
+
+        public static readonly string MinuteRegex = DateTimeDefinitions.TimeMinuteRegex;
+
+        public static readonly string SecondRegex = DateTimeDefinitions.TimeSecondRegex;
+
+        public static readonly string HalfRegex = DateTimeDefinitions.TimeHalfRegex;
+
+        public static readonly string QuarterRegex = DateTimeDefinitions.TimeQuarterRegex;
+
+        // e.g: 十二五十から八|半分|瞬間
```
tellarin, line 42: Don't forget to update the comments/examples.

Specs/DateTime/Japanese/DatePeriodParser.json
```diff
@@ -4567,7 +4567,7 @@
         "Text": "08月から12月まで",
         "Type": "daterange",
         "Value": {
-          "Timex": "(XXXX-08-01,XXXX-12-01,P4M)",
+          "Timex": "(2018-08-01,2018-12-01,P4M)",
```
tellarin, line 4570: The input doesn't specify the year, so ideally the timex should say XXXX. Is this an issue also in Chinese?
We should open a bug to fix it.

tellarin:
Regarding the "Monday two weeks ahead" case, is that the Japanese equivalent to "Monday two weeks from now"?
If so, we can create an issue to add support in both English and Japanese. In English it currently returns two separate entities that ideally should be merged.

tellarin (approved):

tellarin:
@LionbridgeCSII, I've approved the PR as the issues are not blockers. Please reply to the comments.

.NET/Microsoft.Recognizers.Definitions.Common/Japanese/DateTimeDefinitions.cs
LionbridgeCSII: ah, I was not aware of that.

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Extractors/JapaneseTimeExtractorConfiguration.cs
LionbridgeCSII: yes, I updated them already in both Korean and Japanese config files.

LionbridgeCSII:
> Regarding the "Monday two weeks ahead" case, is that the Japanese equivalent to "Monday two weeks from now"?
> If so, we can create an issue to add support in both English and Japanese. In English it currently returns two separate entities that ideally should be merged.

Ah yes, they seem to be the equivalent. And yes, also in Japanese two separate entities are returned. 
I will create the issue then.

Specs/DateTime/Japanese/DatePeriodParser.json
LionbridgeCSII: yes that is what happens in general, but in the Chinese original parser code the year is assigned in certain cases (like if the reference date is within the range), so I thought it was intended.

LionbridgeCSII: I will open the bug

LionbridgeCSII: @tellarin, but this is also the behavior in English and the other languages. For an input like " from August to December", the timex will specify the year if the reference date is within the range and will use XXXX otherwise. For example in English:
"ReferenceDateTime": "2018-09-01T00:00:00" -----> "Timex": "(2018-08-01,2018-12-01,P4M)"
"ReferenceDateTime": "2018-07-01T00:00:00" -----> "Timex": "(XXXX-08-01,XXXX-12-01,P4M)"

tellarin: Yes, this is a bug that went overlooked. We should fix it.
The design is that a timex should be re-resolvable by passing a different reference datetime. In this case it would not work as the reference year is mistakenly hard-coded into the timex.

LionbridgeCSII: I see, I will open a bug then.

Issue #2573, opened 2021-04-28
[DE DateTimeV2] Holiday coverage in German needs improvement


tellarin:
All German holidays listed in https://www.schulferien.org/deutschland/feiertage/ should be supported.
For the purposes of this issue, only consider holidays that translate to a specific date, not a range.
For example, treating Ostern (Easter) as week is outside the scope of the issue. But dates like Ostersonntag (Easter Sunday) should be supported.

Please also add the specs for 
"Ostern ist dieses Jahr früh" ("Easter is early this year")
with Comment attributes explaining the current output for this case.

Lastly, it seems the current resolution for Allerheiligen is incorrect.

tellarin:
@LionbridgeCSII, can you please prioritize this one? Thanks.

tellarin:
Austrian and Swiss holidays not matching the German definitions are outside the scope of this issue and are currently lower priority.

tellarin:
DateTimes in formats like "So 27.02.2022" and "in 15 Tg." should also be supported. If already supported, please include corresponding specs. Ideally with usage in full sentences.

PR #2581, opened 2021-05-04
[DE DateTimeV2] Improvements to Holiday coverage in German (#2573)

LionbridgeCSII:
Fix to issue #2573.
Added support (and test cases) for the holidays in the list https://www.schulferien.org/deutschland/feiertage/ not already covered by the German Recognizer.

### DIFF
--- a/Specs/DateTime/German/DateTimeModel.json
+++ b/Specs/DateTime/German/DateTimeModel.json
@@ -3616,5 +3616,392 @@
         }
       }
     ]
+  },
+  {
+    "Input": "Ich besuche dich an Palmsonntag 2019" ,
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupported": "java, javascript, python",
+    "Results": [
+      {
+        "Text": "palmsonntag 2019",
+        "Start": 20,
+        "End": 35,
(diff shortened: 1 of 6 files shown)


Specs/DateTime/German/DateTimeModel.json
```diff
@@ -3616,5 +3616,392 @@
...
+              "type": "date",
+              "value": "2019-11-01"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Ostern ist dieses Jahr früh" ,
```
tellarin, line 3915: Is "early this year" or "earlier this year" supported in English?

tellarin (approved):

Specs/DateTime/German/DateTimeModel.json
LionbridgeCSII: yes both are supported.

tellarin: Then please open an issue to also support it in German. As this spec shows, the "early/earlier" part doesn't seem to be supported along with "this year". Thanks.

LionbridgeCSII: ok, will do.

PR #2592, opened 2021-05-12
[JavaScript | * DateTimeV2] Added missing filter methods (#2509)

LionbridgeCSII:
Fix to issue #2509.


tellarin (approved):
Thanks! But let's keep the non-dotnet issues as lower priority currently. Unless flagged otherwise.

LionbridgeCSII:
> Thanks! But let's keep the non-dotnet issues as lower priority currently. Unless flagged otherwise.

Yes, I have been working on this when there were no .net issues available.

PR #2595, opened 2021-05-15
Simplify Code

iamcarbon:
This PR:

- Removes private setters from immutable properties
- Enables C# 9
- Uses the is pattern to match constant strings
- Removes unused using statements
- Prefer is null to == null


iamcarbon:
Ready for review.

Issue #2604, opened 2021-05-18
[FR, IT DateTimeV2] dates with slashes resolved incorrectly following a weekday


srandal:
**Describe the bug**
When a date written with forward slashes follows a weekday (e.g. Samedi 23/06/18)
The resolved value ignores the month and year.

**To Reproduce**
1. Create a fr-fr,  fr-ca, or it-it app with datetimeV2
2. Test "Je serai de retour le samedi 23/06/18 " (or "Tornerò il Sabato 23/06/18" for it-it)
3. Both languages will ignore the specified year and month.
    a. For FR, the timex will be "XXXX-WXX-6", resolving to "2021-01-23" and "2021-10-23" (Saturday the 23rd in the current year)
    b. For IT, the timex will be the day date with the current year and month (e.g. "2021-05-23")

**Expected behavior**
The timex should use the month and year info, and resolve to "2018-06-23"

**Platform (please complete the following information):**
 - Platform: 
 - Environment: in browser
 - Version of package: ?

**Additional context**
This is somewhat of an edge case, since in any language "Saturday the 23rd" is more common than "Saturday, May 23rd, 2018" but it does work as expected in other languages.

tellarin:
French was correct. Italian has been fixed.

PR #2612, opened 2021-05-20
[FR, IT DateTimeV2] Fixed dates with slashes resolved incorrectly following a weekday (#2604)

LionbridgeCSII:
Fix to issue #2604 in Italian (the pattern was already working as expected in French).

Test case added to French and Italian DateTimeModel.


tellarin (approved):

tellarin:
@LionbridgeCSII, is this form also supported in the other European languages? If not, please add support after the current issues you're working on.
Thanks.

Issue #2605, opened 2021-05-18
[PT, FR DateTimeV2] Incorrectly splitting weekday from date


srandal:
**Describe the bug**
When a weekday and date appear together, they are returned as separate dates, e.g. "Mon prochain salaire sera le 25 Mardi"

**To Reproduce**
1. Create an FR or PT app with datetimeV2
2. Test "Mon prochain salaire sera le 25 Mardi" for FR or "meu próximo salário será terça-feira dia 25" for PT
3. The result will be two timex values: "XXXX-WXX-2" and "XXXX-XX-25"

**Expected behavior**
The result should be a single date timex (e.g. for ES, "el martes 25" returns the timex "XXXX-WXX-2" and it resolves to "2020-08-25" and "2021-05-25", which are both Tuesdays).

**Platform (please complete the following information):**
 - Platform: 
 - Environment: in browser
 - Version of package: ?

tellarin:
"meu próximo salário será terça-feira 25" should also be supproted.

srandal:
The test case above succeeds when September 2018 is set as the reference date, because in that month the 25th fell on a Tuesday. When the reference date is a month that doesn't contain the specified date/weekday combination, the weekday-date split still happens. A weekday-date combination that doesn't occur in the current month is somewhat of an edge case though.

PR #2613, opened 2021-05-20
[PT, FR DateTimeV2] Fixed incorrectly splitting weekday from date (#2605)

LionbridgeCSII:
Fix to issue #2605.
Modified handling of WeekDayAndDayOfMonthRegex in NumberWithMonth to allow extraction of patterns where the day number precedes the weekday like in the french case "le 25 Mardi".

Test cases added to French and Portuguese DateTimeModel.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Recognizers.Definitions.French
       public const string ThisPrefixRegex = @"(?<order>ce(tte)?|au\s+cours+(du|de))\b";
       public const string RangePrefixRegex = @"(du|depuis|des?|entre)";
       public const string DayRegex = @"(?<day>(?:3[0-1]|[1-2]\d|0?[1-9])(e(r)?)?)(?=\b|t)";
-      public const string WrittenDayRegex = @"(?<day>un|deux|trois|quatre|cinq|six|sept|huit|neuf|dix|onze|douze|treize|quatorze|quinze|seize|dix-sept|dix-huit|dix-neuf|vingt-et-un|vingt et un|vingt|vingt deux|vingt-deux|vingt trois|vingt-trois|vingt quatre|vingt-quatre|vingt cinq|vingt-cinq|vingt six|vingt-six|vingt sept|vingt-sept|vingt huit|vingt-huit|vingt neuf|vingt-neuf|trente|trente et un|trente-et-un)";
+      public const string WrittenDayRegex = @"(?<day>((vingt|trente)(\s*-\s*|\s+)et(\s*-\s*|\s+))?un|(vingt(\s*-\s*|\s+))?(deux|trois|quatre|cinq|six|sept|huit|neuf)|dix|onze|douze|treize|quatorze|quinze|seize|dix-(sept|huit|neuf)|vingt|trente)";
       public const string MonthNumRegex = @"(?<month>1[0-2]|(0)?[1-9])\b";
       public const string SpecialDescRegex = @"(p\b)";
       public static readonly string AmDescRegex = $@"(h\b|{BaseDateTime.BaseAmDescRegex})";
--- a/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Recognizers.Definitions.French
       public const string ThisPrefixRegex = @"(?<order>ce(tte)?|au\s+cours+(du|de))\b";
       public const string RangePrefixRegex = @"(du|depuis|des?|entre)";
       public const string DayRegex = @"(?<day>(?:3[0-1]|[1-2]\d|0?[1-9])(e(r)?)?)(?=\b|t)";
+      public const string WrittenDayRegex = @"(?<day>un|deux|trois|quatre|cinq|six|sept|huit|neuf|dix|onze|douze|treize|quatorze|quinze|seize|dix-sept|dix-huit|dix-neuf|vingt-et-un|vingt et un|vingt|vingt deux|vingt-deux|vingt trois|vingt-trois|vingt quatre|vingt-quatre|vingt cinq|vingt-cinq|vingt six|vingt-six|vingt sept|vingt-sept|vingt huit|vingt-huit|vingt neuf|vingt-neuf|trente|trente et un|trente-et-un)";
       public const string MonthNumRegex = @"(?<month>1[0-2]|(0)?[1-9])\b";
       public const string SpecialDescRegex = @"(p\b)";
       public static readonly string AmDescRegex = $@"(h\b|{BaseDateTime.BaseAmDescRegex})";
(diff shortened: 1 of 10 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -32,6 +32,7 @@ public static class DateTimeDefinitions
       public const string ThisPrefixRegex = @"(?<order>ce(tte)?|au\s+cours+(du|de))\b";
       public const string RangePrefixRegex = @"(du|depuis|des?|entre)";
       public const string DayRegex = @"(?<day>(?:3[0-1]|[1-2]\d|0?[1-9])(e(r)?)?)(?=\b|t)";
+      public const string WrittenDayRegex = @"(?<day>un|deux|trois|quatre|cinq|six|sept|huit|neuf|dix|onze|douze|treize|quatorze|quinze|seize|dix-sept|dix-huit|dix-neuf|vingt-et-un|vingt et un|vingt|vingt deux|vingt-deux|vingt trois|vingt-trois|vingt quatre|vingt-quatre|vingt cinq|vingt-cinq|vingt six|vingt-six|vingt sept|vingt-sept|vingt huit|vingt-huit|vingt neuf|vingt-neuf|trente|trente et un|trente-et-un)";
```
tellarin, line 35: This regex can be optimized a little to avoid repetitions and backtracking. This can come in a separate PR.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/DateTimeDefinitions.cs
```diff
@@ -26,6 +26,7 @@ public static class DateTimeDefinitions
       public const string TillRegex = @"(?<till>\b(at[eé]h?|[aà]s|ao?)\b|--|-|—|——)(\s+\b(o|[aà](s)?)\b)?";
       public static readonly string RangeConnectorRegex = $@"(?<and>(e\s*(([àa]s?)|o)?)|{BaseDateTime.RangeConnectorSymbolRegex})";
       public const string DayRegex = @"(?<day>(?:3[0-1]|[1-2]\d|0?[1-9]))(?=\b|t)";
+      public const string WrittenDayRegex = @"(?<day>um|dois|tres|três|quatro|cinco|seis|sete|oito|nove|dez|onze|doze|dezena|dezenas|treze|catorze|quatorze|quinze|dezesseis|dezasseis|dezessete|dezassete|dezoito|dezenove|dezanove|vinte(\s+e\s+(um|dois|duas|tres|três|quatro|cinco|seis|sete|oito|nove))?|trinta(\s+e\s+um)?)";
```
tellarin, line 29: Less repetition than French, but this could also be optimized. Low-priority.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/DateTimeDefinitions.cs
```diff
@@ -86,7 +87,8 @@ public static class DateTimeDefinitions
       public const string SpecialDayRegex = @"\b((d?o\s+)?(dia\s+antes\s+de\s+ontem|antes\s+de\s+ontem|anteontem)|((d?o\s+)?(dia\s+|depois\s+|dia\s+depois\s+)?de\s+amanh[aã])|(o\s)?dia\s+seguinte|(o\s)?pr[oó]ximo\s+dia|(o\s+)?[uú]ltimo\s+dia|ontem|amanh[ãa]|hoje)|(do\s+dia$)\b";
       public const string SpecialDayWithNumRegex = @"^[.]";
       public const string ForTheRegex = @".^";
-      public const string WeekDayAndDayOfMonthRegex = @".^";
+      public static readonly string FlexibleDayRegex = $@"(?<DayOfMonth>([A-Za-z]+\s)?({WrittenDayRegex}|{DayRegex}))";
```
tellarin, line 90: No need to specify "A-Z" as for perf pre-processing guarantees capital letters won't reach the regex.

Specs/DateTime/Portuguese/DateTimeModel.json
```diff
@@ -1825,5 +1825,82 @@
...
+    "Results": [
+      {
+        "Text": "terça-feira 25",
+        "Start": 25,
+        "End": 38,
+        "TypeName": "datetimeV2.date",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "XXXX-WXX-2",
```
tellarin, line 1892: While this PR solves extraction and resolution for the issue, it seems we have a timex issue to refine.
1- This timex is consistent with the example given in the issue for Spanish. Why does "terça-feira dia 25" return differently (also the French cases)? They should be consistent, as they are the same date reference.
If possible, the timex generation and resolution code should be re-used to avoid such inconsistencies. Please compare to the behaviour in English.
2- Once consistency is corrected, we may also consider if the day and day-of-week should both be added in the generalized timex (something like XXXX-XX-25-2). We'll take a closer look and open an issue as appropriate. This is outside the scope of this issue and PR.

tellarin (changes requested):

Specs/DateTime/Portuguese/DateTimeModel.json
LionbridgeCSII: This happens also in English for the patterns "Tuesday the 25th" and "Tuesday 25", the first is extracted by WeekDayAndDayOfMothRegex and the second by WeekDayAndDayRegex and they follow different parsing steps which produce different results.

LionbridgeCSII: In Spanish, "martes 25" is processed as "terça-feira 25"/"Tuesday 25". Instead, "martes el dia 25" is wrongly parsed as ForTheRegex instead of WeekDayAndDayOfMothRegex. This need to be corrected too.
So, in this PR, I will correct this Spanish behaviour and modify the timex returned by WeekDayAndDayOfMothRegex to be in line with that returned by WeekDayAndDayRegex.

tellarin: Makes sense. Please correct the Spanish behaviour to match the others. Thanks.

tellarin: We'll analyze the unification of regex forms and potential creation of a new form for this case and create a new issue, if necessary.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
LionbridgeCSII: modified

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/DateTimeDefinitions.cs
LionbridgeCSII: modified

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/DateTimeDefinitions.cs
LionbridgeCSII: removed

Specs/DateTime/Portuguese/DateTimeModel.json
LionbridgeCSII: conformed Spanish behavior to match the others

tellarin (approved):

PR #2618, opened 2021-05-25
[.NET] Korean Duration support

LionbridgeCSII:
Extractor: 34 pass, 0 fail.
Parser: 22 pass, 13 only extraction, 0 fail.

Ported method ImplicitDuration from BaseDateExtractor to support cases like "all day", "half day", "next day"... not covered in CJK implementation.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDurationParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDurationParser.cs
@@ -38,10 +38,10 @@ namespace Microsoft.Recognizers.Text.DateTime
                 }
 
                 var unitStr = unitResult.Unit;
-                var numStr = unitResult.Number;
+                var number = string.IsNullOrEmpty(unitResult.Number) ? 1 : double.Parse(unitResult.Number);
 
-                dateTimeParseResult.Timex = TimexUtility.GenerateDurationTimex(double.Parse(numStr), unitStr, BaseDurationParser.IsLessThanDay(unitStr));
-                dateTimeParseResult.FutureValue = dateTimeParseResult.PastValue = double.Parse(numStr) * this.config.UnitValueMap[unitStr];
+                dateTimeParseResult.Timex = TimexUtility.GenerateDurationTimex(number, unitStr, BaseDurationParser.IsLessThanDay(unitStr));
+                dateTimeParseResult.FutureValue = dateTimeParseResult.PastValue = number * this.config.UnitValueMap[unitStr];
                 dateTimeParseResult.Success = true;
             }
 
@@ -124,7 +124,7 @@ namespace Microsoft.Recognizers.Text.DateTime
                 if (unitMatch.Success)
                 {
                     var pr = (DateTimeParseResult)Parse(er);
-                    if (pr.Value != null)
+                    if (pr != null && pr.Value != null)
                     {
                         timexDict.Add(this.config.UnitMap[unitMatch.Groups["unit"].Value], pr.TimexStr);
                         prs.Add(pr);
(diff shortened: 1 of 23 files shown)


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -144,6 +144,12 @@ protected ParseResult ParseFraction(ExtractResult extResult)
                 demoPart = splitResult[1];
                 numPart = splitResult[2];
             }
+            else if (splitResult.Length == 1)
+            {
+                intPart = Config.ZeroChar.ToString(CultureInfo.InvariantCulture);
```
tellarin, line 149: What is this for? Can you add a code comment here with an example?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDurationParser.cs
```diff
@@ -38,10 +38,10 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
                 }
 
                 var unitStr = unitResult.Unit;
-                var numStr = unitResult.Number;
+                var number = string.IsNullOrEmpty(unitResult.Number) ? 1 : double.Parse(unitResult.Number);
```
tellarin, line 41: On all number parsing, please specify the invariant culture.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Korean/Extractors/KoreanNumberWithUnitExtractorConfiguration.cs
```diff
@@ -94,7 +94,9 @@ public void ExpandHalfSuffix(string source, ref List<ExtractResult> result, IOrd
                         var match_suffix = new List<ExtractResult>();
                         foreach (var mr in match)
                         {
-                            if (mr.Start == (start + length))
+                            var subLength = (int)mr.Start - (start + length) >= 0 ? (int)mr.Start - (start + length) : 0;
```
tellarin, line 97: Can you add a code comment here to explain this behaviour difference?

Specs/DateTime/Korean/DurationParser.json
```diff
@@ -569,21 +551,21 @@
...
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "NotSupported": "dotnet",
+    "IgnoreResolution": "true",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
-        "Text": "일주일 그리고 사흘",
+        "Text": "일주일 그리고 사흘 더",
```
tellarin, line 568: Doesn't "더" mean more?
If so, why should it also be tagged along with the duration? Currently we don't have duration ranges and modifiers in durations.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
LionbridgeCSII: comment added (needed to parse 'half' in Korean and Japanese).

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDurationParser.cs
LionbridgeCSII: ok, done.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Korean/Extractors/KoreanNumberWithUnitExtractorConfiguration.cs
LionbridgeCSII: added (needed to take into account whitespaces between result and half unit).

Specs/DateTime/Korean/DurationParser.json
LionbridgeCSII: The English counterpart is simply "I'll leave for one week three days", so we added a new test case without "more" to better match this input: "나는 **일주일 그리고 사흘** 동안 떠날 겁니다". 
But "더" appears to be commonly used in other test cases in expressions like "한 시간 더" meaning "one more hour" (corresponding to the English duration pattern "another hour"), so it needs to be supported (in English there is also a test case where "more than an hour" is extracted).

tellarin (approved):

Issue #2478, opened 2021-02-09
[ES/FR DateTimeV2] "ago" not recognized correctly


cahann:
EN: 30 minutes ago  Returns a datetime based on the current clock
PT:  30 minutos atrás (same as EN)
ES: hace 30 minutos  Returns a duration: PT30M

FR: "Je l'ai commandé il y a 24 heures" ("I ordered it 24 hours ago") returns T00 regardless of the current time

PR #2621, opened 2021-05-27
[ES/FR DateTimeV2] Fixed "ago" not recognized correctly (#2478)

LionbridgeCSII:
Fix to issue #2478.

Test cases added to French and Spanish DateTimeModels.

Also replaced undefined regexes ^[.] with ^\b$ in French yaml file because ^[.] was matching in certain cases and giving issues (for example, here "30 minutes." was extracted as "30 minutes in the future" because FutureSuffixRegex was undefined but matched in the afterStr "." of the duration extraction "30 minutes").


Specs/DateTime/French/DateTimeModel.json
```diff
@@ -20902,5 +20902,53 @@
...
+              "type": "datetime",
+              "value": "2021-05-20 15:00:00"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Je l'ai commandé il y a 24 heures.",
```
tellarin, line 20931: Can "il ya" appear in informal French? If so, we should support it if it doesn't cause undesired side effects.
Should we also have more specs with depuis?
And lastly, can you this spec? "Il y a à peu près un mois".
If not supported and too complex to address, add this new spec as disabled and open an issue to fix it. Thanks.

tellarin (changes requested):

Specs/DateTime/French/DateTimeModel.json
LionbridgeCSII: test cases added.
"Il y a à peu près un mois" is now parsed as date but the modifier "à peu près" is ignored. 
The problem is that Merged looks for modifiers only at the beginning/end of the extraction, so a larger modification is required to target modifiers within the extraction.

tellarin: Got it. Let's open an issue for it, but this is lower priority for now.

tellarin (commented):
Approved.

Issue #2625, opened 2021-06-01
[EN DateTimeV2] "in the daytime" extracted but not resolved


LionbridgeCSII:
"in the daytime" is extracted as timerange but Resolution is null.
Instead, "daytime" is extracted and resolved correctly with timex "TDT".

**Expected behavior**
"in the daytime" should be resolved as "daytime".
Moreover, "nighttime" and "in the nighttime" should also be supported (currently only "night" is extracted as part of day).

LionbridgeCSII:
working on this

PR #2629, opened 2021-06-02
[EN DateTimeV2] Fixed "in the daytime" extracted but not resolved (#2625)

LionbridgeCSII:
Fix to issue #2625.

"Nighttime" is now resolved to the range "00:00-08:00". 
Should it instead be resolved as "night" ("20:00-00:00")? Or, by symmetry with "daytime" ("08:00-18:00"), as "20:00-06:00"? (However, this last option would require to move the pattern from TimePeriod to DateTimePeriod)


tellarin (approved):
For now let's kee the nighttime resolution as you've added. This should be changed in a bigger resolver refactoring.

Issue #2628, opened 2021-06-02
[* DateTimeV2] Duration entities with modifiers should not always be merged


LionbridgeCSII:
In certain cases, Duration entities are merged when they should instead stay separate. For example, the following inputs:
"The project was submitted last month and 3 weeks later it was approved" -> "month and 3 weeks later";
"They worked on the problem for the last 3 months and 2 days ago they found a solution" -> "3 months and 2 days ago".

**Expected behavior**
When two duration entities are both preceded/followed by modifiers (e.g. "this", "last", "next", "ago"...), they should not be merged. For example:
"The project was submitted last month and 3 weeks later it was approved" -> "last month", "3 weeks later";
"They worked on the problem for the last 3 months and 2 days ago they found a solution" -> "last 3 months", "2 days ago".

LionbridgeCSII:
working on this

PR #2631, opened 2021-06-03
[* DateTimeV2] Fixed duration entities with modifiers should not always be merged (#2628)

LionbridgeCSII:
Fix to issue #2628.

Test cases added to English DateTimeModel.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
@@ -220,26 +220,11 @@ namespace Microsoft.Recognizers.Text.DateTime
                 return extractorResults;
             }
 
-            // If the first and last extractions are both preceded/followed by modifiers,
-            // they should not be merged, e.g. "last 2 weeks and 3 days ago"
-            var minStart = extractorResults.Min(er => er.Start);
-            var beforeStr = text.Substring(0, (int)minStart);
-            var beforeMod = this.config.ModPrefixRegex.MatchEnd(beforeStr, trim: true);
-            if (beforeMod.Success)
-            {
-                var maxEnd = extractorResults.Max(er => er.Start + er.Length);
-                var afterStr = text.Substring((int)maxEnd);
-                var afterMod = this.config.ModSuffixRegex.MatchBegin(afterStr, trim: true);
-                if (afterMod.Success)
-                {
-                    return extractorResults;
-                }
-            }
-
             var unitMap = this.config.UnitMap;
             var unitValueMap = this.config.UnitValueMap;
             var unitRegex = this.config.DurationUnitRegex;
             List<ExtractResult> ret = new List<ExtractResult>();
+            List<List<ExtractResult>> separaRet = new List<List<ExtractResult>>();
 
             var firstExtractionIndex = 0;
             var timeUnit = 0;
@@ -264,6 +249,9 @@ namespace Microsoft.Recognizers.Text.DateTime
                     firstExtractionIndex++;
                     continue;
                 }
+
+                // Add extraction to list of separate results (needed in case the extractions should not be merged)
+                List<ExtractResult> separaList = new List<ExtractResult>() { extractorResults[firstExtractionIndex] };
 
                 var secondExtractionIndex = firstExtractionIndex + 1;
                 while (secondExtractionIndex < extractorResults.Count)
@@ -301,6 +289,9 @@ namespace Microsoft.Recognizers.Text.DateTime
                         break;
                     }
 
+                    // Add extraction to list of separate results (needed in case the extractions should not be merged)
+                    separaList.Add(extractorResults[secondExtractionIndex]);
+
                     secondExtractionIndex++;
                 }
 
@@ -335,10 +326,35 @@ namespace Microsoft.Recognizers.Text.DateTime
                     ret.Add(extractorResults[firstExtractionIndex]);
                 }
 
+                // Add list of separate extractions to separaRet, so that there is a 1 to 1 correspondence
+                // between ret (list of merged extractions) and separaRet (list of unmerged extractions)
+                separaRet.Add(separaList);
+
                 firstExtractionIndex = secondExtractionIndex;
             }
 
-            return ret;
+            // If the first and last elements of a group of contiguous extractions are both preceded/followed by modifiers,
+            // they should not be merged, e.g. "last 2 weeks and 3 days ago"
+            List<ExtractResult> newRet = new List<ExtractResult>();
+            for (int i = 0; i < ret.Count; i++)
+            {
+                var start = (int)ret[i].Start;
+                var end = start + (int)ret[i].Length;
+                var beforeStr = text.Substring(0, start);
+                var afterStr = text.Substring(end);
+                var beforeMod = this.config.ModPrefixRegex.MatchEnd(beforeStr, trim: true);
+                var afterMod = this.config.ModSuffixRegex.MatchBegin(afterStr, trim: true);
+                if (beforeMod.Success && afterMod.Success)
+                {
+                    newRet.AddRange(separaRet[i]);
+                }
+                else
+                {
+                    newRet.Add(ret[i]);
+                }
+            }
+
+            return newRet;
         }
     }
 }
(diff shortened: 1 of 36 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -220,6 +220,22 @@ private List<ExtractResult> MergeMultipleDuration(string text, List<ExtractResul
                 return extractorResults;
             }
 
+            // If the first and last extractions are both preceded/followed by modifiers,
+            // they should not be merged, e.g. "last 2 weeks and 3 days ago"
+            var minStart = extractorResults.Min(er => er.Start);
```
tellarin, line 225: What happens if there are more than two extracted entities here? Can this happen?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -20627,5 +20627,85 @@
         }
       }
     ]
+  },
+  {
+    "Input": "The project was submitted last month and 3 weeks later it was approved",
```
tellarin, line 20632: Can you add the same/similar specs in other languages too?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
LionbridgeCSII: yes it can happens, there are test cases where more than 2 durations are merged. Now, if the first and last are preceded by modifiers, all the extractions are returned separately. 
I considered merging the intermediate extractions, but it is hard to decide which extractions should be merged. 
For example: "They worked on the problem for the last 3 months and 2 weeks and 2 days ago they found a solution",
it is not clear if "2 weeks" should be merged with "last 3 months" or with "2 days ago". 
But I will try to modify the code to allow merging when the units are not contiguous, e.g. "For the last 3 months, they worked on the problem and 2 weeks and 2 days ago they found a solution", in this case "2 weeks and 2 days ago" can be merged unambiguously.

tellarin: Thanks. Please add a spec example of the 3 entities not merged and one for the non-contiguous case.

LionbridgeCSII: modified and test cases added

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: added in German, Italian and Spanish

tellarin:
@LionbridgeCSII, can you resolve the merge conflicts to continue the review? Thanks.

LionbridgeCSII:
> @LionbridgeCSII, can you resolve the merge conflicts to continue the review? Thanks.

conflicts solved

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -335,10 +326,35 @@ private List<ExtractResult> MergeMultipleDuration(string text, List<ExtractResul
...
+                // between ret (list of merged extractions) and separaRet (list of unmerged extractions)
+                separaRet.Add(separaList);
+
                 firstExtractionIndex = secondExtractionIndex;
             }
 
-            return ret;
+            // If the first and last elements of a group of contiguous extractions are both preceded/followed by modifiers,
+            // they should not be merged, e.g. "last 2 weeks and 3 days ago"
+            List<ExtractResult> newRet = new List<ExtractResult>();
```
tellarin, line 338: Do we really need so many lists? Can't we re-use some, or avoid creating others in certain cases?
There is already too much GC pressure.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -220,26 +220,11 @@ private List<ExtractResult> MergeMultipleDuration(string text, List<ExtractResul
...
-                if (afterMod.Success)
-                {
-                    return extractorResults;
-                }
-            }
-
             var unitMap = this.config.UnitMap;
             var unitValueMap = this.config.UnitValueMap;
             var unitRegex = this.config.DurationUnitRegex;
             List<ExtractResult> ret = new List<ExtractResult>();
```
tellarin, line 226: Let's rename "ret" and "separateRet" for clarity. Maybe "mergedResults" and "separateResults"?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
```diff
@@ -265,6 +250,9 @@ private List<ExtractResult> MergeMultipleDuration(string text, List<ExtractResul
                     continue;
                 }
 
+                // Add extraction to list of separate results (needed in case the extractions should not be merged)
+                List<ExtractResult> separaList = new List<ExtractResult>() { extractorResults[firstExtractionIndex] };
```
tellarin, line 254: This will have at most two elements, right? Couldn't this auxiliary list be replaced by two flags or a simple struct?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
tellarin, line 338: If still using this list object, rename to something like "finalResults".

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -20888,5 +20888,125 @@
         }
       }
     ]
+  },
+  {
+    "Input": "They worked on the problem for the last 3 months and 2 weeks and 2 days ago they found a solution",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T16:12:00"
+    },
+    "NotSupportedByDesign": "javascript,python,java",
```
tellarin, line 20897: Add a Comment attribute here describing that the middle entity is not merged due to ambiguity.

tellarin: Same in the equivalent cases in other languages.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
LionbridgeCSII: ah maybe instead of newRet, one could insert the unmerged results from separaRet directly into ret replacing the merged ones. I will check.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
LionbridgeCSII: no, in cases where there are 3 or more contiguous duration entities, they are all added to this list.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
LionbridgeCSII: removed newRet

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
LionbridgeCSII: renamed as 'results' and 'separateResults' (since 'results' at the end will contain both merged and unmerged results)

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseDurationExtractor.cs
LionbridgeCSII: removed

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: added comment (and test case in Italian and Spanish)

tellarin (approved):

Issue #2632, opened 2021-06-04
[* DateTimeV2] Misinterpretation of a relative duration if prefixed by a number


tellarin:
It seems the recognizers are misinterpreting relative durations prefixed by numbers.

Cases like:
"etfs with negative flows of at least $1k last month" currently return a range of 1000 months instead of just last month.
It seems to interepret it as "the 1000 last months".

tellarin:
@LionbridgeCSII, can you prioritize this? Thanks.

PR #2634, opened 2021-06-07
[* DateTimeV2] Fixed misinterpretation of a relative duration if prefixed by a number (#2632)

LionbridgeCSII:
Fix to issue #2632.

Test cases added to English DateTimeModel.


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -20627,5 +20627,80 @@
...
+              "start": "2016-10-01",
+              "end": "2016-11-01"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "We will be only 3 next week.",
```
tellarin, line 20657: Can you add some cases with the plurals too, if they don't exists? Like "We'll be gone for only the 3 next weeks".
Please add the corrected support specs in a couple other languages too.

tellarin (changes requested):
Looks good. Just some more specs pending. Thanks!

Specs/DateTime/English/DateTimeModel.json
LionbridgeCSII: added

LionbridgeCSII: also in French, Portuguese and Spanish

tellarin (approved):

Issue #2586, opened 2021-05-07
[* DateTimeV2] Reference year incorrectly assigned to Timex in DatePeriod patterns


LionbridgeCSII:
Currently, for an input like " from August to December", the Timex specifies the year if the reference date is within the range and uses XXXX otherwise. 
For example in English:
"ReferenceDateTime": "2018-09-01T00:00:00" -----> "Timex": "(2018-08-01,2018-12-01,P4M)"
"ReferenceDateTime": "2018-07-01T00:00:00" -----> "Timex": "(XXXX-08-01,XXXX-12-01,P4M)"

**Expected behavior**
Years in Timex should always be XXXX (independently of the reference date) when the extractions do not refer (explicitly or implicitly) to specific years.

LionbridgeCSII:
working on this

PR #2646, opened 2021-06-21
[* DateTimeV2] Fixed reference year incorrectly assigned to Timex in DatePeriod patterns (#2586)

LionbridgeCSII:
Fix to issue #2586.

Updated affected test cases in Chinese, English, Hindi, Japanese, Spanish and Turkish specs.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
@@ -492,7 +492,16 @@ namespace Microsoft.Recognizers.Text.DateTime
                             datePeriodTimexType = DatePeriodTimexType.ByWeek;
                         }
 
-                        ret.Timex = TimexUtility.GenerateDatePeriodTimex(futureBegin, futureEnd, datePeriodTimexType, pastBegin, pastEnd);
+                        // If startResolution and endResolution have fuzzy years, also the combined range timex will use fuzzy years.
+                        if (startResolution.Timex.StartsWith(Constants.TimexFuzzyYear, StringComparison.Ordinal) &&
+                            endResolution.Timex.StartsWith(Constants.TimexFuzzyYear, StringComparison.Ordinal))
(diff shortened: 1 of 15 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -492,7 +492,16 @@ private DateTimeResolutionResult ParseComplexDatePeriod(string text, DateObject
                             datePeriodTimexType = DatePeriodTimexType.ByWeek;
                         }
 
-                        ret.Timex = TimexUtility.GenerateDatePeriodTimex(futureBegin, futureEnd, datePeriodTimexType, pastBegin, pastEnd);
+                        // If startResolution and endResolution have fuzzy years, also the combined range timex will use fuzzy years.
+                        if (startResolution.Timex.StartsWith(Constants.TimexFuzzyYear, StringComparison.Ordinal) &&
```
tellarin, line 496: I'd prefer if these two were a call to a single method that receives the necessary context and returns the appropriate timex. Can you take a stab at it and let me know if it would be too complex to change now?
If it is, please add a code comment in both place to unify later.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
LionbridgeCSII: I see, I can move the if statement to a new method defined in TimexUtility, that should be straightforward to do. 
Or do you mean to unify the two definitions of GenerateDatePeriodTimex in TimexUtility? There is a TODO that ask to do so but I am not sure if it is possible without breaking current behavior.

LionbridgeCSII: updated method

tellarin (approved):

PR #2654, opened 2021-06-26
[.NET] Japanese TimePeriod support

LionbridgeCSII:
Extractor: 56 pass, 0 fail.
Parser: 64 pass, 0 fail.

No input has been modified.

Modified BaseCJKTimePeriodParser to add "start"/"end" mods for cases like "early/late morning" as in standard implementation.
Modified method Handle in TimePeriodFunctions to support patterns like "from 11 to 3 morning" (with desc only in the second term). Originally, only patterns like "from 11 morning to 3" (with desc only in the first term) was supported.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
@@ -82,7 +82,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Japanese
             // Modify time period if "early"/"late" is present
             if (DateTimeDefinitions.EarlyHourTermList.Any(o => trimmedText.EndsWith(o, StringComparison.Ordinal)))
             {
-                endHour = beginHour + 2;
+                endHour = beginHour + Constants.HalfMidDayDurationHourCount;
 
                 // Handling special case: night ends with 23:59.
                 if (endMin == 59)
@@ -93,7 +93,7 @@ namespace Microsoft.Recognizers.Text.DateTime.Japanese
 
             if (DateTimeDefinitions.LateHourTermList.Any(o => trimmedText.EndsWith(o, StringComparison.Ordinal)))
             {
-                beginHour = beginHour + 2;
+                beginHour = beginHour + Constants.HalfMidDayDurationHourCount;
             }
 
             return true;
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
@@ -63,6 +63,10 @@ namespace Microsoft.Recognizers.Text.DateTime.Japanese
             {
                 timeOfDay = Constants.Night;
             }
+            else if (DateTimeDefinitions.BusinessHourTermList.Any(o => trimmedText.EndsWith(o, StringComparison.Ordinal)))
+            {
+                timeOfDay = Constants.BusinessHour;
+            }
             else
             {
                 timex = null;
@@ -74,6 +78,23 @@ namespace Microsoft.Recognizers.Text.DateTime.Japanese
             beginHour = parseResult.BeginHour;
             endHour = parseResult.EndHour;
             endMin = parseResult.EndMin;
+
+            // Modify time period if "early"/"late" is present
+            if (DateTimeDefinitions.EarlyHourTermList.Any(o => trimmedText.EndsWith(o, StringComparison.Ordinal)))
+            {
+                endHour = beginHour + 2;
+
+                // Handling special case: night ends with 23:59.
+                if (endMin == 59)
+                {
+                    endMin = 0;
+                }
+            }
+
+            if (DateTimeDefinitions.LateHourTermList.Any(o => trimmedText.EndsWith(o, StringComparison.Ordinal)))
+            {
+                beginHour = beginHour + 2;
+            }
 
             return true;
         }
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKTimePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKTimePeriodParser.cs
@@ -107,12 +107,12 @@ namespace Microsoft.Recognizers.Text.DateTime
             }
 
             // Add "early"/"late" Mod
-            if (endHour == beginHour + 2 && (beginHour == Constants.HalfDayHourCount - 4 || beginHour == Constants.HalfDayHourCount))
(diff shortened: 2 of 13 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
```diff
@@ -75,6 +79,23 @@ public bool GetMatchedTimexRange(string text, out string timex, out int beginHou
             endHour = parseResult.EndHour;
             endMin = parseResult.EndMin;
 
+            // Modify time period if "early"/"late" is present
+            if (DateTimeDefinitions.EarlyHourTermList.Any(o => trimmedText.EndsWith(o, StringComparison.Ordinal)))
```
tellarin, line 83: This logic should be the same across all languages. Please move to a common method that receives what the TOD is and returns the range.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKTimePeriodParser.cs
```diff
@@ -106,10 +106,22 @@ private DateTimeResolutionResult ParseTimeOfDay(string text, DateObject referenc
                 return new DateTimeResolutionResult();
             }
 
+            // Add "early"/"late" Mod
+            if (endHour == beginHour + 2 && (beginHour == Constants.HalfDayHourCount - 4 || beginHour == Constants.HalfDayHourCount))
```
tellarin, line 110: This should also be a common method. When extracting it, please also define constants for '2' and '4'.

Specs/DateTime/Japanese/TimePeriodParser.json
```diff
@@ -789,14 +765,13 @@
...
     },
-    "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
         "Text": "午前11時から11時50分まで",
         "Type": "timerange",
         "Value": {
-          "Timex": "(T11,T11:50,PT50M)",
+          "Timex": "(T11,T11:50,PT0H50M)",
```
tellarin, line 774: This is technically correct. But the previous timex is preferrable.

Specs/DateTime/Japanese/TimePeriodParser.json
```diff
@@ -870,14 +843,13 @@
...
     },
-    "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
         "Text": "午後3時から午後3時30分まで",
         "Type": "timerange",
         "Value": {
-          "Timex": "(T15,T15:30,PT30M)",
+          "Timex": "(T15,T15:30,PT0H30M)",
```
tellarin, line 852: Same, and so on.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
LionbridgeCSII: But this is handled in a different way in standard implementation (looking for 'early'/'late' named groups in the match and modifying the begin/end hours accordingly). 
It has been implemented differently in Japanese because, for example, 'early morning' is not always the combination of 'early' and 'morning' but can be a different word.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKTimePeriodParser.cs
LionbridgeCSII: also here the standard implementation follows different implementation as described above

Specs/DateTime/Japanese/TimePeriodParser.json
LionbridgeCSII: we were updating this, but since it breaks many cases in Chinese we decided to follow Chinese approach. 
Should we update Chinese test cases to follow standard format (without '0H')?

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
tellarin: I meant the mapping between the abstract definition of "morning" or "early morning" into an actual range.
Now this type of code is spread around all languaes. A centralized location should receive the relevant abstracted signals/parameters, not language specific terms, and return a struct/tuple with the range results.
We can start with refactoring CJK, then do the same for the other languages. This will also help in the future to allow users to define their own customized ranges.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKTimePeriodParser.cs
tellarin: But the same as in other languages, the definitions of what early/late mean or the lengths of the ranges should be uniform.

Specs/DateTime/Japanese/TimePeriodParser.json
tellarin: We should fix Chinese and Korean too. The code that generates the timex should be the same (centralized object/method/...).

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
LionbridgeCSII: this seems to be already accomplished by ParseTimeOfDay in TimexUtility (used by both standard and CJK implementation). 
We moved the begin/end hours defined there to Constants

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKTimePeriodParser.cs
LionbridgeCSII: used value defined in Constants

Specs/DateTime/Japanese/TimePeriodParser.json
LionbridgeCSII: modified and updated specs in Chinese and Japanese

Specs/DateTime/Japanese/TimePeriodParser.json
LionbridgeCSII: done

.NET/Microsoft.Recognizers.Text.DateTime/Japanese/Parsers/JapaneseTimePeriodParserConfiguration.cs
tellarin: I think this is OK in the context of this PR. But there's still a lot of repeated code across culture configs for basically the same thing.

tellarin (approved):

Issue #2685, opened 2021-08-02
First Week of the Year Timex resolves to incorrect End Date


nikhilkandur:
**Describe the bug**
First Week of the Year e.g., 2020-W01, 2021-W02 resolves to incorrect end date

**To Reproduce**
```python
def test_datatypes_resolver_dateRange_first_week_of_the_year():
    today = datetime(2021, 1, 1)
    resolution = TimexResolver.resolve(["2021-W01"], today)
    print(f'Timex:{resolution.values[0].timex}') 
    print(f'Type:{resolution.values[0].type}') 
    print(f'Start:{resolution.values[0].start}') 
    print(f'End:{resolution.values[0].end}')
    print("=======")

    resolution = TimexResolver.resolve(["2020-W01"], today)
    print(f'Timex:{resolution.values[0].timex}') 
    print(f'Type:{resolution.values[0].type}') 
    print(f'Start:{resolution.values[0].start}') 
    print(f'End:{resolution.values[0].end}')
```
**Result**
```bash
Timex:2021-W01
Type:daterange
Start:2020-12-28
End:2020-12-04
=======
Timex:2020-W01
Type:daterange
Start:2019-12-30
End:2019-12-06
```

**Expected behavior**
```bash
Timex:2021-W01
Type:daterange
Start:2020-12-28
End:2021-01-04
=======
Timex:2020-W01
Type:daterange
Start:2019-12-30
End:2020-01-06
```

**Platform:**
 - Platform: Python
 - Environment: console app
 - Version of package: recognizers-text-suite==1.0.2a2, datatypes-timex-expression==1.0.2a2

nikhilkandur:
This line [datatypes_timex_expression/timex_resolver.py#L167](https://github.com/microsoft/Recognizers-Text/blob/master/Python/libraries/datatypes-timex-expression/datatypes_timex_expression/timex_resolver.py#L167) should be of `end` datetime rather than `start' datetime i.e.,

```python
Timex(year=end.year, month=end.month, day_of_month=end.day)
```

nawanas:
I tried the query of the "The first week of 2021" and "The first week of 2020" and I am getting the following date ranges, which seem correct

Timex:2021-W01
Type:daterange
Start:2021-01-04
End:2021-01-11
=======
Timex:2020-W01
Type:daterange
Start:2019-12-30
End:2020-01-06

nikhilkandur:
[...]

PR #2686, opened 2021-08-03
feat(TimexResolver): fixed first week of the year timex resolution

nikhilkandur:
- Timex like '2021-W01', '2020-W01' are now resolved correctly
- Added a test case, `test_datatypes_resolver_dateRange_first_week`


nikhilkandur:
PR for issue #2685

tellarin (changes requested):
This is actually incorrect.
2021-W1 is actually from 2021-01-04 to 2021-01-11. The recognizers use ISO weeks per definition.
The week in the test below is Week 53, 2020.

nikhilkandur:
yeah, this PR only changed the typo error of `start` to `end` and nothing with the logic of first week calculation.  Looking into @nawans comment, I said it here, [2685#issuecomment-892932735](https://github.com/microsoft/Recognizers-Text/issues/2685#issuecomment-892932735). And also thoughts on this comment, [2685#issuecomment-892951931](https://github.com/microsoft/Recognizers-Text/issues/2685#issuecomment-892951931)

nikhilkandur:
And also, I think, we need to revisit quarters calculation as well as the python one may not be following ISO standards.

PR #2699, opened 2021-08-25
Suppurt hebrew number

hershkoy:
Made the changes in the files, but at the moment it is still not working. I tried running the tests for Hebrew, but ava returns an error:

```
Recognizers-Text\JavaScript>npm test test/runner.js --match='*Hebrew*'

> recognizers@1.3.0 test E:\WORK\ML\develop\NLP\Recognizers-Text\JavaScript
> ava "test/runner.js"



  × No tests found in test\runner.js, make sure to import "ava" at the top of your test file

```


tellarin:
@hershkoy, can you remove the duplicate cases in the specs, so the build can go through?

hershkoy:
I removed the duplicates and now I see there are regex syntax errors in the Hebrew files.
I will check that and I will update (it can take some time though)
Thanks for the help

tellarin:
Thanks for keeping pushing this, @hershkoy! I'll try to take a closer look and see if I can help you a bit.
We can start by merging partially the pieces that are already good. Especially the specs.
How many of the total English Number specs have you localized into Hebrew?

hershkoy:
I am currently working only on `Specs\Number\Hebrew\NumberModel.json`. I localized all specs inside it, and currently about half of the cases are failing (mainly the Fraction phrases still require work).

If you want to merge, maybe I should add:
```
 "NotSupported": "javascript, java, python",
```
On the failing tests?

Issue #2694, opened 2021-08-17
[EN DateTimeV2] Incorrect range result resolving of "starting on Sep 2019" + incorrect modifier


substill:
**Describe the bug**
The system cannot correctly resolve "starting on Seq 2019". 
Here is the current result that "start: 2019-10-01" that is not correct. 
{
  "Text": "starting on sep 2019",
  "Start": 0,
  "End": 19,
  "TypeName": "datetimeV2.daterange",
  "Resolution": {
    "values": [
      {
        "timex": "2019-09",
        "Mod": "after",
        "type": "daterange",
        "sourceEntity": "datetimerange",
        "start": "2019-10-01"
      }
    ]
  }
}

**To Reproduce**
input "starting on Sep 2019"

**Expected behavior**
It should be: 
{
  "Text": "starting on sep 2019",
  "Start": 0,
  "End": 19,
  "TypeName": "datetimeV2.daterange",
  "Resolution": {
    "values": [
      {
        "timex": "2019-09",
        "Mod": "after",
        "type": "daterange",
        "sourceEntity": "datetimerange",
        "start": "2019-09-01"
      }
    ]
  }
}

nawanas:
the Timex is interpreting starting on as being after the date range, hence it is only surfacing the 2019-10-01 date (Hence the mode After). will need to check when the mod is set to after (vs before or simply a date range

tellarin:
@nawanas I believe the modifier here is the issue. It should be "since", not "after".

gooooloo:
Will somebody merge the PR fix for this issue?

substill:
@tellarin , how can this fix be merged soon?

tellarin:
@gooooloo @substill The PR is missing a spec that shows the fix cover the case described in the issue. I'll test it ASAP and add the missing spec, if it works.

PR #2717, opened 2021-09-19
[EN DateTimeV2] Incorrect modifier fix (#2694)

ZulaMostafa:
PR for [issue #2694](https://github.com/microsoft/Recognizers-Text/issues/2694)

- Modified regex for both since and after modifiers.
- Modified some test cases as the modifier's expected output was after but it should be since.


tellarin (changes requested):
Looks good, but we need to add a spec corresponding to the reported issue to show that case works correctly.

tellarin (approved):

Issue #2711, opened 2021-09-05
[* Currency] Recognition of currency value when using [currencyname]$ (e.g. USD$) 


nawanas:
USD$ 115 is detected as currency but the value is returned as Null, where it should return the value of 115.  

the value is detected with USD 115 and $ 115

nawanas:
Adding MXN$ to this issue specifically and generalizing to all Dollar currencies

tellarin:
Workaround merged. But I'd leave this issue open, as a proper fix is needed.

LionbridgeCSII:
looking into this

nawanas:
> looking into this

this has already been fixed

tellarin:
> > looking into this
> 
> this has already been fixed

Not really. There is a mitigation in place for a couple case, but we need a more general soluton an many currency codes can appear alongisde $, like BRL$. 

The positioning of the space is also a potential issue (BRL$20 vs BRL$ 20 vs BRL $20), though I think this is secondary and should only be addressed if there's a solution that doesn't negatively impact perf.

PR #2721, opened 2021-09-26
Added USD$ and MXN$ to recognizable currencies

AhmedLeithy:
This PR solves #2711.

Added the tokens 'USD$' and 'MXN$' to currency suffix and prefix lists as a temporary solution to blocking issues in #2711. Constant files we regenerated for each programming language and test cases were added to the specs.


tellarin (approved):
Approving as-is as workaround. But we need a proper fix.

Issue #2689, opened 2021-08-04
[EN DateTimeV2] "last week of this month" is resolving to an invalid week in the previous month


jgreer013:
**Describe the bug**
"last week of this month" is resolving to an invalid week in the previous month, specifically resolving to the last week of the previous month. Presumably, it's conflating "last week" with "last month" and combining them.

**To Reproduce**
1. "Last week of this month"
2. Receive a timex of 2021-07-W05 when reference time is in August
3. Receive a timex of 2021-06-W05 when reference time is in July

**Expected behavior**
Expected timex of 2021-08-W04 and 2021-07-W04 respectively

**Platform (please complete the following information):**
 - Environment: LUIS

tellarin:
Also incorrect for "last week of the month"

PR #2726, opened 2021-10-10
[EN DateTimeV2] "last week of this month" resolving to previous month

AhmedLeithy:
There are two issues that can be identified in #2689 for the inputs "last week of this month" and "last week of the month". 

- Firstly, the resolution timex refers to the week 'W05', which the issue states is an "invalid" week. However, when going through the code, this seems to be an intentional design decision where any "last" week of the month will always return with a W05. This is stated in the [following docstring](https://github.com/microsoft/Recognizers-Text/blob/bf8eb3c78146e4559e1b3189073691dab897cda8/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs#L2211). 
  - However, I noticed that this behavior is not consistent in Python and JavaScript (the output is W04). We should decide on which behavior is correct and ensure consistency (perhaps as a separate issue). 

- The date would resolve to last month's dates instead of this month. This has been fixed for all 4 languages. Examples tested include: "last week of this month", "last week of the month", "last week of last month",  and "last week of next month".

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
@@ -1817,7 +1817,8 @@ namespace Microsoft.Recognizers.Text.DateTime
             int month;
             if (string.IsNullOrEmpty(monthStr))
             {
-                var monthText = !string.IsNullOrEmpty(match.Groups["relmonth"].Value) ? match.Groups["relmonth"].Value : trimmedText;
+                var relMonthValue = match.Groups["relmonth"].Value;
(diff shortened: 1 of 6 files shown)


AhmedLeithy:
> I suggest changing the other platforms to match it in this PR.

Done. The python and javascript code looks a bit different to .NET so I needed to add a new variable to keep track of whether this is the "5th" (i.e. last) week of the month.

Manually tested both Python and JS, and the output is now correct. Ran python tests and there are no regressions.

tellarin (approved):

TarasFedorenkoFTV:
Hello. I have the same problem in my .Net application, but with a slightly different query, "last Friday of this month," the result is returned for the previous month. So I think the problem is not solved. 
{B1C77C4A-732A-47AF-A578-31E282522370}

PR #2728, opened 2021-10-12
[Python] Add German support for recognizers-date-time

BrianInGermany:
- I converted the German Datetime YAML into a python file and attempted to adapt the English config files to the german patterns.
- I ran the pytest in the Python/tests directory and everything passed, except 8 tests, which were skipped.
- A few of the configs had to be commented out because I could not identify the as being supported in German
- The German holidays different than English ones are not all registered correctly and are thus not all recognizing or resolving correctly.
- Looking forward to feedback.

#2197


BrianInGermany:
I see there are some issues here... I will have a look

tellarin:
@BrianInGermany Are you still looking into this? I suggest cleaning the PR, so we can start by merging the basic code skeleton for German and its tests. Then implement and refine sub-type by sub-type.

tellarin:
Closing this as #2914 has added German Number/Currency/DateTimeV2 close to parity with .NET.

PR #2730, opened 2021-10-18
[EN Number] Fixed bug with "a next" getting parsed to infinite value fraction

### LINKED ISSUE #2690: [EN Number] "a next" got parsed as an "infinite" value number result
**Describe the bug**
Text "a next" got recognized as an "∞" (infinite) value number type with subtype "fraction".

**To Reproduce**
1. A query contains the phrase "a next".
2. Call NumberModel.Parse(query).
3. Check the result.

**Expected behavior**
This result should be avoided.

**Sample input/output**
As applicable, add examples of text input and json output to help explain the problem.

 - Platform: [e.g. .NET, JavaScript, Python, Java] .NET
 - Environment: [e.g. nuget package, npm package, LUIS, console app, in browser] nuget
 - Version of package [e.g. v1.0.8.1] I tried from 1.2.3 till latest 1.8.0

**Additional context**

AhmedLeithy:
Closes #2690.

The following queries all returned fraction entities with infinite value:
 - A next
 - An next
 - A previous
 - thirty and a next
 - Fourty two and a next

After investigation, I concluded that the issue stems from the extractor regex [FractionNounWithArticleRegex](https://github.com/microsoft/Recognizers-Text/blob/d26b9bf68f4904d3f6b6b62cf8fa560165d35a82/Patterns/English/English-Numbers.yaml#L105). This regex uses `AllOrdinalRegex` for extraction in the fraction part of the number, which also contains relative ordinal terms such as "next, previous, current" etc. 

Since these terms are not applicable to fractions ( You can say "A third", but you can't say "A next"), I've removed the relative ordinals from the corresponding regexes. In fact, I do not believe relative ordinals can never be used in expressing a fraction part of a number. Current behavior in all languages is:

- A next -> "next" is an ordinal. Same for other relative ordinals.
- thirty and a next -> "thirty" is an integer. "next" is an ordinal.

Please let me know if there are any other cases worth testing, or if there is something wrong with this approach.


.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
```diff
@@ -54,8 +54,10 @@ public static class NumbersDefinitions
       public const string RelativeOrdinalRegex = @"(?<relativeOrdinal>(next|previous|current)\s+one|(the\s+second|next)\s+to\s+last|the\s+one\s+before\s+the\s+last(\s+one)?|the\s+last\s+but\s+one|(ante)?penultimate|last|next|previous|current)";
       public static readonly string BasicOrdinalRegex = $@"({NumberOrdinalRegex}|{RelativeOrdinalRegex})";
       public static readonly string SuffixBasicOrdinalRegex = $@"(?:(((({TensNumberIntegerRegex}(\s+(and\s+)?|\s*-\s*){ZeroToNineIntegerRegex})|{TensNumberIntegerRegex}|{ZeroToNineIntegerRegex}|{AnIntRegex})(\s+{RoundNumberIntegerRegex})+)\s+(and\s+)?)*({TensNumberIntegerRegex}(\s+|\s*-\s*))?{BasicOrdinalRegex})";
+      public static readonly string SuffixNumberOrdinalRegex = $@"(?:(((({TensNumberIntegerRegex}(\s+(and\s+)?|\s*-\s*){ZeroToNineIntegerRegex})|{TensNumberIntegerRegex}|{ZeroToNineIntegerRegex}|{AnIntRegex})(\s+{RoundNumberIntegerRegex})+)\s+(and\s+)?)*({TensNumberIntegerRegex}(\s+|\s*-\s*))?{NumberOrdinalRegex})";
```
tellarin, line 57: Why did you have to introduce a new regex?

Specs/Number/English/NumberModel.json
```diff
@@ -3011,5 +3011,21 @@
         "End": 17
       }
     ]
+  },
+  {
+    "Input": "forty two and a next",
```
tellarin, line 3016: Can you add an example that's a real sentence? This is very valuable for proper evaluation and comparison between models/systems.
Can you also add other cases that may be affected negatively by your change related to "a"?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/English/NumbersDefinitions.cs
AhmedLeithy: The original setup was as follows: 

`FractionNounWithArticleRegex` (the regex causing this bug) included `AllOrdinalRegex` which makes use of `SuffixBasicOrdinalRegex` to include regexes for all ordinals. This includes the relative ordinals (next, previous etc.) that we do not want to include.

I wanted to create a parallel setup without the inclusion of the relative ordinals. Also I did not simply replace `AllOrdinalRegex` because it is used in other areas. Since it uses `SuffixBasicOrdinalRegex`, I could not replace that either, so I just created a new pair of regexes for only number and rounded number ordinals. They are:

`SuffixNumberOrdinalRegex` and `AllNumberOrdinalRegex` (since this only includes number ordinals and rounded number ordinals without including relative ordinals).

Please let me know if there is an alternative approach or better solution that I can adopt instead.

tellarin: Perhaps this can be extended to handle #2826, if it doesn't already.
If it works, let's add the relevant specs.

tellarin: @aitelint Is #2826 on your list?

aitelint: yes, but it seems it is possibly addressed by this PR. Should I look into it?

tellarin: Please do and validate if the current approach has perf impact that could be solved differently. Thanks.

tellarin:
Closing this PR as superseded by #2841. Added @AhmedLeithy as co-author there.

Issue #2681, opened 2021-07-22
[EN DateTimeV2] "one morning" gets incorrectly extracted as 1:00 AM


jgreer013:
**Describe the bug**
"one morning" is treated similarly to "one in the morning", in that it returns the next 1:00 AM time, but is not generally referring to that

**To Reproduce**
1. Input "Let's find one morning together"
2. Output is an exact time corresponding to the next 1:00 AM time

**Expected behavior**
Either nothing should be extracted, or only "morning" as a time range. Note that "one in the morning" should maintain its current correct behavior.

**Platform (please complete the following information):**
 - Environment: LUIS

tellarin:
This usage happens thorught the LUIS runtime enabling CalendarMode in the recognizers. Issue should be fixed in all modes.

LionbridgeCSII:
working on this

tellarin:
@LionbridgeCSII welcome back!
Please check both the regular mode and Calendar mode.
Thanks.

LionbridgeCSII:
@tellarin thanks!
yes will do. 
I am also addressing similar patterns like "The subject was discussed at three afternoon sessions." ("three afternoon" is also currently extracted as time).

PR #2746, opened 2021-11-02
 [EN DateTimeV2] fixed "one morning" incorrectly extracted as 1:00 AM (#2681)

LionbridgeCSII:
Fix to issue #2681 in .NET.

Test cases added to DateTimeModel and DateTimeModelCalendarMode.

One test case disabled because the pattern "three night" is extracted as Duration. This could be addressed in a separate issue.

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -21185,5 +21185,177 @@
         "ReferenceDateTime": "2021-06-18T18:00:00"
     },
     "Results": []
+  },
+  {
+    "Input": "Let's find one morning together",
+    "Context": {
+      "ReferenceDateTime": "2018-11-30T12:00:00"
+    },
+    "NotSupported": "java, javascript, python",
+    "Results": [
+      {
+        "Text": "morning",
+        "Start": 15,
+        "End": 21,
+        "TypeName": "datetimeV2.timerange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "TMO",
+              "type": "timerange",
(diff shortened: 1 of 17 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -21185,5 +21185,177 @@
...
+        }
+      }
+    ]
+  },
+  {
+    "Input": "They work at three night markets.",
+    "Context": {
+      "ReferenceDateTime": "2018-11-30T12:00:00"
+    },
+    "Comment": "Currently not supported because 'three night' is extracted as duration.",
```
tellarin, line 21317: Can you add a case for "night market/markets" in the ambiguity filter?

LionbridgeCSII: yes, but there seem to be many other such expressions: "night shifts, night jobs, night clubs..."
And the same happens with other duration units, e.g.
"2 year long projects" -> "2 year",
"3 day hospitals" -> "3 day"...

Would not be better to address the issue in DurationExtractor? 
We already implemented a similar solution in DatePeriod and TimePeriod (#2634).

tellarin: Checking frequencies, at least "night market/markets, shift/shits, and club/clubs" should be in the ambiguity filter. 
There will of course still be cases not caught. But we can filter these frequent ones. If the error happens in duration, then the filter should be in that type.

LionbridgeCSII: filter added

tellarin (approved):

PR #2756, opened 2021-11-09
 [* Currency] Expanded recognition of currency value when using [currencyname]$ (e.g. USD$) (#2711)

aitelint:
Fix to issue #2711 in English.

Expanded CurrencyPrefixList by combining CurrencyNameToIsoCodeMap with symbol '$'.

Test cases added to CurrencyModel.


tellarin (changes requested):
Looks reasonable, can it be propagated to the other Indo-European languages?

tellarin:
Was there any issue propagating to ES/PT/FR?

aitelint:
> Was there any issue propagating to ES/PT/FR?

They use localized currency names in some cases and English names in others (e.g. in French, 'FJD' corresponds to 'Dollar de Fidji' in CurrencyPrefixList and to 'Fijian dollar' in CurrencyNameToIsoCodeMap), this raises an error because the system tries to generate a dictionary with multiple entries with the same key.

The lists of currencies should be reviewed and modified in a consistent way.

tellarin:
> > Was there any issue propagating to ES/PT/FR?
> 
> They use localized currency names in some cases and English names in others (e.g. in French, 'FJD' corresponds to 'Dollar de Fidji' in CurrencyPrefixList and to 'Fijian dollar' in CurrencyNameToIsoCodeMap), this raises an error because the system tries to generate a dictionary with multiple entries with the same key.
> 
> The lists of currencies should be reviewed and modified in a consistent way.

Can you list such issues? Yes, all currency lists should be consistent. Please open an issue with your findings. Thanks!

@nawanas, this is related to the canonical outputs we've discussed previosuly. Like some languages having the type names translated, which should no longer happen in vNext.

aitelint:
> Can you list such issues? Yes, all currency lists should be consistent. Please open an issue with your findings. Thanks!

ok, will do.

aitelint:
> > Can you list such issues? Yes, all currency lists should be consistent. Please open an issue with your findings. Thanks!
> 
> ok, will do.

@tellarin, since the list of affected units was shorter than anticipated, I pushed directly a fix for FR, PT and ES.

Issue #2735, opened 2021-10-26
[DatetimeV2] Extending known holiday weekends


nawanas:
The phrase "Thanksgiving weekend" is currently realized as two separate dates, "Thanksgiving" and "weekend" leading to a confused resolution to the customers. This should actually boil down to one date which resembles the "Thanksgiving weekend". This applies to other known holidays/vacation like Easter, Labour Day, etc... This also applies to "the weekend of Nov 5th" which currently returns this weekend and the Nov 5th date as opposed to Nov 5-7

**Describe alternatives you've considered**
An alternative would be two dates that are actually representative of the request, so thanks giving weekend would be

{
  "Text": "thanksgiving",
  "Start": 0,
  "End": 11,
  "TypeName": "datetimeV2.date",
  "Resolution": {
    "values": [
      {
        "timex": "XXXX-11-WXX-4-4",
        "type": "date",
        "value": "2020-11-26"
      },
      {
        "timex": "XXXX-11-WXX-4-4",
        "type": "date",
        "value": "2021-11-25"
      }
    ]
  }
}
{
  "Text": "weekend",
  "Start": 13,
  "End": 19,
  "TypeName": "datetimeV2.daterange",
  "Resolution": {
    "values": [
      {
        "timex": "2021-W47-WE",
        "type": "daterange",
        "start": "2021-11-26",
        "end": "2021-11-28"
      }
    ]
  }
}

PR #2757, opened 2021-11-09
[EN DateTimeV2] Adding support for holiday-weekend mention patterns

AhmedLeithy:
This PR closes issue #2735 by adding supporting for extraction and resolution of holiday weekend dateranges.

### Approach

**Extraction**
Simply added a new regex `HolidayWeekendRegex`, which extracts expressions like the following examples:
-  Halloween weekend {year}?
- weekend of halloween {year}?
- Thanksgiving 2021 weekend
This regex was added to the "simple cases" regex list in the EnglishDatePeriodExtractor Configuration.

If the holiday extractor later extracts and matches to a specific holiday, it is added to a new field in the ExtractionResult Metadata called `HolidayName`. For example, the ExtractionResult for the text "Christmas weekend of 2021" will contain the metadata attribute holidayName value "Christmas".

**Parsing**
A function `ParseHolidayWeekend` is added to our BaseDatePeriodParser's Parse function. This attempts to resolve the input Extraction Result into the weekend closest to our holiday. This makes use of the `HolidayName` previously mentioned, otherwise we would need to attempt to extract the holiday again for resolution.

To be able to use the `HolidayParser` for holiday resolution, I needed to add the HolidayParser to all `IDatePeriodParserConfiguration` instances. However, the HolidayParser is only initialized in the `EnglishDatePeriodParserConfiguration`, since this feature has only been added to English for now.

### Tests
- Added a few test cases for weekend cases (true positives).
- Ensured that cases that might be affected (holiday dates and non-holiday weekends) do not suffer regressions.
- Reran all DateTime Recognizer tests for all supported languages (French, German etc.) to ensure there are no regressions.


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -235,6 +235,7 @@ public static class DateTimeDefinitions
...
       public const string HolidayList4 = @"(?<holiday>ramad(h)?an|ram(a)?zan|ramathan|eid al(-|\s+)adha|eid al(-|\s+)azha|eidul(-|\s+)azha|feast of the sacrifice|(islamic|arabic|hijri) new year|eid al(-|\s+)fitr|festival of breaking the fast)";
       public static readonly string HolidayRegex = $@"\b(({StrictRelativeRegex}\s+({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4}))|(({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+(of\s+)?({YearRegex}|{RelativeRegex}\s+year))?))\b";
+      public static readonly string HolidayWeekendRegex = $@"\b((({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+weekend)(\s+((of\s+)?({YearRegex}|{RelativeRegex}\s+year))))|(({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+((of\s+)?({YearRegex}|{RelativeRegex}\s+year)))?(\s+weekend))|(((the\s+)?weekend\s+of\s+)({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+((of\s+)?({YearRegex}|{RelativeRegex}\s+year)))?))\b";
```
tellarin, line 238: A better solution for perf would be to add the "weekend" term as an optional named group in HolidayRegex. This way the whole regex is only evaluated once, and a quick check for the group name in matches can be used to control code flow.

tellarin: Adding a new regex, and mixing the code with DatePeriodParser is not ideal in my view.
Couldn't HolidayExtractor/Parser produce either a date Holiday or a range Holiday? This seems a more general design that would also allow us to catch multi-day holidays.

tellarin: The year pieces could also be named groups to help with parsing, instead of running yet other regexes.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -242,6 +242,16 @@ private void AddTo(List<ExtractResult> dst, List<ExtractResult> src, string text
                         }
                         else
                         {
+                            if (result.Metadata?.IsHoliday ?? false)
```
tellarin, line 245: Whenever some new nested condition like this is introduced, please add a code comment describing the rationale and examples of the situations it's designed to cover.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -21171,6 +21171,118 @@
...
+    "Results": [
+      {
+        "Text": "the weekend of halloween",
+        "Start": 28,
+        "End": 51,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "XXXX-10-31",
```
tellarin, line 21189: Timex here is incorrect. It should mention it's a weekend ("WE"). And we should take care to use the right anchor date.
Also, what two resolutions are being returned here?
Shouldn't they be the weekend before the holiday and the weekend after the holiday?
Or is it meant to be only weekends that include the holiday?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -21171,6 +21171,118 @@
...
+    "Results": [
+      {
+        "Text": "labor day weekend",
+        "Start": 22,
+        "End": 38,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "XXXX-09-WXX-1-1",
```
tellarin, line 21245: It seems there's no consistency currently on the returned timexes?

tellarin (changes requested):

tellarin:
@AhmedLeithy, please also write better PR description titles, as these will be used when merging the PR in the commit history for the project.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
nawanas: if i am understanding the comment correctly, the adding of the weekend to the holiday would resolve it to a day and not to the actual weekend since it is a date and not a daterange, 

Also the holidays added don't all lend themselves to the holiday weekend structure (holidayList4 doesn't for sure)

tellarin: Holiday extraction/parser should be agnostic to one day or multi-day. We should try to move that way.
In any case, adding new regexes in multiple places is seldom desirable, as it negatively affects perf. Especially in big ever groing regexes like holidays.

Specs/DateTime/English/DateTimeModel.json
AhmedLeithy: Using the parsing for holidays as a reference:
When the input is the word "halloween" for example, two resolutions are returned: one for the previous halloween and the next halloween. This is the case for when the holiday is ambigious: "halloween 2021" would only resolve to the halloween of that year. 

A similar approach is implemented for weekends. The two resolutions are the two weekends corresponding to the previous and next instances of that holiday. (halloween 2021 weekend and halloween 2022 weekend).

As for the daterange itself, the resolved daterange itself extends from the holiday to the nearest weekend.

This is the behavior we thought was most appropriate, but it can be changed if you do not think it is correct.

Specs/DateTime/English/DateTimeModel.json
AhmedLeithy: This inconsistency stems from the HolidayParser. 

The current behavior is that if the holiday is ambiguous (no year specified), the timex returned by the holiday parser will be used. The holiday parser does not always return the same form of timex, which is the case even if you try the two queries "Halloween" and "Labor day" as individual date inputs. I believe it makes sense to keep the timex format consistent between holiday dates, and holiday weekends. What do you think?

AhmedLeithy:
Yea sorry, forgot to change the default title after writing the description.

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
AhmedLeithy: I agree that your suggestion could lead to better design overall; But I would like to note that the `HolidayExtractor` and the `HolidayParser` both only deal with date types. I checked the specs and attempted multiple multi-day holidays, but they all resolve to single point dates with the 'date' type (unless I am missing something). This is why it made sense to me to include holiday weekends in the dateperiodparser because they are 1. dateranges, 2. a special kind of weekend.

I'll address the regex change with groups now. I can also change the HolidayExtractor and HolidayParser to create and deal with dateranges and include the holiday-weekend logic if we still think that this change is worth doing. What do you think?

AhmedLeithy:
The following changes have been implemented:

- Now only one weekend regex is used and named groups are being used for extracting the date + year. The `HolidayWeekendRegex` is kept for readability, but effectively on the `HolidayRegex` is being used and it is being used once.

- The Extraction and Parsing logic has been completely moved to the BaseHolidayExtractor and BaseHolidayParser. They now both deal with both types: dates and dateranges. This can be expanded upon later for other multi-day holidays.

- The resolution for ambiguous holiday weekends like 'halloween weekend' will now return the correct timex for the weekends corresponding to the past and future resolutions of the holiday. (halloween 2021 and 2022 for example)

.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
AhmedLeithy: The performance issue has been addressed, and so is the design suggestion of moving the logic to the holidayExtractor/Parser. Please check [the updates summary.](https://github.com/microsoft/Recognizers-Text/pull/2757#issuecomment-965860347)

Specs/DateTime/English/DateTimeModel.json
AhmedLeithy: This has been updated, we can resolve the conversation if the new format is correct.

Patterns/English/English-DateTime.yaml
```diff
@@ -553,10 +553,14 @@ HolidayList3: !simpleRegex
...
+  def: (?<holidayWeekend>(({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+weekend)(\s+((of\s+)?({YearRegex}|{RelativeRegex}\s+year))))|(({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+((of\s+)?({YearRegex}|{RelativeRegex}\s+year)))?(\s+weekend))|(((the\s+)?weekend\s+of\s+)({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+((of\s+)?({YearRegex}|{RelativeRegex}\s+year)))?))
+  references: [ HolidayList1, HolidayList2, HolidayList3, HolidayList4, YearRegex, RelativeRegex, StrictRelativeRegex ]
 # Combined holiday regex
 HolidayRegex: !nestedRegex
-  def: \b(({StrictRelativeRegex}\s+({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4}))|(({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+(of\s+)?({YearRegex}|{RelativeRegex}\s+year))?))\b
-  references: [ HolidayList1, HolidayList2, HolidayList3, HolidayList4, YearRegex, RelativeRegex, StrictRelativeRegex ]
+  def: \b({HolidayWeekendRegex}|({StrictRelativeRegex}\s+({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4}))|(({HolidayList1}|{HolidayList2}|{HolidayList3}|{HolidayList4})(\s+(of\s+)?({YearRegex}|{RelativeRegex}\s+year))?))\b
```
tellarin, line 562: This still duplicates all holidays list both inside HolidayWeekendRegex and HolidayRegex. can't the "weekend" piece itself be the named group? No need to capture the holiday names along with it.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -21171,6 +21171,118 @@
...
+        "Resolution": {
+          "values": [
+            {
+              "timex": "2021-W43-WE",
+              "type": "daterange",
+              "start": "2021-10-30",
+              "end": "2021-11-01"
+            },
+            {
+              "timex": "2022-W43-WE",
```
tellarin, line 21195: Is it really the same week number? Just double checking.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -21171,6 +21171,118 @@
...
+        "Resolution": {
+          "values": [
+            {
+              "timex": "2021-W35-WE",
+              "type": "daterange",
+              "start": "2021-09-04",
+              "end": "2021-09-07"
+            },
+            {
+              "timex": "2022-W35-WE",
```
tellarin, line 21251: Same

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseHolidayParser.cs
```diff
@@ -28,6 +30,24 @@ public ParseResult Parse(ExtractResult result)
         }
 
         public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
+        {
+            if (string.IsNullOrEmpty(er.Metadata?.HolidayName))
```
tellarin, line 34: This check should be on the weekend. Not on the holiday name itself, right? Perhaps have two separate pieces of metadata?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseHolidayExtractor.cs
```diff
@@ -45,9 +49,29 @@ private List<Token> HolidayMatch(string text)
...
                 foreach (Match match in matches)
                 {
-                    ret.Add(new Token(match.Index, match.Index + match.Length));
+                    var metaData = new Metadata();
+
+                    // The objective here is to not lose the information of the holiday name
+                    // and year (if captured) when choosing. The data is extracted from the match
+                    // groups.
+
+                    if (match.Groups["holidayWeekend"].Success)
```
tellarin, line 61: Define the group names as constants (Constants.cs).

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseHolidayExtractor.cs
```diff
@@ -30,10 +31,13 @@ public List<ExtractResult> Extract(string text, DateObject reference)
             var ers = Token.MergeAllTokens(tokens, text, ExtractorName);
             foreach (var er in ers)
             {
-                er.Metadata = new Metadata
+                // If this is a daterange that contains a holiday, we should change its
+                // type to indicate that.
+
+                if (!string.IsNullOrEmpty(er.Metadata.HolidayName))
                 {
-                    IsHoliday = true,
```
tellarin, line 35: A holiday weekend is still a holiday. Can't we reuse the flow?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseHolidayExtractor.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Recognizers.Text.DateTime
     public class BaseHolidayExtractor : IDateTimeExtractor
     {
         private const string ExtractorName = Constants.SYS_DATETIME_DATE; // "Date";
+        private const string WeekendExtractorName = Constants.SYS_DATETIME_DATEPERIOD; // "Daterange";
```
tellarin, line 13: It could be named RangeExtractorName as it can be reused in the future for multi-day holidays. A code comment can explain the difference to the regular ExtractorName.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseHolidayExtractor.cs
AhmedLeithy: When the 'IsHoliday' flag is set, it makes the code flow towards the DateTime parsing/resolution (Single day holidays). 

In this instance the name is a bit misleading because what it effectively means is that the extracted entity is entirely a holday, like 'labor's day' or 'thanksgiving'. Should I change it to something like 'IsExactHoliday' or 'IsDateTimeHoliday'?

tellarin: I think so, or use two entries in metadata when relevant.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseHolidayParser.cs
AhmedLeithy: Definitely agreed. For readability, I have added two pieces of metadata: 
- metadata.IsHolidayRange
- metadata.IsHolidayWeekend
This would also allow us to later extend holiday ranges for other non-weekend holidays without having to revisit the logic of the flow.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseHolidayExtractor.cs
AhmedLeithy: I've went with the latter. Having the additional property "IsHolidayRange" makes this more readable.

Patterns/English/English-DateTime.yaml
AhmedLeithy: I've merged the regexes. There are three patterns for holiday weekends:
-Halloween weekend (year)?
-Halloween (year)? weekend
-Weekend of Halloween (year)?

Instead of repeating the holiday list for each pattern, I've merged the first two into HolidayRegex, and needed to repeat the holiday list once for the third pattern. I can remove this repetition but this would capture incorrect input patterns like:

"weekend of halloween weekend 2021" or "weekend of halloween 2021 weekend"

For that reason, I will keep the repetition for the third pattern (which starts with a weekend prefix, so most strings would be rejected quickly). If you would prefer removing the extra repetition, I can do hotfix on this early morning tomorrow before we merge.

Specs/DateTime/English/DateTimeModel.json
AhmedLeithy: Yea, I just double checked https://www.epochconverter.com/weeks/2021 (and the 2022 page), and it looks correct to me.

tellarin (approved):

Issue #2753, opened 2021-11-06
[ES currency] Currency with 'con' for decimals not completely recognized


srandal:
**Describe the bug**
Currency strings that use 'con' for the decimal only get the whole number recognized:
25 euros con 50 => 25 euro

Currency strings that use 'y' for the decimal get both parts recognized:
25 euros y 50 => 25,5 euro

Native speakers tell us that both forms are common in Spain.

**To Reproduce**
1. Run SimpleConsole for es-es
2. Type '25 euros con 50'
3. The value for the returned currency entity will only have the whole number portion of the input string.

**Expected behavior**
The value for the returns currency should capture both parts of the input string.

**Sample input/output**
25 euros con 50

I found the following entities (3):
{
  "Text": "25",
  "Start": 0,
  "End": 1,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "25"
  }
}
{
  "Text": "50",
  "Start": 13,
  "End": 14,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "50"
  }
}
{
  "Text": "25 euros",
  "Start": 0,
  "End": 7,
  "TypeName": "currency",
  "Resolution": {
    "isoCurrency": "EUR",
    "unit": "Euro",
    "value": "25"
  }
}

**Platform (please complete the following information):**
- Platform: .NET
- Environment: console app
- Version of package: 1.7.0

aitelint:
workin on this

PR #2762, opened 2021-11-10
 [ES Currency] Fixed currency with 'con' for decimals not completely recognized (#2753)

aitelint:
Fix to issue #2753.

Test cases added to ES DateTimeModel.


nawanas:
count we also add "com" for PT

aitelint:
> count we also add "com" for PT

added.

tellarin (approved):

Issue #2744, opened 2021-10-31
[.NET TimexLib] ISO week number erros


nawanas:
continuation to this bug reported in Python [[Python] Timex ISO week number fix by omid-jf · Pull Request #2565 · microsoft/Recognizers-Text (github.com)] (https://github.com/microsoft/Recognizers-Text/pull/2565)

Current week number resolutions are inconsistent with LUIS results since LUIS is using ISO format.
This pull fixes issues #188 #328 #2428 that still exist in the Python code.

this inconsistency exists in .NET

aitelint:
working on this

PR #2764, opened 2021-11-11
[.NET TimexLib] Fixed ISO week number erros (#2744)

aitelint:
Fix to issue #2744.

As commented in the code, a better solution would be to use ISOWeek.GetWeekOfYear, 
but it seems unavailable (even if it is supported in .NET Core 3.1 and .NET Standard 2.1).

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
@@ -533,7 +533,7 @@ namespace Microsoft.Recognizers.Text.DataTypes.TimexExpression.Tests
         public void DataTypes_Resolver_DateRange_Demaical_Period_EN()
         {
             var sourceLanguage = CultureInfo.CurrentCulture;
-            var testLanguage = new CultureInfo("en-En", false);
+            var testLanguage = new CultureInfo("en-US", false);
             CultureInfo.CurrentCulture = testLanguage;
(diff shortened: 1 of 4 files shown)


.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
```diff
@@ -533,7 +533,7 @@ public void DataTypes_Resolver_DateRange_Decimal_Period_PT()
         public void DataTypes_Resolver_DateRange_Demaical_Period_EN()
         {
             var sourceLanguage = CultureInfo.CurrentCulture;
-            var testLanguage = new CultureInfo("en-En", false);
+            var testLanguage = new CultureInfo("en-US", false);
```
AhmedLeithy, line 536: Is there any specific reason for this change?

AhmedLeithy (approved):

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexResolver.cs
aitelint: "en-EN" is not always recognized because it is not a supported identifier according to this [list](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/a9eac961-e77d-41a6-90a5-ce1a8b0cdb9c?redirectedfrom=MSDN) (I was getting an error when running the test).

tellarin (approved):

PR #2765, opened 2021-11-11
[FR|PT|ES Currency] Fixed expanded recognition of currency value when using [currencyname]$ (e.g. USD$) (#2711)

aitelint:
Fix to issue #2711 in FR, PT, ES.

Test cases added to CurrencyModels.


AhmedLeithy:
This looks good, but I have two concerns:

- Should we also implement this change in english? I implemented this [workaround for usd$ and mxn$](https://github.com/microsoft/Recognizers-Text/pull/2721) a few weeks ago, but your solution is more robust.

- I notice that this solution doubles the number of currencies in the match regex (since each currency will have a version without a $ sign, and one with a dollar sign). Would having the regex construction be as follows: "usd\\$? | mxn\\$?.." be more efficient than "(usd)| (mxn)| (usd\$) | (mxn\$).." when matching?

aitelint:
@AhmedLeithy, The fix for English and other languages has already been merged (#2756).

About the second point, yes NumberWithUnit uses a a tokenizer and a trie search to find matches instead of regexes. 
(I was also experimenting with another solution based on modifying the tokenizer that did not require doubling the currencies, but the current approach is simpler).

AhmedLeithy (approved):

AhmedLeithy:
Great, thanks for the clarifications.

tellarin (approved):

PR #2767, opened 2021-11-14
[Sequence] Address issue with emails not properly extracted (#2766)

### LINKED ISSUE #2766: [Sequence Email] Certain email addresses with multiple periods are not extracted
Email-ids autogenerated by CRM and other portals are not being extracted. These email addresses tend to have some numbers (example: environment-id) and multiple periods in them.

nawanas:
Fixes #2766.

Uniformingly preprocess all Sequence input to match the overall recognizers design.

- Add relevant e-mail spec test cases;
- Update all auto-gen resources across platforms.


nawanas:
Addresses issue #2766 

Added the support for capital letters in the RFC5322  Regex

tellarin (changes requested):
Changing the YAML is not enough. PRs should contain the auto-gen code to make sure the generation process works, as well as specs to show the change behaves correctly.
But in this specific case, I believe this is not the best approach due to perf. The recognizers design lowercases everything before recognition. Perhaps for email the pre-processing is not correclty in place? If capitalization is the issue, that should solve it.

PR #2776, opened 2021-11-22
[EN DateTimeV2] Increasing coverage for recurrence mentions

JahnaviChowdary:
Adding few more relevant keywords to each unit regex to increase the coverage for recurrence


tellarin (approved):
OK to merge. But next time, please add more spec for the new cases, ok?
Also, adding at least some new specs to CalendarMode is also desirable.
Thanks!

Issue #2747, opened 2021-11-02
[ZH Number] When the number string is long the extraction take a long time


nawanas:
**Describe the bug**
When trying a long string of number, "发议时，订单号 A202107211321363588510367 可用作" the extraction takes a long time, the longer the number the longer the time. 

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: Console app
 - Version of package v1.8

AhmedLeithy:
Investigated this issue in more details and found the following pieces of information:

- I've tested similar input in different languages including German, Arabic, Korean, and Japanese. No other languages experience similar slowdowns with long numbers. 

- This issue persists even if the digits are in Chinese.

- The slowdown depends on the length of the number part of the query. The longer the number (bigger), the slower our recognizers are. This applies in the cases for both Chinese digits and English digits within a Chinese sentence. The slowdown is unaffected by the length of the irrelevant parts of the query.

- For English digits within a Chinese sentence (like the example in the issue), changing the `NumberExtractor's` number mode in [ChineseNumberWithUnitExtractor Configuration](https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.NumberWithUnit/Chinese/Extractors/ChineseNumberWithUnitExtractorConfiguration.cs) to `CJKNumberExtractorMode.Default` solves the slow down issues. Currently this mode is set to `CJKNumberExtractorMode.ExtractAll`, which results in a single match for each digit within the number; which I hypothesized is the reason for the slow down.
  - The two recognizers that caused this slowdown for the example provided in the issue (english digits) are the `DateTimeRecognizer `and the `NumberWithUnitRecognizer.RecognizeCurrency`
  
  - The `CJKNumberExtractorMode.Default` change has no effect when the digits in the query are in Chinese. The slow down persists.
  
  - Unsurprisingly, there is a regression for the Numbers with Unit tests when the  ChineseNumberWithUnitExtractor configuration is changed to use a the CJKNumberExtractorMode.Default number mode. This is the only recognizer with regressions.

The following remain to be investigated:

- Which recognizers cause the slowdown when the numbers are in chinese? (it is not the same as when the characters are in English)
- What is the behavior of the regex matching in the Chinese numbers case? Does it also match each digit into its own regex match? Is this the reason for the slowdown?

aitelint:
working on this

PR #2796, opened 2021-12-09
[ZH Number] Fixed when the number string is long the extraction take a long time (#2747)

aitelint:
Fix to issue #2747.

The slowdown was caused by complex lookarounds used to avoid extracting numbers from percentage expressions in certain patterns. But since in English this kind of filtering is not implemented, they do not seem to be necessary and have been removed.

Test cases added to Chinese NumberModel and CurrencyModel.

Another issue observed is that long numbers in Chinese characters do not seem to be resolved correctly.


tellarin:
Hmmm. This is a bit tricky. In other languages a number is returned too if "inside" a percentage, like "17" in "17%". But here the whole number and percentage are recognized together.
Regarding the long number resolution issue, any insights on what is the type of failure?

tellarin (changes requested):
Marking as not to merge directly. We need to consider if there's a side effect first.

tellarin (approved):
Actually, upon a closer check these seem OK. Not sure why the previous extra processing was there.

tellarin:
@aitelint Regarding the long number resolution issue, any insights on what is the type of failure?
Should we open an issue for it?

aitelint:
> @aitelint Regarding the long number resolution issue, any insights on what is the type of failure? Should we open an issue for it?

I have not investigated the problem for now, only observed that the resolution is not correct, e.g.
"四千五百六十九万八千三百一十一亿四千八百七十二万五千三百八十二" -> 831194415382 instead of 4569831148725382
"二千零二十一万零七百二十一亿一千三百二十一万三千六百三十五" -> 72133423635 instead of 2021072113213635

Yes, the problem could be addressed in a separate issue (not sure however if these long expressions are used in practice)

Issue #2786, opened 2021-12-01
[ES Number] Large number expressed as a fraction not recognized correctly


srandal:
**Describe the bug**
When a large number is expressed as a fraction, e.g. "1 millón y medio", only the first part is recognized as a number.

**To Reproduce**
1. Run SimpleConsole for es-us
2. Type "1 millón y medio"
3. The hypothesized number will have a value of "1000000" (ignoring "y medio")

**Expected behavior**
The hypothesized number should be "1500000"

**Sample input/output**
1 millón y medio

I found the following entities (1):
{
  "Text": "1 millón",
  "Start": 0,
  "End": 7,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "1000000"
  }
}

**Platform (please complete the following information):**
- Platform: .NET
- Environment: console app
- Version of package: 1.7.0

aitelint:
working on this

PR #2810, opened 2021-12-20
[NL|EN|FR|DE|IT|PT|ES Number] Fixed large number expressed as a fraction not recognized correctly (#2786)

aitelint:
Fix to issue #2786 in EN|ES|IT.

Test cases added to NumberModels.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Spanish/Parsers/SpanishNumberParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.Number/Spanish/Parsers/SpanishNumberParserConfiguration.cs
@@ -104,10 +104,10 @@ namespace Microsoft.Recognizers.Text.Number.Spanish
             var resLen = result.Count;
             if (resLen > 2)
             {
-                if (result[resLen - 1] == "medio" && result[resLen - 2] == "y")
+                if (result[resLen - 1] == NumbersDefinitions.OneHalfTokens[1] && result[resLen - 2] == NumbersDefinitions.WordSeparatorToken)
                 {
-                    result[resLen - 2] = "con";
-                    result.Insert(resLen - 1, "un");
+                    result[resLen - 2] = NumbersDefinitions.WrittenFractionSeparatorTexts[0];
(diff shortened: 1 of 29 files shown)


nawanas:
this should generalize to more than just the half and quarter, the fractions up to a tenth would make sense, including multiples of these like two thirds, and in the other languages supported by number extractors

tellarin:
@aitelint Will you add the changes suggested by @nawanas soon? I'm fine with merging the current and having a separate issue opened for the suggested generalization.

.NET/Microsoft.Recognizers.Text.Number/Spanish/Parsers/SpanishNumberParserConfiguration.cs
```diff
@@ -95,6 +97,20 @@ public override IEnumerable<string> NormalizeTokenSet(IEnumerable<string> tokens
...
             }
 
+            // The following piece of code is needed to compute the fraction pattern number+'y medio'
+            // e.g. 'cinco y medio' ('five and a half') where the numerator is omitted in Spanish.
+            // It works by inserting the numerator 'un' ('a') in the list fracWords
+            // so that the pattern is correctly processed.
+            var resLen = result.Count;
+            if (resLen > 2)
+            {
+                if (result[resLen - 1] == "medio" && result[resLen - 2] == "y")
```
tellarin, line 107: Please move any hard coded terms to YAML resources.

tellarin (changes requested):
Should the relevant changes also be added to PT?

aitelint:
> @aitelint Will you add the changes suggested by @nawanas soon? I'm fine with merging the current and having a separate issue opened for the suggested generalization.

I am working on it right now, I think I will be able to push later today, or tomorrow

aitelint:
@tellarin, changes added and generalized to other European languages

.NET/Microsoft.Recognizers.Text.Number/Spanish/Parsers/SpanishNumberParserConfiguration.cs
aitelint: moved

tellarin (approved):

PR #2811, opened 2021-12-21
[* Unit] Micrograms not recognized as dimension type (#2798)

### LINKED ISSUE #2798: [* Unit] Micrograms not recognized as dimension type
**Describe the bug**

Using Text Analytics the service is unable to identify the micrograms as a dimension type.

For example, when sending the text:
100 mg;
100 mm;
100 μg;
100 micrograms
Only the first two are classified as dimension type, while the latter are classified as number type.

**To Reproduce**

Use TextAnalyticsClient.cs version 5.0.0.0 and RecognizeEntitiesBatchAsync() method.
Can repro issue on Postman with both v3.0 and v3.1 version.

**Expected behavior**
micrograms to be detected as dimension

**Sample input/output**

 - Platform: .NET
 - Environment: console app
 - Version of package: 5.0.0.0

**Additional context**

nawanas:
Fixes #2798.

A few things that surfaced
1- Weight is not yet supported in AR, JP, BG (units not yet supported in those languages.
2- This was already supported in PT and ES
~~3- Couldn't figure out how to do this in ZH~~
4- The spec sentences are not fullyconsistent, and in many cases miss all the possible versions of the different units. I added the spec in the ones that had Kilogram.


Specs/NumberWithUnit/English/DimensionModel.json
```diff
@@ -1234,6 +1234,38 @@
...
+        "End": 19,
+        "TypeName": "dimension",
+        "Resolution": {
+          "subtype": "Weight",
+          "unit": "Microgram",
+          "value": "120"
+        }
+      }
+    ]
+  },
```
tellarin, line 1268: Minor: When adding new cases, please always append at the end of the spec files. This makes merging easier.
Also, do not create a branch inside this repo. It's preferable to fork the repo and create branches for each PR there.

Specs/NumberWithUnit/French/DimensionModel.json
```diff
@@ -272,6 +272,38 @@
...
+        "Resolution": {
+          "subtype": "Weight",
+          "unit": "Microgramme",
+          "value": "120"
+        },
+        "Start": 16,
+        "End": 21
+      }
+    ]
+  },
```
tellarin, line 305: Nitpick: Be careful about identation and EOL and EOF changes.

tellarin (changes requested):
I'll add the values for Chinese and I'll update the original issue description to mention now only Arabic and Japanese are missing.

tellarin (approved):

PR #2820, opened 2022-01-05
[SV DateTimeV2] Patterns and specs for TimeZone extraction and parsing

pete1854:
Besides an issue w/ swedish diacritics which are removed by default (since line 34 in BaseTimeZoneExtractor.cs doesn't take into account any culture specific configurations) this should be ok.


Patterns/Swedish/Swedish-TimeZone.yaml
```diff
@@ -348,6 +695,101 @@ AmbiguousTimezoneList: !list
...
+   uzt: 300
+   vut: 660
+   wft: 720
+   wgst: -120
+   wgt: -180
+   wib: 420
+   wit: 540
+   wita: 480
+   #wst: 60
+   #wst: 780
```
tellarin, line 790: When there is ambiguity, please add like
wst: -10000
to represent it.

Specs/DateTime/Swedish/TimeZoneExtractor.json
```diff
@@ -0,0 +1,360 @@
+[
+  {
+    "Input": "Boka ett rum åt mig Afghansk tid",
```
tellarin, line 3: Isn't this just a duplicate of the next? The only difference seems to be capitalization in the entity, which should be ignored at the model level and final result. Feel free to leave it in to show that both work.

Specs/DateTime/Swedish/TimeZoneExtractor.json
```diff
@@ -0,0 +1,360 @@
...
+    ]
+  },
+  {
+    "Input": "Klockan är  Östamerikansk sommartid",
+    "Comment": "This currently doesn't work since diacritics are processed (removed) for all cultures. Line 34 in BaseTimeZoneExtractor.cs. Östamerikansk -> Ostamerikansk.",
+    "NotSupportedByDesign": "javascript, python, java",
+    "NotSupported": "dotnet",
+    "Results": [
+      {
+        "Text": "Östamerikansk sommartid",
```
tellarin, line 45: Adding "Ostamerikansk" and similar without the special chars in the dictionary doesn't solve it (i.e. the dictionary contains both versions of the string)? 
We can consider a better solution later.

Specs/DateTime/Swedish/TimeZoneExtractor.json
```diff
@@ -0,0 +1,360 @@
...
+    "NotSupportedByDesign": "javascript, python, java",
+    "Results": [
+      {
+        "Text": "Afghansk tid",
+        "Type": "timezone",
+        "Start": 20,
+        "Length": 12
+      }
+    ]
+  },
```
tellarin, line 13: But before we can mege the PR, we need to add the TimeZoneParser.json specs. To show that resolution also works.

tellarin (changes requested):

Specs/DateTime/Swedish/TimeZoneParser.json
```diff
@@ -1,4 +1,26 @@
 [
+  {
+    "Input": "Boka ett rum åt mig Amazonsk sommartid",
+    "Debug": true,
```
tellarin, line 4: Debug flag. :)

Patterns/Swedish/Swedish-TimeZone.yaml
tellarin: We still need to add the -1000 for the ambiguous case, like "amt", "ast", etc.

Specs/DateTime/Swedish/TimeZoneParser.json
```diff
@@ -0,0 +1,1048 @@
...
+            "UtcOffsetMins": 270
+          }
+        },
+        "Start": 20,
+        "Length": 18
+      }
+    ]
+  },
+  {
+    "Input": "Book me a room at beijing time",
```
tellarin, line 25: Just translation of these specs and adding the ambiguous resolution values in the dict and we're good to merge. 
Please add a couple new sentences that use Swedish-specific abbreviations too, including ambiguous ones, to document the behaviour in those cases.
Almost there. Thanks!

tellarin (changes requested):

Patterns/Swedish/Swedish-TimeZone.yaml
```diff
@@ -314,7 +611,7 @@ FullNameList: !list
     - Pacific
     - Eastern
 BaseTimeZoneSuffixRegex: !simpleRegex
-  def: ((\s+|-)(friendly|compatible))?(\s+|-)time(zone)?
+  def: ((\s+|-)(friendly|compatible))?(\s+|-)(ti(d|me))(zone)?
```
tellarin, line 614: Is there a coloquial expression similar to "friendly" or "compatible", as in the example in spec that would be used in Swedish between the location/timezone and the time/timezone word?

Specs/DateTime/Swedish/TimeZoneParser.json
pete1854: Will do. The biggest issue right now is the "you don't need to add 'time' or 'time zone'" to the entries in the FullToMinMapping dictionary. I need to sort that out first, since it currently prevents the parser from finding the entry in the dictionary resulting in "UTC+xx:xx..." resolutions. It isn't just a matter of adding "tid" to the regex... But, I will find a solution.

Specs/DateTime/Swedish/TimeZoneExtractor.json
pete1854: I temporarily added it to check that capitalization wasn't an issue. It is now removed.

Patterns/Swedish/Swedish-TimeZone.yaml
pete1854: Not that I can think of.

Specs/DateTime/Swedish/TimeZoneParser.json
pete1854: Hi @tellarin! I think I have found the reason for the parser not being able to get the correct timezone offset. It turns out that the parser calls `BaseTimeZoneParser` which is not localized, i.e. it doesn't use the swedish version of `TimeZoneDefinitions` when looking up the TZ in the `FullToMinMapping` dictionary of that class. As I see it we need to update `BaseTimeZoneParser` to be localizeable just as the corresponding `BaseTimeZoneExtractor` class.

pete1854: I have added a draft PR #2832 as a suggestion of how to solve this.

tellarin: @pete1854 Merged. Thanks a lot!

Specs/DateTime/Swedish/TimeZoneExtractor.json
pete1854: Implemented according to your suggestion.

Patterns/Swedish/Swedish-TimeZone.yaml
```diff
@@ -348,6 +645,101 @@ AmbiguousTimezoneList: !list
...
  entries:
+   # Swedish extensions
+   acwst: 525
+   almt: 360
+   #amt: 240
+   #amt: -240
+   anat: 720
+   aqtt: 300
+   #ast: 180
+   #ast: -240
```
tellarin, line 656: We're still missing the -1000 resolutions for the ambiguous timezone names. Can you add thise? Then I think we can merge. Thanks!

tellarin (changes requested):

Patterns/Swedish/Swedish-TimeZone.yaml
pete1854: Yeah, I'll fix this. I just wanted to commit my changes to see that it all worked before I refine the patterns.

tellarin: @pete1854 Sorry for the repetition. I saw your comment that you had implemented one of the required changes and as the build was passing I assumed the PR was ready for re-review.
What about from now on you leave a comment when the PR is ready to review after changes? Thanks for the great work BTW.

pete1854:
I think it is good to go now

pete1854:
This should be ready for merging now.

Patterns/Swedish/Swedish-TimeZone.yaml
```diff
@@ -348,6 +646,102 @@ AmbiguousTimezoneList: !list
 AbbrToMinMapping: !dictionary
  types: [ string , int ]
  entries:
+   # Swedish extensions
+   acwst: 525
+   almt: 360
+   #amt: 240
+   #amt: -240
+   anat: 720
```
tellarin, line 654: It looks like we're still missing the ambiguous resolution for some cases. Here in your file there are two commented out AMT, which suggests AMT is an ambiguous abbreviation, so there should be an entry for amt: -1000 to represent that.
AST and PYT are other cases. Are there more?
If these are not relevant in the target language, and you don't want/need to also accept English abbreviations, then the commented out duplicates could be deleted.

Patterns/Swedish/Swedish-TimeZone.yaml
```diff
@@ -348,6 +646,102 @@ AmbiguousTimezoneList: !list
...
+   gilt: 720
+   gyt: -240
+   hovt: 420
+   ict: 420
+   idt: 180
+   kgt: 360
+   kost: 660
+   lhdt: 660
+   lhst: 630
+   #mart: 570
```
tellarin, line 695: If the correct one is -570, what is this commented out one?

tellarin (changes requested):

pete1854:
New review

Patterns/Swedish/Swedish-TimeZone.yaml
```diff
@@ -692,9 +684,7 @@ AbbrToMinMapping: !dictionary
    kost: 660
    lhdt: 660
    lhst: 630
-   #mart: 570
-   # Used mart entry is for Marquesas Islands (French Polynesia) in the Pacific Ocean
-   mart: -570
+   mart: -10000
```
tellarin, line 687: In such ambiguous cases, we should keep the possible values also in the YAML file as documentation of the ambiguity. In the future, these can be used to offer alternatives.

tellarin (approved):

Issue #2803, opened 2021-12-14
[EN DateTime V2] Not all phrases of a form [day abbreviation][number] (e.g. "Mon 13th") are consistently recognized


apolya01:
**Describe the bug**
Shortened day-name + Date phrases are not consistently recognized/resolved. Examples of phrases:

Mon 13th 
Tue 13th 
Wed 13th 
Thu 14th 
Fri 14th 
Sat 14th 
Sun 14th

**Expected behavior**
All phrases conforming to this format should be recognized and resolved in a consistent manner - ideally with the entire phrase as one date - not either only the day part, or the day part and number part as separate entities.

**Sample input/output**
`recognizeDateTime("Mon 14th")`

```

(2) [{…}, {…}]
0:{start: 0, end: 2, resolution: {…}, text: 'mon', typeName: 'datetimeV2.date'}
end:2
resolution:{values: Array(2)}
values:(2) [{…}, {…}]
[[Prototype]]:Object
start:0
text:'mon'
typeName:'datetimeV2.date'
[[Prototype]]:Object
1:{start: 4, end: 7, resolution: {…}, text: '13th', typeName: 'datetimeV2.date'}
end:7
resolution:{values: Array(2)}
values:(2) [{…}, {…}]
```

`recognizeDateTime("Tue 14th")`
```
(1) [{…}]
0:{start: 0, end: 2, resolution: {…}, text: 'tue', typeName: 'datetimeV2.date'}
end:2
resolution:{values: Array(2)}
values:(2) [{…}, {…}]
[[Prototype]]:Object
0:{timex: 'XXXX-WXX-2', type: 'date', value: '2021-12-07'}
1:{timex: 'XXXX-WXX-2', type: 'date', value: '2021-12-14'}
[[Prototype]]:Object
length:2
[[Prototype]]:Array(0)
start:0
text:'tue'
typeName:'datetimeV2.date'
```

**Platform (please complete the following information):**
 - Platform: JavaScript
 - Environment: NodeJS service instance
 - 1.3.0

tellarin:
@aitelint Could you validate that this doesn't happen in .NET?

aitelint:
> @aitelint Could you validate that this doesn't happen in .NET? Is this a case of date and DOW not matching?

ok, I will check

aitelint:
@tellarin, yes these patterns work fine in .NET. 
Only "Mon 13th" gives problems because it is caught as "on 13th", I have pushed a fix for that.
[...]

PR #2823, opened 2022-01-06
[EN DateTime V2] Fixed not all phrases of a form [day abbreviation][number] (e.g. "Mon 13th") are consistently recognized in .NET (#2803)

aitelint:
Fix to issue #2803 in .NET.


tellarin (approved):
LGTM, but it seems if it's a regex change only, the specs would work in other platforms?

aitelint:
> LGTM, but it seems if it's a regex change only, the specs would work in other platforms?

But the issue was reported in JavaScript, so I assumed these patterns are failing at least there (but I did not check)

PR #2832, opened 2022-01-10
[* DateTimeV2] Make TimeZoneParser configurable

pete1854:
To be able to add more languages for the TimeZone parser it needs to be configurable. The current implementation is hardcoded for the English culture resulting in all TZ parsing being restricted to use TimeZone definitions as specified in the `English-TimeZone.yaml` file. This PR tries to make the TimeZoneParser configurable.


Specs/Number/Swedish/NumberModel.json
```diff
@@ -1381,6 +1381,7 @@
   {
     "Input": "en tjugoförstedels andel",
     "NotSupportedByDesign": "javascript, python, java",
+    "Debug": true,
```
tellarin: Debug flag.

.NET/Microsoft.Recognizers.Text.DateTime/Swedish/Parsers/SwedishTimeZoneParserConfiguration.cs
```diff
@@ -0,0 +1,33 @@
...
+        public SwedishTimeZoneParserConfiguration(IDateTimeOptionsConfiguration config)
+            : base(config)
+        {
+        }
+
+        string ITimeZoneParserConfiguration.TimeZoneEndRegex => TimeZoneEndRegex;
+
+        Dictionary<string, int> ITimeZoneParserConfiguration.FullToMinMapping => FullToMinMapping;
+
+        Regex ITimeZoneParserConfiguration.DirectUtcRegex => DirectUtcRegex;
```
tellarin, line 29: Ideally, some shareable resources like DirectUtcRegex should be in a BaseTimeZones YAML as it's not language dependent, but this looks pretty good.

.NET/Microsoft.Recognizers.Text.sln
```diff
@@ -15,12 +15,16 @@ EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Recognizers.Text.DateTime", "Microsoft.Recognizers.Text.DateTime\Microsoft.Recognizers.Text.DateTime.csproj", "{19522B81-C63F-4DD5-AB4C-674CD44D67D6}"
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Data Files", "Data Files", "{F070C131-80C0-4B0E-A462-2F6DB796759D}"
+	ProjectSection(SolutionItems) = preProject
+		..\Patterns\Turkish\Turkish-TimeZone.yaml = ..\Patterns\Turkish\Turkish-TimeZone.yaml
```
tellarin, line 19: Nit: Added twice? This first reference can be removed.

tellarin (changes requested):

tellarin (approved):
I've made the minor changes.

Issue #2722, opened 2021-09-29
Support for angles of rotation


tomconte:
**Is your feature request related to a problem? Please describe.**

We are using LUIS to implement voice commands for robot control. We would like to be able to understand sentences like "rotate 90 degrees" or "turn left 45 degrees". 

**Describe the solution you'd like**

Ideally the angle of rotation would be recognised and returned as a dimension, like length units. For our use case we probably only need to support degrees, but another useful unit would be a "turn", as in "make a half-turn" or "rotate half a turn"... Other applications might want to support radians.

**Describe alternatives you've considered**

We are currently getting the angle of rotation from a number entity, with no indication of the unit. We just assume that the angle is expressed in degrees.

aitelint:
working on this

PR #2833, opened 2022-01-11
[NL|EN|FR|DE|IT|PT|ES NumbersWithUnit] Added support for angles of rotation (#2722)

aitelint:
Implement extension #2722 in .NET.

Test cases added to DimensionModel.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Parsers/FrenchNumberWithUnitParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Parsers/FrenchNumberWithUnitParserConfiguration.cs
@@ -21,7 +21,9 @@ namespace Microsoft.Recognizers.Text.NumberWithUnit.French
             this.InternalNumberExtractor = NumberExtractor.GetInstance(numConfig);
             this.InternalNumberParser = AgnosticNumberParserFactory.GetParser(AgnosticNumberParserType.Number,
                                                                               new FrenchNumberParserConfiguration(numConfig));
-            this.ConnectorToken = NumbersWithUnitDefinitions.ConnectorTokenWithSpace;
+
+            // A space is added to the token to avoid interpreting part of a unit as a connector (e.g. 'de' in 'degrés')
+            this.ConnectorToken = NumbersWithUnitDefinitions.ConnectorToken + " ";
 
             this.TypeList = DimensionExtractorConfiguration.DimensionTypeList;
         }
--- a/.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Parsers/FrenchNumberWithUnitParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Parsers/FrenchNumberWithUnitParserConfiguration.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Recognizers.Text.NumberWithUnit.French
             this.InternalNumberExtractor = NumberExtractor.GetInstance(numConfig);
             this.InternalNumberParser = AgnosticNumberParserFactory.GetParser(AgnosticNumberParserType.Number,
                                                                               new FrenchNumberParserConfiguration(numConfig));
-            this.ConnectorToken = NumbersWithUnitDefinitions.ConnectorToken;
+            this.ConnectorToken = NumbersWithUnitDefinitions.ConnectorTokenWithSpace;
 
             this.TypeList = DimensionExtractorConfiguration.DimensionTypeList;
         }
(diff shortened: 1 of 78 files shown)


Specs/NumberWithUnit/English/DimensionModel.json
```diff
@@ -1360,5 +1360,90 @@
         "End": 89
       }
     ]
+  },
+  {
```
tellarin, line 1364: Can you add some spec cases to show that multiple temperature cases don't return false positives?

tellarin: Similarly to temperature to not catch "turn a 45 degrees angle" and similar as ambiguity filter.

tellarin (changes requested):

Specs/NumberWithUnit/English/DimensionModel.json
pete1854: Hi - Will this be propagated to the other languages currently supporting dimensions?

aitelint: addded specs and ambiguity filters to disambiguate temperature/angle cases.
propagate solution to other languages (NL|FR|DE|IT|PT|ES).

Patterns/French/French-NumbersWithUnit.yaml
```diff
@@ -723,7 +723,7 @@ BuildPrefix: !simpleRegex
   def: (?<=(\s|^|\P{L}))
 BuildSuffix: !simpleRegex
   def: (?=(\s|\P{L}|$))
-ConnectorToken: de
+ConnectorToken: 'de '
```
tellarin, line 726: Why add this whitespace?

tellarin (changes requested):

Patterns/French/French-NumbersWithUnit.yaml
aitelint: because otherwise 'de' in 'degres' is interpreted as connector and it fails to resolve the unit 'gres'.

tellarin: Could you spend some time trying to find a better solution? It seems not a good choice to add an arbritary whitespace here for one language and also pay the cost of trimming thousand of times across all cultures.
At the very least, there should be a comment in the YAML about it.
As the issue was not high-priority, we'll hold merging it a bit.

aitelint: ok, I will check

aitelint: Implemented new solution where the token with space is used only by the Parser. In this way there is no need to trim in the Extractor and other languages are not affected. 
Also, now Temperature supports the connector as well.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/French/Parsers/FrenchNumberWithUnitParserConfiguration.cs
```diff
@@ -21,7 +21,7 @@ public FrenchNumberWithUnitParserConfiguration(CultureInfo ci)
             this.InternalNumberExtractor = NumberExtractor.GetInstance(numConfig);
             this.InternalNumberParser = AgnosticNumberParserFactory.GetParser(AgnosticNumberParserType.Number,
                                                                               new FrenchNumberParserConfiguration(numConfig));
-            this.ConnectorToken = NumbersWithUnitDefinitions.ConnectorToken;
+            this.ConnectorToken = NumbersWithUnitDefinitions.ConnectorTokenWithSpace;
```
tellarin, line 24: As this is a French-only wokaround. I wonder if it wouldn't be simpler to just append a space to ConnectorToken in this specific config and have a code comment explaining the workaround.
Where exactly doesn't problem happen of mixing the connector and the unit term?

aitelint: yes, it would be the same, I will make the change.

When the extracted text is processed by the Parser, it is split at the found numbers in order to extract the units and the connector ends up attached to the following unit since there is no number between them (in the for loop at line 71 of NumberWithUnitParser).
Then later the connector is stripped from the unit if the unit string is found to start with the connector (line 111).

aitelint: moved space to French config file

tellarin (approved):
Approving with the FR workaround. We can look into solving the root cause when refactoring the units matcher.

Issue #2804, opened 2021-12-15
[PR Datetimev2] the phrase "5 e 45" is not extracted as a time


nawanas:
**Describe the bug**
a phrase lik e"reunião para 5 e 45" doesn't return the extraction of time (05:45 or 17:45) and only extracts them as numbers

the current returned object is
{
  "Text": "5",
  "Start": 13,
  "End": 13,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "5"
  }
}
{
  "Text": "45",
  "Start": 17,
  "End": 18,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "45"
  }
}

while it should be expected to return 
{
  "Text": "5",
  "Start": 13,
  "End": 13,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "5"
  }
}
{
  "Text": "45",
  "Start": 17,
  "End": 18,
  "TypeName": "number",
  "Resolution": {
    "subtype": "integer",
    "value": "45"
  }
}
{
  "Text": "5 e 45",
  "Start": 13,
  "End": 18,
  "TypeName": "datetimeV2.time",
  "Resolution": {
    "values": [
      {
        "timex": "T05:45",
        "type": "time",
        "value": "05:45:00"
      },
      {
        "timex": "T17:45",
        "type": "time",
        "value": "17:45:00"
      }
    ]
  }
}

tellarin:
This one is quite ambiguous. "reunião para as 5 e 45" is clear, but without the preposition "as", it can lead to some ambiguous cases.

PR #2834, opened 2022-01-11
[PT DateTimeV2] Fixed the phrase "5 e 45" is not extracted as a time (#2804)

aitelint:
Fix to issue #2804. 

test case added to DateTimeModel.


Patterns/Portuguese/Portuguese-DateTime.yaml
```diff
@@ -311,8 +311,8 @@ MidTimeRegex: !nestedRegex
   references: [ MidnightRegex, MidmorningRegex, MidafternoonRegex, MiddayRegex, MidEarlyMorning ]
 AtRegex: !nestedRegex
   # "as quatro" "às 3"
-  def: \b(((?<=\b([aà]s?)\s+)({WrittenTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex})(\s+horas?|\s*h\b)?|(?<=\b(s(er)?[aã]o|v[aã]o\s+ser|^[eé]h?)\s+|^\s*)({WrittenTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex})(\s+horas?|\s*h\b))(\s+{OclockRegex})?|{MidTimeRegex})\b
-  references: [ HourNumRegex, BaseDateTime.HourRegex, WrittenTimeRegex, OclockRegex, MidTimeRegex ]
+  def: \b(((?<=\b([aà]s?)\s+)({WrittenTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex}(\s+e\s+{BaseDateTime.MinuteRegex})?)(\s+horas?|\s*h\b)?|(?<=\b(s(er)?[aã]o|v[aã]o\s+ser|^[eé]h?)\s+|^\s*)({WrittenTimeRegex}|{HourNumRegex}|{BaseDateTime.HourRegex})(\s+horas?|\s*h\b))(\s+{OclockRegex})?|{MidTimeRegex})\b
+  references: [ HourNumRegex, BaseDateTime.HourRegex, BaseDateTime.MinuteRegex, WrittenTimeRegex, OclockRegex, MidTimeRegex ]
 ConnectNumRegex: !nestedRegex
```
tellarin, line 316: I believe we need to add at least some abiguity filter for cases like "entre 3 e 10" ("betweeb 3 and 10"). Also, does "reunião às cinco e quarenta" also works? I.e., same case with written numbers?

tellarin (changes requested):

Patterns/Portuguese/Portuguese-DateTime.yaml
aitelint: As for your comment in the issue, the pattern is extracted only if preceded by the preposition "às", so cases like "entre 3 e 10" are already excluded.
Patterns with written numbers should already be supported (they are included in WrittenTimeRegex) but I just noticed  they are not parsed correctly, so I will try to fix that.

tellarin: Cool. Can you add a spec case with "entre" to show it works as expected (no extraction). 
Also, does a complex case like "entre as 5 e as 7 da manha" work as a timerange? If not, let's open a low-priority issue for it.

aitelint: fixed parsing of written numbers and added specs.

tellarin (approved):

Issue #2826, opened 2022-01-09
[EN Number] "a current" or similar phrases determined as ordinal detected as a number fraction 


nawanas:
**Describe the bug**
The phrase "'I was a current student last semester" would detected current and last as ordinals, as well as "a current" as a number fraction. (The same would apply if a precedes "last" This is not a correct resolution to this term. This applies to all languages with RelativeOrdinalRegex defined 

 - Version of package v1..8.2]

tellarin:
Seems similar to #2690

PR #2841, opened 2022-01-14
[EN Number] Fixed "a current" or similar phrases determined as ordinal detected as a number fraction (#2826)

aitelint:
Fix to issue #2826 and #2690.

Test cases added to NumberModel.

### DIFF
--- a/Specs/Number/English/NumberModel.json
+++ b/Specs/Number/English/NumberModel.json
@@ -3478,5 +3478,36 @@
         "End": 23
       }
     ]
+  },
+  {
+    "Input": "Let's try one last time",
+    "Results": [
+      {
+        "Text": "one",
+        "TypeName": "number",
+        "Resolution": {
+          "subtype": "integer",
+          "value": "1"
+        },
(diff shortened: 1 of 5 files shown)


Specs/Number/English/NumberModel.json
```diff
@@ -3478,5 +3478,36 @@
         "End": 23
       }
     ]
+  },
+  {
+    "Input": "Let's try one last time",
```
tellarin, line 3483: Can you add the relevant specs to OrdinalModel and OrdinalModelSupress too to document that terms are properly processed there? Thanks.

tellarin (changes requested):

Specs/Number/English/NumberModel.json
tellarin: Please also bring the specs from #2730

aitelint: specs addded

tellarin (approved):

PR #2843, opened 2022-01-17
[JA Currency] Currency refinements

aitelint:
54 cases pass; 3 cases fail.

Failing cases:
- "人民元八千元", "人民元五十元": The unit is split around the number but NumberWihUnitExtractor explicitly prevents to extract both prefix and suffix units for Currency to avoid other issues.
- "金融市場で運用する資金運用事業を合わせた財投総額も過去最大の同二％減の四十九兆九千五百九十二億円となっている。": NumberExtractor does not correctly extract the large number.

Modified cases:
- A typo, '憶' instead of '億' has been corrected in a few inputs.
- "東芝の再編はもう一万憶が必要、取引銀行に借り入れを申請した。"  -> "東芝の再編はもう一万億円が必要、取引銀行に借り入れを申請した。", missing expected unit '円'.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Japanese/NumbersDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Japanese/NumbersDefinitions.cs
@@ -129,9 +129,9 @@ namespace Microsoft.Recognizers.Definitions.Japanese
       public const string PairRegex = @".*[対膳足]$";
       public const string RoundNumberIntegerRegex = @"(十|百|千|万(?!万)|億|兆)";
       public const string AllowListRegex = @"(。|，|、|（|）|”｜国|週間|時間|時|匹|キロ|トン|年|個|足|本|で|は|\s|$|つ|月|の|と)";
-      public static readonly string NotSingleRegex = $@"(?<!(第|だい))(({RoundNumberIntegerRegex}+(({ZeroToNineIntegerRegex}+|{RoundNumberIntegerRegex})+|{ZeroToNineFullHalfRegex}+|十)\s*(以上)?))|(({ZeroToNineIntegerRegex}+|{ZeroToNineFullHalfRegex}+|十)\s*({RoundNumberIntegerRegex}\s*){{1,2}})\s*(([零]?({ZeroToNineIntegerRegex}+|{ZeroToNineFullHalfRegex}+|十)\s*{RoundNumberIntegerRegex}{{0,1}})\s*)*\s*(\s*(以上)?)";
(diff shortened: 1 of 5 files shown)


tellarin:
> * "人民元八千元", "人民元五十元": The unit is split around the number but NumberWihUnitExtractor explicitly prevents to extract both prefix and suffix units for Currency to avoid other issues.

How common is this situation? We may need to have special handling for Japanese.

> * "金融市場で運用する資金運用事業を合わせた財投総額も過去最大の同二％減の四十九兆九千五百九十二億円となっている。": NumberExtractor does not correctly extract the large number.

Is this a somewhat "typical" case? If so, please open a bug to address it. 

Thanks.

aitelint:
> > * "人民元八千元", "人民元五十元": The unit is split around the number but NumberWihUnitExtractor explicitly prevents to extract both prefix and suffix units for Currency to avoid other issues.
> 
> How common is this situation? We may need to have special handling for Japanese.

I will check how to handle this pattern, it is used also in Chinese (where is also not supported)

aitelint:
> > * "金融市場で運用する資金運用事業を合わせた財投総額も過去最大の同二％減の四十九兆九千五百九十二億円となっている。": NumberExtractor does not correctly extract the large number.
> 
> Is this a somewhat "typical" case? If so, please open a bug to address it.

Actually here the number is extracted correctly by the NumberExtractor in Default mode but it fails in ExtractAll mode used by NumberWithUnit, so it should not be too difficult to fix, I will have a look.

tellarin:
> > > * "人民元八千元", "人民元五十元": The unit is split around the number but NumberWihUnitExtractor explicitly prevents to extract both prefix and suffix units for Currency to avoid other issues.
> > 
> > 
> > How common is this situation? We may need to have special handling for Japanese.
> 
> I will check how to handle this pattern, it is used also in Chinese (where is also not supported)

Any updates on this? Depending on how frequent, we could merge as-is and open an extension issue for this new behaviour.
I believe it's not super common...

aitelint:
> Any updates on this? Depending on how frequent the cases and how complex to fix, we could merge as-is and open an extension issue for this new behaviour. I believe it's not super common...

Probably yes, our Japanese linguist said expressions like "人民元10元" sound a bit odd and "10人民元" would be preferable (but on the other hand, a web search returns more results for "人民元10元" than for "10人民元").

I was trying to solve the issue by simply skipping the currency checks in NumberWihUnitExtractor for CJK cultures, but this creates other issues (e.g. "4人民币 5美元" returning a single result instead of 2). 

So, I would say to address this in a separate issue because a more careful treatment is needed.

PR #2847, opened 2022-01-19
[JA DateTimeV2] Date parser refinements

aitelint:
All test cases pass.


.NET/Microsoft.Recognizers.Text.DateTime/Utilities/DurationParsingUtil.cs
```diff
@@ -110,15 +110,18 @@ public static DateObject GetNextBusinessDay(DateObject startDate, bool isFuture
...
             var futureOrPast = future ? 1 : -1;
 
+            // timexUnitMap needs to be an ordered collection because the result depends on the order of the shifts.
+            // For example "1 month 21 days later" produces different results depending on whether the day or month shift is applied first
+            // (when the reference month and the following month have different numbers of days).
             foreach (var pair in timexUnitMap)
             {
-                var unitStr = pair.Key;
-                var number = pair.Value;
+                var unitStr = pair.Item1;
```
tellarin, line 123: Instead of a List of two elements, wouldn't a named tuple work better?

tellarin: This can be changed in a future PR.

tellarin (approved):

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/DurationParsingUtil.cs
aitelint: ok, I will change it.

Issue #2827, opened 2022-01-09
[* Number] Relative Ordinal not complete across languages


nawanas:
RelativeOrdinal has discrepancies across the different languages that needs to be unified. It is fully defined in EN, and partially defined, but not matching all the items in IT, JP,  DE, HI, KO, SW, TR. It is totally missing in the remaining supported languages.

aitelint:
working on this

tellarin:
@aitelint reopening issue as @nawanas mentioned some partial coverage in the other languages, but I'm not sure of the details.

tellarin:
@nawanas, can we close this now?

nawanas:
yes it could be closed

PR #2849, opened 2022-01-21
 [NL|FR|DE|IT|PT|ES Number] Fixed Relative Ordinal not complete across languages (#2827)

aitelint:
Fix to issue #2827 in NL|FR|DE|IT|PT|ES. 
Relative ordinals seem to be already supported in AR|HI|JA|KO|SV|TR.


.NET/Microsoft.Recognizers.Definitions.Common/German/NumbersDefinitions.cs
```diff
@@ -43,11 +43,12 @@ public static class NumbersDefinitions
       public static readonly string AllIntRegexWithLocks = $@"((?<=\b){AllIntRegex}(?=\b))";
       public static readonly string AllIntRegexWithDozenSuffixLocks = $@"(?<=\b)(((ein\s+)?halbes\s+dutzend)|({AllIntRegex}\s+dutzend(e)?))(?=\b)";
       public const string RoundNumberOrdinalRegex = @"(hundertst(er|es|en|el|e)?|tausendst(er|es|en|el|e)?|millionst(er|es|en|el|e)?|milliardst(er|es|en|el|e)?|billionst(er|es|en|el|e)?)";
-      public const string RelativeOrdinalRegex = @"(?<relativeOrdinal>(ante)?penultimate|letzter|nächster|vorheriger|aktueller)";
+      public const string RelativeOrdinalRegex = @"(?<relativeOrdinal>letzte[nr]?|nächste[nr]?|vorherige[nr]?|aktuelle[nr]?|jetzige[nr]?|vorletzte[nr]?|zuletzt|früher)";
```
tellarin, line 46: Shouldn't we also cover drittletzter?

tellarin (changes requested):
Thanks a lot! It would also be nice to have at least one spec per language in OrdinalModelSuppressExtendedTypes.json to show the flag works correctly.

.NET/Microsoft.Recognizers.Definitions.Common/German/NumbersDefinitions.cs
tellarin: vorletzte[nr]? and letzte[nr]? can also be combined in a simplified regex.

aitelint: added drittletzter and simplified regex

aitelint:
> Thanks a lot! It would also be nice to have at least one spec per language in OrdinalModelSuppressExtendedTypes.json to show the flag works correctly.

Added specs to OrdinalModelSuppressExtendedTypes.json

tellarin:
@aitelint Build is failing as suppress cases are not disabled for JS. Also, please take a look at the issue as it seems there's something missing in the other languages too. Thanks.

aitelint:
> @aitelint Build is failing as suppress cases are not disabled for JS. Also, please take a look at the issue as it seems there's something missing in the other languages too. Thanks.

yes, I will fix that soon.
I had a closer look at the other languages and apart for ZH that has no implementation of relative ordinals yet, there are a few specs missing in TR and SV (I will add them). 
The specs in the remaining languages (AR|HI|JA|KO) seem to cover all the patterns.
@nawanas, did you notice something missing in these languages?

tellarin (approved):

PR #2851, opened 2022-01-24
[KO Number] Korean NumberRange fixed extraction of pattern 'is 30 or at least 30'

aitelint:
"X 과 같거나 최소 Y" can be translated to "is X or at least Y" or "is equal to X or at least Y". 
In English, "is X or at least Y" returns one result, "at least Y", and "is equal to X or at least Y" returns 2 results, "equal to X" and "at least Y".
Now in Korean, "X 과 같거나 최소 Y" also returns 2 results: "X 과 같" (equal to X) and "최소 Y" (at least Y).

### DIFF
--- a/Specs/Number/Korean/NumberRangeModel.json
+++ b/Specs/Number/Korean/NumberRangeModel.json
@@ -670,6 +670,13 @@
     "NotSupportedByDesign": "javascript,python,java",
     "Results": [
       {
+        "Text": "5000과 같",
+        "TypeName": "numberrange",
+        "Resolution": {
+          "value": "[5000,5000]"
+        }
+      },
+      {
         "Text": "5000 보다 작다",
         "TypeName": "numberrange",
         "Resolution": {
@@ -1135,6 +1142,15 @@
     "Input": "그의 점수는 30과 같거나 최소 30이다",
     "NotSupportedByDesign": "java, javascript, python",
     "Results": [
+      {
+        "Text": "30과 같",
+        "TypeName": "numberrange",
+        "Resolution": {
+          "value": "[30,30]"
+        },
+        "Start": 7,
+        "End": 11
+      },
       {
         "Text": "최소 30",
         "TypeName": "numberrange",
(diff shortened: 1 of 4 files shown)


Specs/Number/Korean/NumberRangeModel.json
```diff
@@ -1136,12 +1136,12 @@
     "NotSupportedByDesign": "java, javascript, python",
     "Results": [
       {
-        "Text": "과 같거나 최소 30",
+        "Text": "최소 30",
```
tellarin, line 1139: Shouldn't the equals to 30 be it's own entity too?
I'm quite confused by the intended meaning of this sentenced.

tellarin: Can you add in the PR description an explanation as detailed as possible of the interpretation and a comparison to the implemented English behaviour? Thanks.

tellarin (changes requested):

Specs/Number/Korean/NumberRangeModel.json
tellarin: If I recall correctly, cases like "equals to 30" in English have special behaviour that "is 30" doesn't.

aitelint: yes, but according to our Korean linguist "과 같거나" here is ambiguous: 
"X 과 같거나 최소 Y" can be interpreted both as "is X or at least Y" or "is equal to X or at least Y", so we decided not to extract it.
There is another non-ambiguous case "그의 점수는 30과 같 거나 30보다 높다." (His score is equal to 30 or greater than 30) where "30과 같" is extracted.

aitelint: But I can add also "과 같거나" to the EqualRegex, maybe it is better to extract it.

tellarin: Can you add a comment to the spec case about "과 같거나" being ambiguous, to document the current behaviour? The rest looks good to me.

aitelint: added

tellarin (approved):

PR #2855, opened 2022-02-01
Add German Date Time Support in Java

mrussek:
Hi! My team at Workday needed to add German Date Time parsing support to the Java implementation. We did this mostly the same way we did the French implementation in 2020, by duplicating the logic from the C# version. This current version passes most of the German tests, and the only ones on which it fails are ignored in other languages as well.

Please take a look and see if this contribution will be acceptable. Thank you!


mrussek:
@tellarin would you mind taking a look at this?

tellarin:
> @tellarin would you mind taking a look at this?

@mrussek We'll review it ASAP. Thanks for the contribution.

BTW, we're looking into wrapping the .NET version in Java somehow, to avoid the need for porting new features/fixes. Is that something that could match your usage? We can discuss approaches if so.

mrussek:
@tellarin, if we could actually wrap the .NET version for Java, that would be excellent! I'm not sure how we would go about that though?

Specs/DateTime/German/DateExtractor.json
```diff
@@ -28,7 +28,7 @@
     "Context": {
       "ReferenceDateTime": "2016-01-01T00:00:00"
     },
-    "NotSupported": "javascript, python",
+    "NotSupported": "javascript, python, java",
```
tellarin, line 31: If you're adding German support, why are these not supported? Will support still come in a follow up PR?
Please add a comment in the description summarizing why these few cases are not yet supported to the best of your analysis of the reasons.

tellarin (approved):
I see a couple cases are not yet supported. But this PR adds enough value to merge.
Especially as the remaining cases are being worked on.

PR #2857, opened 2022-02-09
[Python] Fix parsed values format in number with unit (es-es)

sanhal:
- Using BaseMergedUnitParser and BaseMergedUnitExtractor for es-es CurrencyModel.
- Update __get_number_value to return number_value in format based on culture instead of general format.
- 
 
With this change, CurrencyModel for es-es will use **_BaseCurrencyParser_** and **___merged_compound_units_** (for extraction) 
and be able to resolve fractional units like "15 dolares y 15 centavos por favor" , "Son 25 dolares 50 por libra"

Resolved value is formatted to culture-based format with changes to function **___get_number_value_**

### DIFF
--- a/Python/libraries/recognizers-number-with-unit/recognizers_number_with_unit/number_with_unit/number_with_unit_recognizer.py
+++ b/Python/libraries/recognizers-number-with-unit/recognizers_number_with_unit/number_with_unit/number_with_unit_recognizer.py
@@ -177,7 +177,7 @@ class NumberWithUnitRecognizer(Recognizer[NumberWithUnitOptions]):
 
         # region Spanish
         self.register_model('CurrencyModel', Culture.Spanish, lambda options: CurrencyModel(
-            [ExtractorParserModel(BaseMergedUnitParser(SpanishCurrencyExtractorConfiguration(
+            [ExtractorParserModel(BaseMergedUnitExtractor(SpanishCurrencyExtractorConfiguration(
             )), BaseMergedUnitParser(SpanishCurrencyParserConfiguration()))]
         ))
         self.register_model('TemperatureModel', Culture.Spanish, lambda options: TemperatureModel(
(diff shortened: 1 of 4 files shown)


Python/libraries/recognizers-number-with-unit/recognizers_number_with_unit/number_with_unit/number_with_unit_recognizer.py
```diff
@@ -177,8 +177,8 @@ def initialize_configuration(self):
 
         # region Spanish
         self.register_model('CurrencyModel', Culture.Spanish, lambda options: CurrencyModel(
-            [ExtractorParserModel(NumberWithUnitExtractor(SpanishCurrencyExtractorConfiguration(
-            )), NumberWithUnitParser(SpanishCurrencyParserConfiguration()))]
+            [ExtractorParserModel(BaseMergedUnitParser(SpanishCurrencyExtractorConfiguration(
```
tellarin, line 180: This seems to be an invalid change as a model requires an extractor.

tellarin (changes requested):
This PR breaks a lot of test cases.

sanhal:
Example results before and after resolution value

> _"15 dolares y 15 centavos por favor"_

**Before**
{'value': '15', 'unit': 'Dólar'}
{'value': '15', 'unit': 'Centavo'}

**After**
{'value': '15.15', 'unit': 'Dólar'}

> _"Son 25 dolares 50 por libra"_

**Before**
{'value': '25', 'unit': 'Dólar'}
{'value': None, 'unit': 'Libra'}

**After**
{'value': '25.5', 'unit': 'Dólar'}
{'value': None, 'unit': 'Libra'}

tellarin (commented):
Could you please add a relevant spec test case or enable existing cases that didn't work in Python before?
Every PR with fixes or new features needs automated tests to both showcase the implemented functionality and to be used to avoid regressions in the future.

tellarin (approved):

PR #2865, opened 2022-02-23
Updating auto-gen resources across platforms

tellarin:


tellarin:
@chexi-ms, could you please review? Thanks.

chexi-ms (approved):

PR #2872, opened 2022-02-24
Update Portuguese datetime and Numbers with unit regex

samhickey25:
Update Portuguese datetime regex for SuffixAfterRegex, so it does not match on a period at the end of a datetime input. eg "11:55."

Update Portuguese NumbersWithUnit CompoundUnitConnectorRegex, so that compound sentences for currency will resolve eg "15 dólares e 15 centavos, por favor"

### DIFF
--- a/Specs/NumberWithUnit/Portuguese/CurrencyModel.json
+++ b/Specs/NumberWithUnit/Portuguese/CurrencyModel.json
@@ -1542,7 +1542,7 @@
         "Text": "15 dólares e 15 centavos",
         "TypeName": "currency",
         "Resolution": {
-          "value": "15.15",
+          "value": "15,15",
           "unit": "Dólar"
         },
         "Start": 14,
@@ -1558,7 +1558,7 @@
         "Text": "treze euros e quarenta e cinco centavos",
         "TypeName": "currency",
         "Resolution": {
-          "value": "13.45",
+          "value": "13,45",
           "unit": "Euro"
         },
         "Start": 14,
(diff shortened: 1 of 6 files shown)


tellarin (changes requested):
Looks OK to me, but can you add a couple new spec cases to show the change works as intended?
All PRs should have tests that show the change working and that can be used to avoid future regressions.
If possible, can you test if Spanish suffers the same issue and also fix there if needed?
Thanks.

samhickey25:
> 

I've added some specs, and made some changes to they python parsers for number_with_unit. This issue does not exist fro Spanish

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
```diff
@@ -1536,13 +1536,13 @@
...
     "Input": "custou apenas 15 dólares e 15 centavos.",
-    "NotSupported": "javascript, python, java",
+    "NotSupported": "javascript, java",
     "Results": [
       {
         "Text": "15 dólares e 15 centavos",
         "TypeName": "currency",
         "Resolution": {
-          "value": "15,15",
+          "value": "15.15",
```
tellarin, line 1545: You should not modify the separators. Please revert these changes so we can merge.

tellarin: This is a legacy design decision, that hopefully will be finally addressed and uniformly sorted out in release 2.0. But that will take a while.

samhickey25: No problem, reverted those changes

tellarin: There is only one revert pending for "3,99" in the output of "3 euros com 99 centavos".

samhickey25: Ah yes sorry! fixed now

tellarin (approved):

PR #2874, opened 2022-02-25
[JA NumberWithUnit] Japanese Dimension Support

aitelint:
All test cases pass.

- Modified BaseCJKNumberParser to support parsing of numbers in the form "一九九九" (1999).
- Modified NumberWithUnitParser to support parsing of split units like "秒速100メートル" ('speed per second 100 meters' = 100m/s).

Modified test cases:
-  "キロ" (kilo) was used in various test cases to indicate either "kilometer" or "kilogram". It has been replaced with the unambiguous expressions "キロメートル" (kilometer) and "キロメートル" (kilogram).


tellarin:
Is the interpretation for "キロ" by itself usually one or the other? Or somewhat equally frequent?

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
```diff
@@ -655,6 +661,12 @@ private double GetIntValue(string intStr)
 
                 hasPreviousDigits = char.IsDigit(intStr[i]);
 
+                // Japanese numbers in the form "一九九九" (1999) must be processed as digit numbers
+                if (Config.CultureInfo.Name == "ja-JP" && !hasPreviousDigits)
```
tellarin, line 665: Use a constant for the culture definition.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Constants.cs
```diff
@@ -36,6 +36,11 @@ public static class Constants
         // For cases like '2:00 pm', both 'pm' and '00 pm' are not dimension
         public const string AMBIGUOUS_TIME_TERM = BaseUnits.AmbiguousTimeTerm;
 
+        // In certain cultures the unit can be split around the number,
+        // e.g. in Japanese "秒速100メートル" ('speed per second 100 meters' = 100m/s).
+        // This flag tells the recognizer to combine prefix and suffix in order to parse the unit correctly.
+        public const string PARTIAL_UNIT = "Partial unit";
```
tellarin, line 42: I'd rather have constants with no whitespace in them as much as possible. Would split_unit be a better term?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/NumberWithUnitParser.cs
```diff
@@ -136,6 +136,24 @@ public ParseResult Parse(ExtractResult extResult)
                         resolutionStr += halfValue?.ResolutionStr.Substring(1);
                     }
 
+                    // In certain cultures the unit can be split around the number,
+                    // e.g. in Japanese "秒速100メートル" ('speed per second 100 meters' = 100m/s).
+                    // Here prefix and suffix are combined in order to parse the unit correctly.
+                    if (unitValue == Constants.PARTIAL_UNIT && unitKeys.Count > 1 && this.Config.CheckFirstSuffix)
```
tellarin, line 142: Don't other cultures already have a mechanism for this?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Parsers/NumberWithUnitParser.cs
aitelint: yes, but here it is a bit different. For common split units encountered so far, one part is sufficient to determine the whole unit, e.g. "Degree 30 Celsius". In this case ('speed per second 100 meters') both parts are required to parse the unit correctly.

aitelint:
> Is the interpretation for "キロ" by itself usually one or the other? Or somewhat equally frequent?

Equally frequent, it can be both and disambiguated by context ("I bought 2 kilos of oranges" (kg), "Next exit is in 2 kilos" (km)).

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseCJKNumberParser.cs
aitelint: done

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Constants.cs
aitelint: modified name

tellarin:
> > Is the interpretation for "キロ" by itself usually one or the other? Or somewhat equally frequent?
> 
> Equally frequent, it can be both and disambiguated by context ("I bought 2 kilos of oranges" (kg), "Next exit is in 2 kilos" (km)).

In this case, please add the ambiguous cases as currently NotSupported. Later, once DateTime has progressed, let's try to add support by checking some word in the immediate context of the mention for kilo, ok?

aitelint:
> > > Is the interpretation for "キロ" by itself usually one or the other? Or somewhat equally frequent?
> > 
> > 
> > Equally frequent, it can be both and disambiguated by context ("I bought 2 kilos of oranges" (kg), "Next exit is in 2 kilos" (km)).
> 
> In this case, please add the ambiguous cases as currently NotSupported. Later, once DateTime has progressed, let's try to add support by checking some word in the immediate context of the mention for kilo, ok?

ok, re-added original test cases.

tellarin (approved):

PR #2875, opened 2022-03-01
Fixing incorrect PT spec and updating resources across platforms

tellarin:


tellarin:
@chexi-ms, could you please review? Thanks.

PR #2902, opened 2022-03-15
[JA DateTimeV2] DatePeriod refinements

aitelint:
All test cases pass.
In DateTimeModel: 141 pass, 421 fail.

Methods from the standard implementation have been added and adapted to the CJK implementation when missing.

Modified test cases:

DatePeriodExtractor:
- Due to the ambiguity of 'after' and 'in', the result of the following cases has been modified and new cases were added to cover both patterns: "3年後に不在にします。", "Input": "3週間後に不在にします。", "Input": "3か月後に不在にします。"

DatePeriodParser:
- "私の週の残りの日は不在にします。" -> "私は週の残りの日は不在にします。", modified because duplicate.
- Resolution of cases involving 'decades' has been updated because there is no such term in Japanese e.g. "それは過去20年間に起きた。" ("It happened in the past 2 decades.") translates to "It happened in the past 20 years.". Similarly "それは次の10年間に起きた。", "それは次の30年間に起きた。", "日付がこの10年の売上".
- "今週の終わりごろ会うための時間をつくってください。" -> "今月下旬会うための時間をつくってください。", "今月の終わりごろ会うための時間をつくってください。" -> "今月下旬ごろ会うための時間をつくってください。", modified because duplicate.
- "一月十日から廿日にまで" -> "一月十日から廿日まで", grammatical error.
- Fixed various wrongly resolved cases, e.g. "4月の最終月曜日から10月1日まで" ("From the last Monday of April to October 1st") "Timex": "(XXXX-08-WXX-1-[#]4,XXXX-10-01,P36D)" -> "Timex": "(XXXX-04-WXX-1-[#]5,XXXX-10-01,P155D)"

Both Extractor and Parser:
- Various fixes to the extracted texts, e.g. adding missing prepositions "の間" (between), "まで" (to)..., and changing number formatting "２００８" -> "2008".


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/CJK/BaseCJKDatePeriodExtractor.cs
```diff
@@ -35,12 +35,19 @@ public List<ExtractResult> Extract(string text)
 
         public List<ExtractResult> Extract(string text, DateObject referenceTime)
         {
+            // Normalize input
+            if (this.config.NormalizeCharMap != null)
```
tellarin, line 39: This should be happening before it reaches here. Why is it not? Otherwise it will be called multiple times and also on other sub-types.

tellarin: And does each CJK language needs its own map? Can't one be shared?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -459,18 +474,6 @@ private DateTimeResolutionResult ParseSimpleCases(string text, DateObject refere
             }
             else
             {
-                match = this.config.SpecialMonthRegex.MatchExact(text, trim: true);
```
tellarin, line 462: Where did this go?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -742,18 +757,39 @@ private DateTimeResolutionResult ParseDayToDay(string text, DateObject reference
                     beginDay = this.config.DayOfMonth[dayFrom];
                     endDay = this.config.DayOfMonth[dayTo];
                 }
+                else if (match.Groups["half"].Success)
```
tellarin, line 760: Minor: Named group names should be constants.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -742,18 +757,39 @@ private DateTimeResolutionResult ParseDayToDay(string text, DateObject reference
...
+                {
+                    var er = this.config.DurationExtractor.Extract(match.Groups["half"].Value, referenceDate);
```
tellarin, line 762: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -742,18 +757,39 @@ private DateTimeResolutionResult ParseDayToDay(string text, DateObject reference
...
+                    var pr = this.config.DurationParser.Parse(er[0], referenceDate);
+
+                    var numberStr = pr.TimexStr.Substring(pr.TimexStr.IndexOf('P') + 1, pr.TimexStr.IndexOf('D') - 1);
```
tellarin, line 765: This should be a method in a timex class that parses it for number, no?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -837,10 +873,19 @@ private DateTimeResolutionResult ParseDayToDay(string text, DateObject reference
...
+                        beginLuisStr = DateTimeFormatUtil.LuisDate(beginYearForFutureResolution, beginMonthForFutureResolution, beginDay);
+                        endLuisStr = DateTimeFormatUtil.LuisDate(endYearForFutureResolution, endMonthForFutureResolution, endDay);
+                    }
+                    else
+                    {
+                        beginLuisStr = DateTimeFormatUtil.LuisDate(undefinedValue, undefinedValue, beginDay);
+                        endLuisStr = DateTimeFormatUtil.LuisDate(undefinedValue, undefinedValue, endDay);
+                    }
+
+                    ret.Timex = $"({beginLuisStr},{endLuisStr},P{durationDays}D)";
```
tellarin, line 888: Please declare a method in timexutility that receives the correct parameters and generates the timex. The whole date to string part should probably also happen there.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -1008,7 +1076,14 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                     var swift = 0;
                     if (nextMatch.Success)
                     {
-                        swift = 1;
+                        if (nextMatch.Groups["after"].Success)
```
tellarin, line 1079: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -939,9 +984,32 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
...
+
+                if (match.Success)
+                {
+                    return ParseLaterEarlyPeriod(text, referenceDate);
+                }
+            }
+
             if (match.Success)
             {
                 var monthStr = match.Groups["month"].Value;
```
tellarin, line 1015: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -1024,7 +1099,7 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                     if (this.config.IsWeekOnly(trimmedText))
                     {
                         var monday = referenceDate.This(DayOfWeek.Monday).AddDays(7 * swift);
-                        ret.Timex = DateTimeFormatUtil.ToIsoWeekTimex(monday);
+                        ret.Timex = isReferenceDatePeriod ? TimexUtility.GenerateWeekTimex() : DateTimeFormatUtil.ToIsoWeekTimex(monday);
```
tellarin, line 1102: Also a method in timexutil.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -1033,19 +1108,59 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                         ret.Success = true;
                         return ret;
                     }
+                    else if (!string.IsNullOrEmpty(match.Groups["restof"].Value))
```
tellarin, line 1111: Same and so on.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -1033,19 +1108,59 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
...
+                                endDate = DateObject.MinValue.SafeCreateFromValue(beginDate.Year, beginDate.Month, 1);
+                                endDate = endDate.AddMonths(1).AddDays(-1);
+                                diff = endDate.Day - beginDate.Day + 1;
+                                durationTimex = "P" + diff + Constants.TimexDay;
+                                break;
+
+                            case Constants.TimexYear:
+                                endDate = DateObject.MinValue.SafeCreateFromValue(beginDate.Year, 12, 1);
+                                endDate = endDate.AddMonths(1).AddDays(-1);
+                                diff = endDate.DayOfYear - beginDate.DayOfYear + 1;
```
tellarin, line 1137: Diff and timex to a timex gen method instead of repeating string concat everywhere.

tellarin: Actually, the whole timex in 1142 too.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -1407,25 +1921,51 @@ private DateTimeResolutionResult ParseNumberWithUnit(string text, DateObject ref
                 }
             }
 
+            return ret;
+        }
+
+        private DateTimeResolutionResult ParseDuration(string text, DateObject referenceDate)
```
tellarin, line 1927: Why a parseduration here, especially from text, and not reuse the existing duration parser?

tellarin: It's hard to understand the whole flow of what's calling what.

Specs/DateTime/Japanese/DatePeriodExtractor.json
```diff
@@ -742,25 +685,42 @@
   },
   {
     "Input": "3年後に不在にします。",
-    "NotSupported": "dotnet",
     "NotSupportedByDesign": "javascript,python,java",
-    "Results": []
+    "Results": [
```
tellarin, line 689: Please add a Comment tag in these specs to explain the ambiguity and the decision for the current behaviour.

Specs/DateTime/Japanese/DatePeriodExtractor.json
```diff
@@ -2212,22 +2062,28 @@
...
+    "NotSupported": "javascript, python, java",
     "Results": [
       {
         "Text": "1500",
         "Type": "daterange",
         "Start": 0,
         "Length": 4
+      },
+      {
+        "Text": "年",
```
tellarin, line 2074: Shouldn't year at most be a duration? Or supressed if just a "unit" for an actual year?

Specs/DateTime/Japanese/DatePeriodExtractor.json
```diff
@@ -2993,5 +2867,20 @@
         "Length": 4
       }
     ]
+  },
+  {
+    "Input": "3か月で不在にします。",
+    "NotSupportedByDesign": "javascript,python,java",
```
tellarin, line 2873: Please also add comments.

Specs/DateTime/Japanese/DatePeriodParser.json
```diff
@@ -1161,15 +1119,14 @@
     ]
   },
   {
-    "Input": "私の週の残りの日は不在にします。",
+    "Input": "私は週の残りの日は不在にします。",
```
tellarin, line 1122: Is "の" really incorrect or not used in practice here? If not, we should have both.

Specs/DateTime/Japanese/DatePeriodParser.json
```diff
@@ -5373,41 +5212,15 @@
       }
     ]
   },
-  {
-    "Input": "２００８年　十二月",
```
tellarin, line 5377: Does this case exist anywhere else?

tellarin (changes requested):
Partial review.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/CJK/BaseCJKDatePeriodExtractor.cs
aitelint: That is true, here we followed the same approach used in Japanese Date. 
I was thinking we could move it to MergedExtractor, so it would be called only once. 
But is not it that a user can also call each sub-entity separately? In that case the normalization would be missed.

aitelint: For the moment only Japanese needs normalization. For Chinese and Korean the map is null.

Specs/DateTime/Japanese/DatePeriodParser.json
aitelint: this case has been modified because duplicate, the version with "の" is still in the specs

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: This pattern is now parsed in ParseOneWordPeriod (in the if clause starting at line 1167: "if (this.config.IsMonthOnly(trimmedText))")

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: replaced all names

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: moved to TimexUtility

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: replaced with GenerateDatePeriodTimex

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: replaced DateTimeFormatUtil.ToIsoWeekTimex with TimexUtility.GenerateWeekTimex

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: done

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: replaced with GenerateDatePeriodTimexWithDiff

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: this is analogous to the method of the same name in standard implementation and deals with cases that involve duration ("2 upcoming days", "5 previous years"). 
The internal structure was already in BaseCJKDatePeriodParser and it has been refactored here.

Specs/DateTime/Japanese/DatePeriodParser.json
aitelint: added back (space was removed because normally it is not used between those characters, but it is not wrong to have it)

Specs/DateTime/Japanese/DatePeriodExtractor.json
aitelint: yes, it had been kept because it can be equivalent to "the year" in English. But since it is problematic, we added an ambiguity filter to remove its extraction.

Specs/DateTime/Japanese/DatePeriodExtractor.json
aitelint: added

Specs/DateTime/Japanese/DatePeriodExtractor.json
aitelint: added

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/CJK/BaseCJKDatePeriodExtractor.cs
aitelint: ah, I found what you mean, the normalization already occurs at the model level. So for the various sub-entities I will only add it to the corresponding TestBase methods.

aitelint: updated

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -195,7 +196,7 @@ public void TestDateTimeExtractor(TestModel testSpec)
 
             var referenceDateTime = testSpec.GetReferenceDateTime();
 
-            var actualResults = Extractor.Extract(testSpec.Input.ToLowerInvariant(), referenceDateTime);
+            var actualResults = Extractor.Extract(QueryProcessor.Preprocess(testSpec.Input), referenceDateTime);
```
tellarin, line 199: Shouldn't this preprocess should happen inside Extract, like in English?

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
```diff
@@ -219,7 +220,7 @@ public void TestDateTimeParser(TestModel testSpec)
 
             var referenceDateTime = testSpec.GetReferenceDateTime();
 
-            var extractResults = Extractor.Extract(testSpec.Input.ToLowerInvariant(), referenceDateTime);
+            var extractResults = Extractor.Extract(QueryProcessor.Preprocess(testSpec.Input), referenceDateTime);
```
tellarin, line 223: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -517,10 +520,10 @@ private DateTimeResolutionResult ParseYearToYear(string text, DateObject referen
 
                 if (yearMatch.Count == 2)
                 {
-                    var yearFrom = yearMatch[0].Groups["year"].Value;
-                    var yearTo = yearMatch[1].Groups["year"].Value;
-                    beginYear = int.Parse(yearFrom, CultureInfo.InvariantCulture);
-                    endYear = int.Parse(yearTo, CultureInfo.InvariantCulture);
+                    var yearFrom = yearMatch[0].Groups[Constants.YearGroupName].Value;
+                    var yearTo = yearMatch[1].Groups[Constants.YearGroupName].Value;
+                    beginYear = ConvertCJKToInteger(yearFrom);
```
tellarin, line 525: Do we still need this here as normalization is now supposed to always happen first?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
tellarin: If the diff is no used here, why not calc inside the timex gen method?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
tellarin: Please add a code comment for it. Thanks.

Specs/DateTime/Chinese/DateExtractor.json
```diff
@@ -144,9 +144,10 @@
   },
   {
     "Input": "10月12号，星期一",
+    "NotSupported": "java, javascript, python",
```
tellarin, line 147: Is this just a matter of making sure normalization always happen in the other plats too?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
aitelint: what do you mean? Normally this preprocess happens in the Parse method of the models (e.g. in DateTimeModel.cs). But when testing the sub-entities the model is not used.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: yes because ConvertCJKToInteger is used to parse years written in CJK characters or in dynasty form (e.g. "二零一六" or "康熙三" in Chinese).

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: it is used in the "TimexWeek" case to calculate the endDate. 
I have moved all the switch in the timex method now.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
aitelint: comment added

Specs/DateTime/Chinese/DateExtractor.json
aitelint: Since preprocessing does not happens in the other platforms, these test cases now fail because the extracted text is not normalized.

aitelint: added comment

.NET/Microsoft.Recognizers.Text.DataDrivenTests/TestBase.cs
tellarin: I see. My bad. Forgot this are the internal extractors by sub-type.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -457,6 +484,13 @@ public static string ModifyAmbiguousCenturyTimex(string timex)
             return "XX" + timex.Substring(2);
         }
 
+        public static float ParseNumberFromDurationTimex(string timex)
+        {
+            char[] unitChar = new char[] { 'D', 'W', 'M', 'Y', 'B' };
```
tellarin, line 489: This should either be static in class or in Constants.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -457,6 +484,13 @@ public static string ModifyAmbiguousCenturyTimex(string timex)
...
+            var numberStr = timex.Substring(timex.IndexOf('P') + 1, timex.IndexOfAny(unitChar) - 1);
```
tellarin, line 490: 'P' should reference an existing constant instead of hardcoded here.

tellarin (approved):
Approved with minor comments. 
Please add the small changes in a followup PR.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
aitelint: done in #2930

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
aitelint: done in #2930

PR #2894, opened 2022-03-11
[EN .NET] Fix TimexProperty.ToString() to handle DateTimeRanges properly

shana:
This fixes the conversion to string of DateTimeRange Timex types such as`"(2022-03-15T16,2022-03-15T18,PT2H)"` to not assume there's a part of day in the `TimexProperty` object when there isn't one.

Fixes #2893

PR #2903, opened 2022-03-15
[.NET Timexlib] Implement missing features of TimexProperty.ToString()

shana:
Following up on PR #2894 , this implements the missing features of TimexProperty.ToString(), adding support for date ranges, date + duration, etc.

This also fixes a bunch of off-by-one bugs when adding days or months to a Timex, where it would not take into account potential month/year overflows.

It also adds caching of the `Types` hashset of the Timex, so that it doesn't recalculate it every time the property is accessed. Instead, `Types` get cleared whenever a meaningful piece of data in the `TimexProperty` is changed.

Some tests have been converted to using `DataRow` instead of inline asserts, so that the test log will show the data being tested. It should also make it easier to see what the tests are doing. I also moved ToString tests to a separate class from the Roundtrip tests so it's easier to follow.

There's a lot of static helper methods that have been made into extension methods, because they're already in static classes so there's no overhead in doing it, and it makes caller code shorter and hopefully simpler to read.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimex.cs
+++ b/.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimex.cs
@@ -17,133 +17,107 @@ namespace Microsoft.Recognizers.Text.DataTypes.TimexExpression.Tests
         [TestMethod]
         public void DataTypes_Timex_FromDateTime()
         {
-            Assert.AreEqual("2017-12-05T23:57:35", TimexProperty.FromDateTime(new System.DateTime(2017, 12, 5, 23, 57, 35)).TimexValue);
+            Assert.AreEqual("2017-12-05T23:57:35",
+                TimexProperty.FromDateTime(new System.DateTime(2017, 12, 5, 23, 57, 35)).TimexValue);
         }
 
         [TestMethod]
-        public void DataTypes_Timex_Roundtrip_Date()
+        [DataRow("2017-09-27")]
+        [DataRow("XXXX-WXX-3")]
+        [DataRow("XXXX-12-05")]
+        public void DataTypes_Timex_Roundtrip_Date(string timex)
         {
-            Roundtrip("2017-09-27");
-            Roundtrip("XXXX-WXX-3");
-            Roundtrip("XXXX-12-05");
+            Roundtrip(timex);
         }
 
         [TestMethod]
-        public void DataTypes_Timex_Roundtrip_Time()
+        [DataRow("T17:30:45")]
+        [DataRow("T05:06:07")]
+        [DataRow("T17:30")]
+        [DataRow("T23")]
+        public void DataTypes_Timex_Roundtrip_Time(string timex)
         {
-            Roundtrip("T17:30:45");
-            Roundtrip("T05:06:07");
-            Roundtrip("T17:30");
-            Roundtrip("T23");
+            Roundtrip(timex);
         }
 
         [TestMethod]
-        public void DataTypes_Timex_Roundtrip_Duration()
+        [DataRow("P50Y")]
+        [DataRow("P6M")]
+        [DataRow("P3W")]
+        [DataRow("P5D")]
+        [DataRow("PT16H")]
+        [DataRow("PT32M")]
+        [DataRow("PT20S")]
+        public void DataTypes_Timex_Roundtrip_Duration(string timex)
         {
-            Roundtrip("P50Y");
-            Roundtrip("P6M");
-            Roundtrip("P3W");
-            Roundtrip("P5D");
-            Roundtrip("PT16H");
-            Roundtrip("PT32M");
-            Roundtrip("PT20S");
+            Roundtrip(timex);
         }
 
         [TestMethod]
-        public void DataTypes_Timex_Roundtrip_Now()
+        [DataRow("PRESENT_REF")]
+        public void DataTypes_Timex_Roundtrip_Now(string timex)
         {
-            Roundtrip("PRESENT_REF");
+            Roundtrip(timex);
         }
 
         [TestMethod]
-        public void DataTypes_Timex_Roundtrip_DateTime()
+        [DataRow("XXXX-WXX-3T04")]
+        [DataRow("2017-09-27T11:41:30")]
+        public void DataTypes_Timex_Roundtrip_DateTime(string timex)
         {
-            Roundtrip("XXXX-WXX-3T04");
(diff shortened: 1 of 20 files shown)


.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexHelpers.cs
```diff
@@ -217,7 +226,7 @@ public static TimexProperty TimexDateAdd(TimexProperty start, TimexProperty dura
                         return new TimexProperty
                         {
                             Year = start.Year,
-                            Month = (int)(start.Month + duration.Months),
+                            Month = ((int)(start.Month + duration.Months) % 12) + 1, // months are from 1 to 12
```
shana, line 229: I feel like this is still buggy, will add a test for it to tomorrow.

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimex.cs
```diff
@@ -17,138 +17,112 @@ public void DataTypes_Timex_FromDate()
...
+        [DataRow("(2017-09-08T21:19:29,2017-09-08T21:24:29,PT5M)")]
+        [DataRow("(XXXX-WXX-3T16,XXXX-WXX-6T15,PT71H)")]
+        public void DataTypes_Timex_Roundtrip_DateTimeRange_Start_End_Duration(string timex)
         {
-            Roundtrip("(2017-09-08T21:19:29,2017-09-08T21:24:29,PT5M)");
-            Roundtrip("(XXXX-WXX-3T16,XXXX-WXX-6T15,PT71H)");
-        }
-
-        [TestMethod]
-        public void DataTypes_Timex_ToString()
```
tellarin, line 112: What happened to the cases from this one onwards?

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexConvert.cs
```diff
@@ -15,276 +15,314 @@ public void DataTypes_Convert_CompleteDate()
...
-            Assert.AreEqual("morning", TimexConvert.ConvertTimexToString(new TimexProperty("TMO")));
-            Assert.AreEqual("afternoon", TimexConvert.ConvertTimexToString(new TimexProperty("TAF")));
-            Assert.AreEqual("evening", TimexConvert.ConvertTimexToString(new TimexProperty("TEV")));
+            Assert.AreEqual(expected, TimexConvert.ConvertTimexToString(new TimexProperty(timex)));
         }
 
         [TestMethod]
-        public void DataTypes_Convert_FridayEvening()
+        [DataRow("(XXXX-XX-XX,XXXX-XX-XX,P1D)", "1 day")]
+        public void DataTypes_Convert_Period(string timex, string expected)
```
tellarin, line 176: It's better to rename to DataTypes_Convert_Undefined_Period or DataTypes_Convert_Undefined_Range for clarity.

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexConvert.cs
```diff
@@ -15,276 +15,314 @@ public void DataTypes_Convert_CompleteDate()
...
-            var timex = new TimexProperty("(2017-09-08T21:19:29,2017-09-08T21:24:29,PT5M)");
-
-            // TODO
+            Assert.AreEqual(expected, TimexConvert.ConvertTimexToString(new TimexProperty(timex)));
         }
 
         [TestMethod]
-        public void DataTypes_Convert_WednesdayToSaturday()
+        [DataRow("(2017-09-08T21:19:29,2017-09-08T21:24:29,PT5M)", "8th September 2017 9:19:29PM to 9:24:29PM")]
+        public void DataTypes_Convert_Last5Minutes(string timex, string expected)
```
tellarin, line 197: What if the dates are different? Do we have a case for that somewhere for it?

Also, this is technically not "last 5 min". It's just a period of 5 min with those parameters.
And the order in the timex is invalid. Start should be smaller than end.

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexHelpers.cs
```diff
@@ -93,7 +93,7 @@ public void DataTypes_Helpers_TimeRangeFromTimex()
         public void DataTypes_Helpers_DateFromTimex()
         {
             var timex = new TimexProperty("2017-09-27");
-            var date = TimexHelpers.DateFromTimex(timex);
+            var date = TimexHelpers.DateTimeFromTimex(timex);
```
tellarin, line 96: Both should return the same results, but the first one is "more correct" as the timex represents a date, not a datetime.

tellarin (changes requested):
Partial review

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimex.cs
shana: I moved them to `TestTimexToString` because all the tests before this are pure roundtrip tests and it was getting hard to track what tests tested what codepaths.

.NET/Microsoft.Recognizers.Text.DataTypes.DataDrivenTests/TestTimexConvert.cs
shana: I didn't actually do anything with this, I just converted the test to DataRow and added the output that it was missing - the github inline diff is misleading. This is the original code of the test:

```
[TestMethod]
        public void DataTypes_Convert_Last5Minutes()
        {
            // date + time + duration
            var timex = new TimexProperty("(2017-09-08T21:19:29,2017-09-08T21:24:29,PT5M)");

            // TODO
        }
```

I didn't change anything in the minute duration to string conversion, so the output here is the same output that it has always had. I agree that the output doesn't seem to match the output of the test, I just added the output of the original test here so I could see what it was doing.

shana:
Fyi, I've been away for Easter and life and couldn't follow up on this before, I should have more time now to get this thing done, it's not forgotten.

tellarin:
> Fyi, I've been away for Easter and life and couldn't follow up on this before, I should have more time now to get this thing done, it's not forgotten.

Thanks for the update, @shana!

tellarin:
@shana Will you still update this PR? Looking forward to merging it.

Issue #2850, opened 2022-01-22
[* DateTimeV2] Date range (without year) is processed as a single Date with Year


veonua:
**Describe the bug**
the phrase of the Date period from 
https://github.com/microsoft/Recognizers-Text/blob/master/Specs/DateTime/English/DatePeriodExtractor.json#L2133
    "Input": "I'll be out November 19-20",

is gives a single date result of Nov 19th year 2020

**To Reproduce**

DateTimeRecognizer.RecognizeDateTime("I'll be out November 19-20", Culture.English);

![image](https://user-images.githubusercontent.com/1540142/150657109-d88cab74-ad46-4857-889d-2546cbaa12db.png)

**Expected behavior**

daterange [2022-11-19 .. 2022-11-20] 

**Sample input/output**
As applicable, add examples of text input and json output to help explain the problem.

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: nuget package, LUIS
 - Version of package: 1.8.2 (broken)  1.2.9 (fine)

**Additional context**
I've tried to update my library, now my unit tests fail on this query

tellarin:
@aitelint This is an unnoticed regression. When looking into it, please add spec for all model modes. Thanks.

tellarin:
@nawanas Re-tagging as P1 as this is a regression, which should not have happened.

PR #2907, opened 2022-03-22
 [* DateTimeV2] Fix date range (without year) is processed as a single Date with Year (#2850)

aitelint:
Fix to issue #2850 in NL|EN|FR|DE|IT|PT|ES

Test cases added to DateTimeModels.

### DIFF
--- a/Specs/DateTime/Dutch/DateTimeModel.json
+++ b/Specs/DateTime/Dutch/DateTimeModel.json
@@ -1,4 +1,4 @@
-[
+﻿[
   {
     "Input": "Ik ga 4e jan 2019 terug",
     "Context": {
@@ -20023,6 +20023,37 @@
               "type": "daterange",
               "sourceEntity": "datetimerange",
               "end": "2000-01-01"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "Ik ben afwezig november 19-20",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupportedByDesign": "javascript, python, java",
+    "Results": [
+      {
+        "Text": "november 19-20",
+        "Start": 15,
+        "End": 28,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(XXXX-11-19,XXXX-11-20,P1D)",
+              "type": "daterange",
+              "start": "2015-11-19",
+              "end": "2015-11-20"
+            },
+            {
+              "timex": "(XXXX-11-19,XXXX-11-20,P1D)",
+              "type": "daterange",
+              "start": "2016-11-19",
+              "end": "2016-11-20"
             }
           ]
         }
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -1,4 +1,4 @@
-[
+﻿[
   {
     "Input": "I'll go back 04th Jan 2019.",
     "Context": {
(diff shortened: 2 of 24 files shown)


veonua:
could you please merge this

Specs/DateTime/Dutch/DateTimeModel.json
```diff
@@ -19986,5 +19986,36 @@
...
+    "Results": [
+      {
+        "Text": "november 19-20",
+        "Start": 15,
+        "End": 28,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(XXXX-11-19,XXXX-11-20,P1D)",
```
tellarin, line 20005: Resolution is incorrect? Should be P2D start in 19 (inclusive) and ends in 21 (not inclusive).

tellarin: Is this case different from some other behaviour? We may have a breaking problem here.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -23005,5 +23005,35 @@
...
+    "Results": [
+      {
+        "Text": "november 19-20",
+        "Start": 12,
+        "End": 25,
+        "TypeName": "datetimeV2.daterange",
+        "Resolution": {
+          "values": [
+            {
+              "timex": "(XXXX-11-19,XXXX-11-20,P1D)",
```
tellarin, line 23023: Incorrect resolution. Text clearly implies P2D.

tellarin (changes requested):

Specs/DateTime/Dutch/DateTimeModel.json
veonua: I believe it's another big topic - the date range resolutions are messed up, 
there is no clear rule when the last day/week/month is inclusive or not.

aitelint: This is how the parsing of this pattern is currently implemented. 
There are similar cases in the specs that follow the same logic, for example:
- "I'll be out from 4-23 in next month" -> (2016-12-04,2016-12-23,P19D)
- "I'll be out between 3 and 12 of Sept hahaha" -> (XXXX-09-03,XXXX-09-12,P9D)
- "APEC will happen in Korea on November 18-19" -> (XXXX-11-18,XXXX-11-19,P1D)
- "9 to 12 of June: another tapa festival" -> (XXXX-06-09,XXXX-06-12,P3D)

Specs/DateTime/English/DateTimeModel.json
aitelint: Should we modify this behavior? 
But it will affect other existing test cases (like the ones pointed above)

veonua: It would be great to add a flag if the last day must be included.

Because at this moment extra day is added for months and years (I will be off in April -> ends 1st of May)
But for weeks and days it’s not included (April 1-30 -> ends 30 Apr) 

while I (human) read both of these cases the same

tellarin:
@aitelint Can you resolve these conflicts? Thanks.

aitelint:
> @aitelint Can you resolve these conflicts? Thanks.

conflicts solved

Specs/DateTime/English/DateTimeModel.json
aitelint: @tellarin, @venoua, should we address this in a separate issue? (since the parsing problem is not related to the bug solved here)

veonua: date resolution is much bigger issue and should be solved in separated issue. please merge this PR

tellarin: @aitelint Yes, changes will come as part of another issue. Let me create one soon. We're discussing alternatives as it can be considered a breaking change.

Specs/DateTime/Dutch/DateTimeModel.json
tellarin: I see. Some could be considered ambiguous, but some seem clearly incorrect. Let's not modify for now though, as this is somewhat of a breaking change.

tellarin (approved):
Approving as-is as to not modify existing behaviour and re-resolution is beyond the scope of the issue this PR fixes.

PR #2910, opened 2022-03-25
[Python][ES-MX Number] Delimitated number value fix

Conor-Keaney:
This is a fix for using the correct deliminator for ES-MX values i.e. using the comma for a thousand marker instead of a dot


tellarin (changes requested):
Thanks for the contribution. But unfortunately this change is not currently acceptable.
- It is considered a breaking change., so we can't merge it without a plan for a breaking change release.  There is an open discussion that provides more background on the issue.
- This goes against the intended design for when that change comes. I.e., output should always be in a canonical form, and not language dependent, to avoid such issues from cropping up and use of consumption.
-  .NET is the master version and no test case can be disabled for it due to behaviour difference in other platforms. If such changes are to be implemented, support must also exist in .NET.
This will be put on hold until there's a decision on tap discussion. Thanks for understanding.

PR #2913, opened 2022-04-07
MS Recognisers python - Italian Number, Currency, Datetime to parity with .NET

ushmau5:


Python/libraries/recognizers-number/recognizers_number/culture.py
```diff
@@ -12,7 +12,7 @@
     Culture.German: LongFormatMode.DOUBLE_DOT_COMMA,
     Culture.Portuguese: LongFormatMode.DOUBLE_DOT_COMMA,
     Culture.French: LongFormatMode.DOUBLE_DOT_COMMA,
-    Culture.Japanese: LongFormatMode.DOUBLE_COMMA_DOT
+    Culture.Italian: LongFormatMode.DOUBLE_DOT_COMMA,
```
tellarin, line 15: We should not remove the Japanese entry.

Python/tests/test_runner_number.py
```diff
@@ -27,10 +27,12 @@ def test_number_recognizer(culture, model, options,
 
         assert_verbose(actual.type_name, expected['TypeName'], spec_info)
         assert_verbose(actual.text, expected['Text'], spec_info)
-        assert_verbose(actual.start, expected['Start'], spec_info)
-        assert_verbose(actual.end, expected['End'], spec_info)
         assert_verbose(actual.resolution['value'], expected['Resolution']['value'], spec_info)
 
+        if 'Start' in expected and 'End' in expected:
```
tellarin, line 32: Why the need for these? Are 'start' and 'end' missing onlyin Italian specs? If so, we should fix the specs.

Python/tests/test_runner_number_with_unit.py
```diff
@@ -28,8 +28,9 @@ def test_number_with_unit_recognizer(
     for actual, expected in zip(results, expected_results):
         assert actual.text == expected['Text']
         assert actual.type_name == expected['TypeName']
-        assert actual.start == expected['Start']
-        assert actual.end == expected['End']
+        if 'Start' in expected and 'End' in expected:
```
tellarin, line 31: Same.

tellarin (changes requested):
Looks mostly good. Just some minor comments.
Are ypu planning to cotribute the other Unit types?
Thanks!

Python/tests/test_runner_number.py
ushmau5: > 

Yes there are many tests missing the `Start` or `End` values. I noticed this in Italian and German.

ushmau5:
> Looks mostly good. Just some minor comments. Are ypu planning to cotribute the other Unit types? Thanks!
Our client only supports a subset the entities in this library which is why there are only contributions to some entities.

Python/tests/test_runner_number.py
tellarin: Could you write a script to add the missing start/end in the specs instead? ;-)

tellarin:
> > Looks mostly good. Just some minor comments. Are ypu planning to cotribute the other Unit types? Thanks!
> 
> Our client only supports a subset the entities in this library which is why there are only contributions to some entities.

No worries. Thanks!

Python/tests/test_runner_number.py
ushmau5: I'll take a look into this

tellarin (approved):

PR #2915, opened 2022-04-07
[IT DateTimeV2] TillRegex update to support italian word "ad"

Sbardo93:
I met this issue using datetimeV2 prebuilt entity.
Without this change the utterance "da gennaio ad aprile" (from january to april) is not recognized, this is the result:
[{ "timex":["XXXX-01"], "type": "daterange"}, { "timex":["XXXX-04"], "type": "daterange"}]

The utterance "da gennaio a aprile" (same meaning) is correctly recognized as: [{ "timex":["(XXXX-01-01,XXXX-04-01,P3M)"], "type": "daterange"}]


tellarin (changes requested):
Such changes must be applied in the YAML file as the .cs files are autogenerated and this fix will be overwirtten.
Could you also add a spec test case in json to make sure regeressions don't happen in the future?
Thanks!

tellarin (approved):

tellarin (approved):
Piggybacked some other fixes in this.

Issue #2882, opened 2022-03-08
[PT Unit|Currency] Bugs reported from Speech 


nawanas:
The following patterns are not working
1-	 Having the “Value Currency Multiplier” doesn’t work (e.g. 75 USD milhoes)
2-	Having “CurrencySuffixList” number (e.g. dólar 22,5 milhões, and dólar d 225000000) doesn’t match
3-	“um” is a pattern for micrometer, while it is also the number 1
4-	Currency fractions (e.g. 15 USD 50) doesn’t match to $15.50 while “15 dólares 50” does

these are required to comprehensively match combinations of expressing currency

aitelint:
working on this

Issue #2889, opened 2022-03-09
[ES Currency] $$ USD Million not recognized as well as $$ MXN$ not recognized


nawanas:
1-	[Value] [Currency] {Units] such as 125 USD Millones, only the value and currency is detected and not the units (125 USD and not 125 Million USD)
2-	[Value] MXN$ is a missing pattern for the currency

aitelint:
working on this

PR #2916, opened 2022-04-08
[PT Unit|Currency] Fix bugs reported from Speech (#2882)

aitelint:
Fix to issue #2882

- Modified base code to support “Value Currency Multiplier” (e.g. 75 USD milhoes) in PT and EN
- Modified base code to better handle cases like "15 USD 50" (-> 15.50 USD) and "15 USD50" (-> 50 USD).

Test cases added to DimensionModel and CurrencyModel in PT and EN.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
@@ -723,7 +723,7 @@ namespace Microsoft.Recognizers.Definitions.Portuguese
             { @"Decímetro", @"dm|decimetro|decímetro|decimetros|decímetros" },
             { @"Centímetro", @"cm|centimetro|centímetro|centimetros|centimetros" },
             { @"Milímetro", @"mm|milimetro|milímetro|milimetros|milímetros" },
-            { @"Micrômetro", @"µm|micrometro|micrômetro|micrómetro|micrometros|micrômetros|micrómetros|micron|mícron|microns|mícrons|micra" },
+            { @"Micrômetro", @"µm|um|micrometro|micrômetro|micrómetro|micrometros|micrômetros|micrómetros|micron|mícron|microns|mícrons|micra" },
             { @"Nanômetro", @"nm|nanometro|nanômetro|nanómetro|nanometros|nanômetros|nanómetros|milimicron|milimícron|milimicrons|milimícrons" },
             { @"Picômetro", @"pm|picometro|picômetro|picómetro|picometros|picômetros|picómetros" },
             { @"Milha", @"mi|milha|milhas" },
@@ -738,7 +738,8 @@ namespace Microsoft.Recognizers.Definitions.Portuguese
             @"milha",
             @"milhas",
             @"""",
-            @"in"
+            @"in",
+            @"um"
         };
       public static readonly Dictionary<string, string> SpeedSuffixList = new Dictionary<string, string>
         {
@@ -825,7 +826,8 @@ namespace Microsoft.Recognizers.Definitions.Portuguese
         };
       public static readonly Dictionary<string, string> AmbiguityFiltersDict = new Dictionary<string, string>
         {
-            { @"\b\d+\s*\p{L}+$", @"((\d+(\s*\p{L}+[-—–-]|\p{L}+)\d+)|(((\p{L}|\d)[-—–-]\d+\s*|\p{L}\d+)\p{L}+))" }
+            { @"\b\d+\s*\p{L}+$", @"((\d+(\s*\p{L}+[-—–-]|\p{L}+)\d+)|(((\p{L}|\d)[-—–-]\d+\s*|\p{L}\d+)\p{L}+))" },
+            { @"\bum$", @"\p{L}\s+um\b" }
         };
       public static readonly Dictionary<string, string> TemperatureAmbiguityFiltersDict = new Dictionary<string, string>
         {
--- a/.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
@@ -633,9 +633,10 @@ namespace Microsoft.Recognizers.Definitions.Portuguese
             { @"Satoshi", @"SATOSHI" }
         };
       public const string CompoundUnitConnectorRegex = @"\b(?<spacer>e|com)\b";
+      public const string MultiplierRegex = @"\s*\b(mil(h([ãa]o|[õo]es))?|bilh([ãa]o|[õo]es)|trilh([ãa]o|[õo]es)|qua[td]rilh([ãa]o|[õo]es)|quintilh([ãa]o|[õo]es))\b";
       public static readonly Dictionary<string, string> CurrencyPrefixList = new Dictionary<string, string>
         {
-            { @"Dólar", @"$" },
+            { @"Dólar", @"$|dólar|dolar|dólares|dolares" },
             { @"Dólar estadunidense", @"us$|u$d|usd$|usd" },
             { @"Dólar do Caribe Oriental", @"ec$|xcd" },
             { @"Dólar australiano", @"a$|aud" },
@@ -722,7 +723,7 @@ namespace Microsoft.Recognizers.Definitions.Portuguese
             { @"Decímetro", @"dm|decimetro|decímetro|decimetros|decímetros" },
             { @"Centímetro", @"cm|centimetro|centímetro|centimetros|centimetros" },
             { @"Milímetro", @"mm|milimetro|milímetro|milimetros|milímetros" },
-            { @"Micrômetro", @"µm|um|micrometro|micrômetro|micrómetro|micrometros|micrômetros|micrómetros|micron|mícron|microns|mícrons|micra" },
+            { @"Micrômetro", @"µm|micrometro|micrômetro|micrómetro|micrometros|micrômetros|micrómetros|micron|mícron|microns|mícrons|micra" },
             { @"Nanômetro", @"nm|nanometro|nanômetro|nanómetro|nanometros|nanômetros|nanómetros|milimicron|milimícron|milimicrons|milimícrons" },
             { @"Picômetro", @"pm|picometro|picômetro|picómetro|picometros|picômetros|picómetros" },
             { @"Milha", @"mi|milha|milhas" },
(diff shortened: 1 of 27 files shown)


tellarin:
@aitelint Simple conflicts to resolve. Next PR in review stack.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
```diff
@@ -722,7 +723,7 @@ public static class NumbersWithUnitDefinitions
             { @"Decímetro", @"dm|decimetro|decímetro|decimetros|decímetros" },
             { @"Centímetro", @"cm|centimetro|centímetro|centimetros|centimetros" },
             { @"Milímetro", @"mm|milimetro|milímetro|milimetros|milímetros" },
-            { @"Micrômetro", @"µm|um|micrometro|micrômetro|micrómetro|micrometros|micrômetros|micrómetros|micron|mícron|microns|mícrons|micra" },
```
tellarin, line 725: This should not be removed as its sometimes used informally as the unit ("10 um", "mil um"). Perhaps adding it to the ambiguity list to be removed if not prefixed by a number.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
```diff
@@ -818,7 +819,7 @@ public static class NumbersWithUnitDefinitions
         };
       public static readonly Dictionary<string, string> AmbiguityFiltersDict = new Dictionary<string, string>
         {
-            { @"null", @"null" }
+            { @"\b\d+\s*\p{L}+$", @"((\d+(\s*\p{L}+[-—–-]|\p{L}+)\d+)|(((\p{L}|\d)[-—–-]\d+\s*|\p{L}\d+)\p{L}+))" }
```
tellarin, line 828: Can you explain this rule?

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/BaseMergedUnitExtractor.cs
```diff
@@ -216,5 +220,30 @@ private void MergePureNumber(string source, List<ExtractResult> ers)
 
             ers.Sort((x, y) => x.Start - y.Start ?? 0);
         }
+
+        // Add multiplier to extraction when it follows the unit e.g. "10 USD million"
+        private void MergeMultiplier(string source, List<ExtractResult> ers)
```
tellarin, line 225: Wouldn't it be best for perf to treat such cases as "special units", but with special resolution to convert back to the base currency? It can even sometimes show up as, for example, "MUSD" - as in "Direct support to the armed forces, totalling 50 MUSD, has been allocated in the budget".

Also, as far as I'm aware, such mention use is mostly common in finance, so it should probably be blocked for other units.

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
```diff
@@ -1584,7 +1584,7 @@
   },
   {
     "Input": "custa apenas 15 dólares 50.",
-    "NotSupported": "javascript, java",
+    "NotSupported": "javascript, java, python",
```
tellarin, line 1587: Why is this now broken in Python?

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
```diff
@@ -1746,5 +1746,186 @@
...
+          "value": "22500000",
+          "unit": "Dólar"
+        },
+        "Start": 37,
+        "End": 54
+      }
+    ]
+  },
+  {
+    "Input": "Los precios del acuerdo comienzan en dólar 22500000.",
```
tellarin, line 1784: Same

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
```diff
@@ -1746,5 +1746,186 @@
...
+          "unit": "Dólar estadunidense",
+          "isoCurrency": "USD"
+        },
+        "Start": 4,
+        "End": 12
+      }
+    ]
+  },
+  {
+    "Input": "Eles pagaram 75 USD milhões, mas o preço na verdade foi de 50 milhões USD. E se tivessem mais 15 USD 50, teriam pago 80 USD milhões.",
```
tellarin, line 1833: Is "50 milhões de dólares" also supported?

tellarin (changes requested):

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
aitelint: ok, I will add a filter to allow only cases where is preceded by numbers in digits. 
Being a number, it can still be common to have it prefixed by numbers without being a unit. I added such a test case already ("três dois um"), and it can appear in expressions like "vinte um", "cinqüenta um"....

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
aitelint: it was added in English to filter expressions like "24703 L-202". I thought it could be useful in Portuguese too, but is not required by this PR, so it can be removed.

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
aitelint: Because we added "dólares" to the prefix list in order to extract patterns like "dólarer 22,5 milhões" and then modified the method SelectCandidates in .NET to correctly handle cases like "15 dólares 50", "3 $50"... 
But the method is missing in Python, so now it returns "dólares 50" instead  of "15 dólares 50".

aitelint: I could add SelectCandidates in Python if necessary.

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
```diff
@@ -1746,5 +1746,186 @@
...
+          "unit": "Dólar estadunidense",
+          "isoCurrency": "USD"
+        },
+        "Start": 14,
+        "End": 27
+      }
+    ]
+  },
+  {
+    "Input": "Los precios del acuerdo comienzan en dólar 22,5 milhões.",
```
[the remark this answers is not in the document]
aitelint: ok, I will modify the spec. 
Yes, I was planning to add this pattern in Spanish when addressing #2889, after this PR is merged.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
tellarin: No need to remove. I was just considering why as I couldn't understand it jsut reading the regexes. :)

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
tellarin: In "vinte um" and "cinqüenta um" it's part of the number. But you're right that in "três dois um" it's a problem. Hmmm.

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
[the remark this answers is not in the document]
tellarin: Adding the Spanish patterns in #2889 is fine. Thanks.

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
tellarin: If that doesn't take you much time and block other work, please do. Thanks.

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/BaseMergedUnitExtractor.cs
aitelint: mmh, if I understand correctly, the main difference would be to replace the regex search with the TrieTree search.
But since the multiplier regex is small, I do not think the perf impact is noticeable. 
I will try to add the multipliers as special units to the CurrencySuffixList and check.

Yes, that is already the case. MergeMultiplier is inside MergeCompoundUnits which is called only for Currency units.

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
aitelint: yes, added case

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
aitelint: modified

Specs/NumberWithUnit/Portuguese/CurrencyModel.json
aitelint: done

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/BaseMergedUnitExtractor.cs
aitelint: Treating them as special units (where each suffix unit is combined with each multiplier in a new suffix dictionary) increases execution time by 28%.
Using the current approach with regex, the impact is 0.87% (comparable to fluctuations between different executions: 0.84%).
One could also treat only the multipliers as special units but then the processing would be the same as the one already implemented and there would be little difference in performance.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/NumbersWithUnitDefinitions.cs
aitelint: added support for "10 um" but not "mil um".

.NET/Microsoft.Recognizers.Text.NumberWithUnit/Extractors/BaseMergedUnitExtractor.cs
tellarin: Thanks for checking this. Seems good to me then. We'll continue the review soon.

tellarin (approved):
Approving.
But cases like "Direct support to the armed forces, totalling 50 MUSD, has been allocated in the budget" are still not supported. Please open an issue for it in EN and mark as P3.

Issue #2945, opened 2022-05-09
[* DateTimeV2] "before" no longer recognized in multi mentions like "from 2010 to 2018 or before 2000"


gooooloo:
**Describe the bug**
"before" keyword is not recognized in "from 2010 to 2018 or before 2000".

**To Reproduce**
1. ask "show me sales from 2010 to 2018 or before 2000" using version 1.8.0 or 1.8.3
2. result doesn't recognize "before" as datetime range:
![image](https://user-images.githubusercontent.com/1883665/167330834-314e8934-7fe8-4a5a-9437-890c4e984214.png)

**Expected behavior**
should recognize "before" as datetie range, as version 1.7.0.
![image](https://user-images.githubusercontent.com/1883665/167331006-52108dea-8f6c-49dd-90a1-a0dbc97a58d9.png)

**Platform (please complete the following information):**
 - Platform:  .NET
 - Environment: NuGet
 - Version of package 1.8.0, 1.8.3

tellarin:
If reproduced, this is a silent regression that need to be fixed soon. @aitelint, please add it to your list and make sure it works across languages.

gooooloo:
@tellarin @aitelint I tried release 1.8.4 locally and still see this error. But I do see a [test case](https://github.com/microsoft/Recognizers-Text/pull/2947/files#diff-2fa3b2e50c2839d8edde2c9bf727915ab70a33d8f3eeefbf11ad107803f4b5be) in the PR. So I am confused. Does the release 1.8.4 contain the fix inside? Thanks for your help.

aitelint:
@gooooloo, expressions like "from 2010 to 2018 or before 2000" should work fine now, but maybe other patterns are still buggy. What sentence did you entered?

gooooloo:
@aitelint it is "show me sales from 2010 to 2018 or before 2000". It is one of our test cases. Does it matter that it has a "show me sales" prefix?

gpailler:
@tellarin @aitelint I'm confused too. It seems that the version 1.8.4 published on Nuget doesn't include this PR (even if this PR is mentioned in the release notes).

For instance:
- `YearPeriodRegex` is changed in french in the PR (https://github.com/microsoft/Recognizers-Text/pull/2947/files#diff-23e726b1baef0452fb6160766443bac375fd170efe124c463beaaaf6c8ed3b61)
[...]

PR #2947, opened 2022-05-09
[* DateTimeV2] Fix "before" no longer recognized in multi mentions like "from 2010 to 2018 or before 2000" (#2945)

aitelint:
Fix to issue #2945.

Test cases added to EN and NL DateTimeModels. 

The other languages were not affected because they do not use the mechanism responsible for the issue.


tellarin (changes requested):
Looks OK to me.
Can you add specs for the other languages too, to make sure future regressions are caught?

aitelint:
> Looks OK to me. Can you add specs for the other languages too, to make sure future regressions are caught?

specs added

tellarin (approved):

tellarin:
@aitelint I've merged it as it already brings good value and solve the specific reported regression in EN.
But please create a new PR with specs for the missing languages, ok? Especially ZH/JA/KO, and if possible HI/TR. 
If some of them don't support this case, mark the spec as such and we can decide how to prioritize needed fixes.
Thanks!

PR #2950, opened 2022-05-13
[JA DateTimeV2] Merged refinements

aitelint:
All test cases pass.

Added support for the 30-hour clock form (commonly used in Japanese) and relative test cases in DateTimeParser, DateTimePeriodParser, MergedParser, TimeParser and TimePeriodParser.

Modified test cases:
In DateParser: 
- "私は今から2週間以内に戻ります" (I will be back within 2 weeks from now), moved to DatePeriodParser and replaced with "私は今から2週間でに戻ります" (I will be back in 2 weeks from now).

In MergedExtractor:
- "これは2日です。" (This is 2 days), modified type from Duration to Date because it is more commonly interpreted as 2nd day of the month. Added new test case "これは2日間です。" which unambiguously indicates a duration.
- "午前10時の予定を4.3時に変更します。" -> "午前10時の予定を4.3に変更します。" (I'll change the ten am appointment to 4.3), "4.3時" (4.3 o'clock) was used which is not a valid time expression in Japanese. Modified to remove "o'clock" (時).
- Added extraction of time entities when missing:
    - "午前10時の予定を26時に変更します。" (I'll change the 10am appointment to 26 o'clock), 26時 is extracted (valid time in Japanese).
    - "午前10時の予定を25時に変更します。" (I'll change the 10am appointment to 25 o'clock), 25時 is extracted (valid time in Japanese).
    - "午前10時の予定を4時以降に変更します。" (I'll change the ten am appointment to 4 or later), "4時" (4 o'clock) is used which is extracted. Added new test case "午前10時の予定を4以降に変更します。" where "4" is not extracted.


.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateTimePeriodParser.cs
```diff
@@ -126,15 +126,31 @@ private DateTimeResolutionResult MergeDateAndTimePeriod(string text, DateObject
...
+            {
+                pastDate = pastDate.AddDays(1);
+                futureDate = futureDate.AddDays(1);
+            }
+            else if (timexHours.Item2 > Constants.DayHourCount)
+            {
+                swiftDay++;
+            }
+
+            var pastDate1 = pastDate.AddDays(swiftDay);
```
tellarin, line 142: Maybe rename to pastDateAlt as it's an alternative resolution?

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -569,6 +569,30 @@ public static float ParseNumberFromDurationTimex(string timex)
             return float.Parse(numberStr);
         }
 
+        public static int ParseHourFromTimeTimex(string timex)
+        {
+            var start = timex.IndexOf(Constants.TimeTimexPrefix) + 1;
+            var end = timex.IndexOf(Constants.TimeTimexConnector);
+            end = end > 0 ? end : timex.Length;
+            var hourStr = timex.Substring(start, end - start);
```
tellarin, line 577: From now on, let's try to write such functions using Spans, to avoid creating a lot of small string objects garbage collected all the time. Please open an issue to refactor existing Timex parsing code to do the same.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
```diff
@@ -569,6 +569,30 @@ public static float ParseNumberFromDurationTimex(string timex)
...
+            int.TryParse(hourStr, out int hour);
+
+            return hour;
+        }
+
+        public static Tuple<int, int> ParseHoursFromTimePeriodTimex(string timex)
+        {
+            int hour1 = 0, hour2 = 0;
+            var timeList = timex.Split(Constants.TimexSeparator[0]);
```
tellarin, line 586: Same.

tellarin (changes requested):
Mostly good. One minor comment and a request for future code for perf.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDateTimePeriodParser.cs
aitelint: done

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
tellarin: @aitelint Please ACK this comment. Thanks.

tellarin (approved):

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TimexUtility.cs
aitelint: @tellarin, issue created

aitelint: @tellarin, the problem is that net462 does not support parsing of spans, so we would need to convert the span back to string when passing it to int.TryParse

PR #2963, opened 2022-05-27
[NL DateTimeV2] Fix multiple issues from speech across date/time sub-types (#2898)

aitelint:
Fix to issue #2898.

Pattern 4 (Replacing 1 with een in “week 1” is not detected, also the expression “Het gebeurde door een tijdsverschil van een seconde” both “eens” are resolves to the number 1 while only the latter is a number.) has not been addressed because "week one" is not supported by design also in English and 'een' number is indistinguishable from 'een' article.

Test cases added or enabled in DateTimeModel.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Parsers/DutchSetParserConfiguration.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Parsers/DutchSetParserConfiguration.cs
@@ -31,6 +31,12 @@ namespace Microsoft.Recognizers.Text.DateTime.Dutch
         private static readonly Regex YearTypeRegex =
             new Regex(DateTimeDefinitions.YearTypeRegex, RegexFlags);
 
+        private static readonly Regex SemiYearTypeRegex =
+            new Regex(DateTimeDefinitions.SemiYearTypeRegex, RegexFlags);
+
+        private static readonly Regex WeekendTypeRegex =
+            new Regex(DateTimeDefinitions.WeekendTypeRegex, RegexFlags);
+
         public DutchSetParserConfiguration(ICommonDateTimeParserConfiguration config)
             : base(config)
         {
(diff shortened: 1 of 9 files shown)


.NET/Microsoft.Recognizers.Text.DateTime/Dutch/Parsers/DutchSetParserConfiguration.cs
```diff
@@ -125,9 +131,17 @@ public bool GetMatchedDailyTimex(string text, out string timex)
             {
                 timex = "P1Y";
```
tellarin, line 132: Let's move these timex string generations to common timex code.

tellarin: This can come in a follow up PR.

tellarin (approved):

tellarin:
@aitelint, please create an issue to support cases of weeks with spelled out numbers, like "week one", across languages.
Remind me in case this was discussed previously and there was a decision to not support it.

aitelint:
> @aitelint, please create an issue to support cases of weeks with spelled out numbers, like "week one", across languages. Remind me in case this was discussed previously and there was a decision to not support it.

Ok. 
I found only a comment in #2414 about the pattern ordinal + week ("first week", "1st week") that has not been implemented so far.
("first week of 2019", "first week of April" are supported).
Should we also address it?

tellarin:
Please do.

aitelint:
> Please do.

After some more checks the pattern ordinal + week seems ambiguous in resolution. 
Most of the occurrences refer to context dependent starting points: "13th week of pregnancy", "second week of training"...
Should we resolve them similarly to relative ordinal numbers? 
E.g. "second week" --> "timex": "W02", "start": "START_DATE+7", "end": "START_DATE+14"
Or do you have other indications?

PR #2971, opened 2022-06-05
Integration of new mode termed as TasksMode in DatetimeOption.cs

Neelisha-saxena:
**TasksMode**: When the input text is in the form of tasks (i.e., Schedule 1:1 call on Monday), this mode modifies the value of datetime reference of the input text.

Implemented four-digit number handling rule under Tasksmode:
Should not treat a four-digit number as a "daterange" if the input text does not include other reference of month or year. It should not treat 2005 as a daterange in statements like "Total no of milk packet is 2005."


Neelisha-saxena:
Fixing of some issues is left

PR #2972, opened 2022-06-05
[.NET] Integration of new mode termed "TasksMode" in DatetimeOption.cs

Neelisha-saxena:
**TasksMode**: When the input text is in the form of tasks (i.e., Schedule 1:1 call on Monday), this mode modifies the value of datetime reference of the input text.

Implemented four-digit number handling rule under Tasksmode:
Should not treat a four-digit number as a "daterange" if the input text does not include other reference of month or year. It should not treat 2005 as a daterange in statements like "Total no of milk packet is 2005."


.NET/Microsoft.Recognizers.Text.DateTime/DateTimeOptions.cs
```diff
@@ -38,6 +38,11 @@ public enum DateTimeOptions
         /// </summary>
         NoProtoCache = 16,
 
+        /// <summary>
+        /// NoProtoCache
```
tellarin, line 42: Please update comments

tellarin: I'd also personally prefer to use 2^20 as the value.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -270,6 +278,15 @@ private bool ShouldSkipFromToMerge(ExtractResult er)
             return config.FromToRegex.IsMatch(er.Text);
         }
 
+        /*Under TasksMode: Should not treat a four-digit number as a daterange if the input text does not include a month or year reference.
+          It should not treat 2005 as a daterange in statements like "Milk 2005."
+         (The year 2005 should be treated as a number only.)
+        */
+        private bool ShouldSkipOnlyYear(ExtractResult er)
+        {
+            return config.YearRegex.Match(er.Text).Value == er.Text;
```
tellarin, line 287: Shouldn't this be a base regex as it doesn't change across cultures?

Specs/DateTime/English/DateTimeModelTasksMode.json
```diff
@@ -0,0 +1,88 @@
+[
+  {
+    "Input": "How about year 2016 and greater",
```
tellarin, line 3: Ideally such new spec cases should also be added in the regular DateTimeModel to show the default behaviour there was not affected.
Also, "Java" is missing in the NotSupported tags here.

tellarin (changes requested):
It seems this PR is also altering the default behaviour (without the new flag), which should not happen. 
No need to close the PR and open a new one, you can just push the changes to the branch that originated the PR.
Thanks.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
tellarin: I see, you're also filtering written numbers...

Specs/DateTime/English/DateTimeModelTasksMode.json
Neelisha-saxena: Okay, I will update test json files.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeOptions.cs
Neelisha-saxena: Any specific reason for using 2^20 value.

tellarin: It's the highest available value.

tellarin:
@Neelisha-saxena Build passed. So now only the new mirror DateTimeModel.json specs are missing.
For the case(s) where the default mode produce false positives (like in "2000 pieces"), mark it as NotSupported in .NET
and add a Comment attribute with a message saying the default more produces a false positive for it and that it's a case from TasksMode. 
Once we have the specs, we can merge this.
Thanks!

Neelisha-saxena:
@tellarin, I have made changes as asked please review it.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -18407,7 +18407,7 @@
     "Context": {
       "ReferenceDateTime": "2019-12-15T01:00:00"
     },
-    "NotSupported": ".NET, javascript, python, java",
+    "NotSupported": "dotnet, javascript, python, java",
     "Results": [
       {
         "Text": "2000",
```
tellarin, line 18413: This should be empty as the real result should be nothing.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeOptions.cs
tellarin: The file doesn't seem updated here.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseMergedDateTimeExtractor.cs
```diff
@@ -270,6 +278,15 @@ private bool ShouldSkipFromToMerge(ExtractResult er)
             return config.FromToRegex.IsMatch(er.Text);
         }
 
+        /*Under TasksMode: Should not treat a four-digit number as a daterange if the input text does not include a month or year reference.
+          It should not treat 2005 as a daterange in statements like "Milk 2005."
+         (The year 2005 should be treated as a number only.)
+        */
+        private bool ShouldSkipOnlyYear(ExtractResult er)
```
tellarin, line 285: No need to do this in this PR, but it would be interesting if this mode-specific behaviours/code is isolated into it's own module/strategy.
Please move it along with new changes in a followup PR.

tellarin (changes requested):
Minor changes left.

tellarin (approved):
I've editted the two remaining files.

Issue #2964, opened 2022-05-30
[* DateTimeV2] Add support for patterns like "week one"


aitelint:
Add support for the pattern
"week" + [WrittenNumber]

e.g. "Show sales in the week twelve",

Expressions with numbers in digit form like "week 12" are already supported.

aitelint:
working on this

PR #2978, opened 2022-06-08
[* DateTimeV2] Added support for patterns like "week one" (#2964)

aitelint:
Fix to issue #2964.

Added support for "week one" in EN and NL.
Added support for "first week" in EN (where, if the year is missing, the start/end dates are calculated from the reference date).

Test cases added to EN and NL DateTimeModels.


tellarin (changes requested):
Support for week + written number seems OK.
But why add suppoer for ordinal + week w/o context in this way in English? Such cases without context need a different timex.

aitelint:
> But why add suppoer for ordinal + week w/o context in this way in English? Such cases without context need a different timex.

I left a comment in PR [2963](https://github.com/microsoft/Recognizers-Text/pull/2963#issuecomment-1140732941) when we were talking about this pattern, but since there was no answer I followed this approach. 

What timex should they have?

Issue #2968, opened 2022-05-30
[EN Number] Support "nought" as 0


tellarin:
### Discussed in https://github.com/microsoft/Recognizers-Text/discussions/2928

<div type='discussions-op-text'>

<sup>Originally posted by **ktjbrowne** April 20, 2022</sup>
Suggest to add recognizing 'nought' as zero/0
Thanks.
K.</div>

PR #2982, opened 2022-06-13
[EN Number] Added support for "nought" as 0 (#2968)

aitelint:
Fix to issue #2968.

Test cases added to NumberModel.


tellarin (changes requested):
LGTM, but we should probably include "naught" too.

tellarin (approved):

PR #2989, opened 2022-06-15
[SAMPLE] Update Azure Function to v4 and NET 6 (LTS)

iMicknl:
- Updated the Azure Function sample to .NET 6 (LTS) and thus Az Func v4
- Remove FxCopAnalyzers for this csproj, since .NET5+ has this in by default and FxCopAnalyzers is deprecated
- Update host.json and local.settings.json with the default Azure Functions values


iMicknl:
Seems that this PR will be blocked by AppVeyor currently 🙁, since .NET 6 requires MSBuild 2022 / VS 2022.

tellarin (changes requested):
Thanks for the PR. We're preparing for the move to .NET 6 and we'll merge as soon as possible.
Marking as pending changes only to block merge momentarily.

Issue #2960, opened 2022-05-23
[* DateTimeV2] I case of YYYY entry presence in the parsed text the recognizer sometimes outputs incorrect results


Kajman55:
**Bug description**
In case the parsed text contains year value in the form YYYY (perhaps this is not an ehaustive case), the recognizer output deviates from the expected one.

**To Reproduce**
1. Setup the following input string for the recognizer:
 
            string sourceText = "DateTime search phrase\n23.04.2022 14:55\n30.04.2022 12:04";

2. Execute datetime recognizer as below:

            var msRecognizerResult = DateTimeRecognizer.RecognizeDateTime(sourceText, "de-de").ToList();

            Console.WriteLine();
            foreach (var rr in msRecognizerResult)
            {
                Console.WriteLine(rr.Text);
            }

            Console.ReadKey();

3. The Console will output:

           23.04.2022 14:55 
           30.04.2022 12:04

4. Repeat the same procedure but with the following input text:

           string sourceText = "DateTime search phrase\n23.04.2022 14:55\n30.04.2022 12:04 some text 2020";

5. The Console will output:

           2020

**Expected behavior**
In the second case the output should be:

           23.04.2022 14:55 
           30.04.2022 12:04
           2020

**Platform:**
 - Platform: .NET Framework 4.7.2
 - Environment: nuget package Microsoft.Recognizers.Text.DateTime
 - Version of package:  ver 1.8.4

**Additional context**
The issue might be connected with the following code place:

Class:         Microsoft.Recognizers.Text.DateTime.BaseMergedDateTimeExtractor
Method:    List\<ExtractResult\> FilterAmbiguity(List<ExtractResult> extractResults, string text) line 290 

changed in the commit 927c06361764d485bdabd0e0eb47694a08e40567

The new code modifies the 'extractResults' collection during its iteration.

aitelint:
working on this

PR #2990, opened 2022-06-16
[* DateTimeV2] Fixed I case of YYYY entry presence in the parsed text the recognizer sometimes outputs incorrect results (#2960)

aitelint:
Fix to issue #2960.

(Pushed on top of #2983 to take into account refactoring implemented there)

Also, moved method FilterAmbiguity to ExtractResultExtension since is used in multiple entities.

### DIFF
--- a/Specs/DateTime/English/DateTimeModel.json
+++ b/Specs/DateTime/English/DateTimeModel.json
@@ -23834,5 +23834,133 @@
         }
       }
     ]
+  },
+  {
+    "Input": "DateTime search phrase 23.04.2022 14:55",
+    "Context": {
+      "ReferenceDateTime": "2016-11-07T00:00:00"
+    },
+    "NotSupported": "java, javascript, python",
+    "Results": [
+      {
+        "Text": "23.04.2022 14:55",
+        "Start": 23,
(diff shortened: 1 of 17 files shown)


Specs/DateTime/English/DateTimeModel.json
```diff
@@ -23722,5 +23738,229 @@
...
+              "type": "datetime",
+              "value": "2022-04-30 00:04:00"
+            }
+          ]
+        }
+      }
+    ]
+  },
+  {
+    "Input": "DateTime search phrase\n23.04.2022 14:55\n30.04.2022 12:04 some text 2020",
```
tellarin, line 23907: Could we add specs like this to a couple of the other languaes to make sure the issue doesn't happen there?

tellarin (changes requested):
Looks OK. Some extra specs would be helpful.

Specs/DateTime/English/DateTimeModel.json
aitelint: added in NL|FR|DE|IT|PT|ES

tellarin (approved):

PR #2991, opened 2022-06-16
[Python Date] Fix for cases like 'Monday 21'

kevinwalshgen:
This is to bring Python in line with the .NET implementation for handling cases like 'Monday 21'


tellarin (changes requested):
Closing this at is seems superseded by #3015. Also, when adding specs, please add the NotSupported tag for the platforms that don't support the new behavious. And it's also good practice to add a spec case at the Model level too, beyond just Extractor/Parser (i.e., in DateTimeModel.json).

Issue #2985, opened 2022-06-14
[DE DaTeTimeV2] German date recognition inconsistent if containing ordinal numbers as words


martinmueller4voice:
**German dates not recognized if day or month number given as word**

**To Reproduce**
- `DateTimeRecognizer.RecognizeDateTime("siebter April", "de-DE")` returns timex XXXX-04-07
- `DateTimeRecognizer.RecognizeDateTime("7. April", "de-DE")` returns timex XXXX-04-07
-  `DateTimeRecognizer.RecognizeDateTime("7.4.", "de-DE")` returns timex XXXX-04-07
- `DateTimeRecognizer.RecognizeDateTime("siebter 4.", "de-DE")` does **not** match
- `DateTimeRecognizer.RecognizeDateTime("7. vierter", "de-DE")` does **not** match

**Expected behavior**
Dates should be recognized without having to convert cardinal/ordinal numbers into digits first.

**Platform (please complete the following information):**
 - Platform: .NET
 - Environment: nuget package
 - Version of package v1.8.4

martinmueller4voice:
Perfect! Thanks again!

PR #3001, opened 2022-06-27
[DE DateTimeV2] Fixed German date recognition inconsistent if containing ordinal numbers as words (#2985)

aitelint:
Fix to issue #2985.


tellarin (approved):

martinmueller4voice:
Great - thanks for fixing it.
The only thing I noticed: Will it work if you're using the correct cases for the ordinal or cardinal words as well?
I only saw the Nominativ cases in the tests/specs ("Ich komme am Donnerstag siebter 4. zurück.") and the regex definitions.
Don't you have to add cases ending in "-en" as well? ("Ich komme am Donnerstag, den siebten vierten zurück")?

PR #3009, opened 2022-07-18
[EN DateTimeV2] Extended recurrence support in extractor

Neelisha-saxena:


.NET/Microsoft.Recognizers.Definitions.Common/English/DateTimeDefinitions.cs
```diff
@@ -918,6 +918,8 @@ public static class DateTimeDefinitions
       public const string DoubleMultiplierRegex = @"^(bi)(-|\s)?";
       public const string HalfMultiplierRegex = @"^(semi)(-|\s)?";
       public const string DayTypeRegex = @"((week)?da(il)?ys?)$";
+      public const string WeekDayTypeRegex = @"(weekday?)$";
+      public const string FortNightTypeRegex = @"(fortnight?)$";
```
tellarin, line 922: Should be "Fortnight".

.NET/Microsoft.Recognizers.Definitions.Common/Dutch/DateTimeDefinitions.cs
```diff
@@ -239,7 +239,7 @@ public static class DateTimeDefinitions
       public const string EachPrefixRegex = @"\b(?<each>(iedere|elke|eenmaal per)\s*$)";
       public static readonly string SetEachRegex = $@"\b(?<each>(iedere|elke|om\s+de)\s*(?<other>\s+andere)?\s*(week\s*(?={WeekDayRegex}))?)";
       public const string SetLastRegex = @"(?<last>volgende?|komende|aankomende|aanstaande|deze|huidige|aanstaande|vorige?|verleden|laatste)";
-      public const string EachDayRegex = @"^\s*(iedere|elke)\s*dag\b";
```
tellarin, line 242: Why remove all these string start markers?

.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexParsing.cs
```diff
@@ -52,15 +52,33 @@ private static void ExtractDateTime(string s, TimexProperty timexProperty)
 
             if (indexOfT == -1)
             {
+                var indexOfP = s.IndexOf('P');
```
tellarin, line 55: Why didn't this happen before? Can you add a code comment explaining cases that fall here?
Also, change the hardcoded 'P' to reuse a constant.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishSetExtractorConfiguration.cs
```diff
@@ -29,6 +29,9 @@ public class EnglishSetExtractorConfiguration : BaseDateTimeOptionsConfiguration
         public static readonly Regex EachDayRegex =
             new Regex(DateTimeDefinitions.EachDayRegex, RegexFlags);
 
+        public static readonly Regex BeforeEachDayRegex =
```
tellarin, line 32: Why give it a different name? I see in other languages it is its own regex. 
If the reuse of DayRegex is correct, just add a code comment here explaining.

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishSetParserConfiguration.cs
```diff
@@ -125,7 +131,11 @@ public bool GetMatchedDailyTimex(string text, out string timex)
                 multiplier = 0.5f;
             }
 
-            if (DayTypeRegex.IsMatch(trimmedText))
+            if (WeekDayTypeRegex.IsMatch(trimmedText))
+            {
+                durationType = "WD";
```
tellarin, line 136: Can you change this if/else chain to reuse constants for the types?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
```diff
@@ -80,7 +83,24 @@ public List<Token> MatchEachUnit(string text)
             matches = this.config.EachUnitRegex.Matches(text);
             foreach (Match match in matches)
             {
-                ret.Add(new Token(match.Index, match.Index + match.Length));
+                if (match.Groups["unit"].Value.Equals("month"))
```
tellarin, line 86: Which new cases now fall here?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
```diff
@@ -91,25 +111,106 @@ public virtual List<Token> TimeEveryday(string text, DateObject reference)
...
+            if ((config.Options & DateTimeOptions.TasksMode) != 0)
+            {
+                var ers1 = this.config.TimePeriodExtractor.Extract(text, reference);
+                if (ers.Count == 0 && ers1.Count == 1)
+                {
+                    ers = ers1;
+                }
+            }
+
+            foreach (var er in ers)
```
tellarin, line 123: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
```diff
@@ -91,25 +111,106 @@ public virtual List<Token> TimeEveryday(string text, DateObject reference)
...
+                if (match.Success)
+                {
+                    ret.Add(new Token(er.Start ?? 0, (er.Start + er.Length ?? 0) + match.Length + match.Index));
+                }
+            }
+
+            return ret;
+        }
+
+        public virtual List<Token> DayEveryweek(string text, DateObject reference)
```
tellarin, line 145: Please add a code comment describing this method.

tellarin: Also, is there negative performance impact for this new method?

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
```diff
@@ -91,25 +111,106 @@ public virtual List<Token> TimeEveryday(string text, DateObject reference)
...
+            }
+
             foreach (var er in ers)
             {
                 var afterStr = text.Substring(er.Start + er.Length ?? 0);
-                if ((string.IsNullOrEmpty(afterStr) || this.config.CheckBothBeforeAfter) && this.config.BeforeEachDayRegex != null)
+
+                var beforeStr = text.Substring(0, er.Start ?? 0);
+
+                // var beforeMatch = this.config.EachUnitRegex.Match(beforeStr);
```
tellarin, line 165: Please delete the commented out code if not necessary.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -32,61 +33,78 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
 
             if (er.Type.Equals(ParserName, StringComparison.Ordinal))
             {
-                var innerResult = ParseEachUnit(er.Text);
+                var innerResult = ParseEachUnit(er.Text.Trim());
```
tellarin, line 36: Why are these trims needed? If really necessary, trim should be called only once.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -32,61 +33,78 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
...
 
                 if (!innerResult.Success)
                 {
-                    innerResult = ParseEach(config.TimeExtractor, config.TimeParser, er.Text, refDate);
+                    innerResult = ParseEach(config.TimeExtractor, config.TimeParser, er.Text.Trim(), refDate);
+                }
+
+                if (!innerResult.Success)
+                {
+                    innerResult = ParserDayEveryweek(er.Text.Trim(), refDate);
```
tellarin, line 82: ParserDayEveryWeek?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -32,61 +33,78 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
...
+                }
+
+                if (!innerResult.Success)
+                {
+                    innerResult = ParserSingleNumbermonth(er.Text.Trim(), refDate);
```
tellarin, line 87: ParserSingleNumberMonth?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -190,20 +223,143 @@ private DateTimeResolutionResult ParserTimeEveryday(string text, DateObject refD
             var ret = new DateTimeResolutionResult();
 
             var ers = this.config.TimeExtractor.Extract(text, refDate);
+            var ers1 = this.config.TimePeriodExtractor.Extract(text, refDate);
+
+            if (ers.Count == 1)
```
tellarin, line 228: Which cases now fall here?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -190,20 +223,143 @@ private DateTimeResolutionResult ParserTimeEveryday(string text, DateObject refD
...
+                var match = this.config.EachDayRegex.Match(afterStr);
+
+                if (match.Success)
+                {
+                    var pr = this.config.TimeParser.Parse(ers[0], DateObject.Now);
+                    ret = SetHandler.ResolveSet(ref ret, pr.TimexStr);
+
+                    if ((config.Options & DateTimeOptions.TasksMode) != 0)
+                    {
+                        ret = TasksModeSetHandler.TasksModeResolveSet(ref ret, pr.TimexStr + "P1D");
```
tellarin, line 240: Please avoid hardcoded timex-related values. Reuse constants or methods to build expressions.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -190,20 +223,143 @@ private DateTimeResolutionResult ParserTimeEveryday(string text, DateObject refD
...
+            var ers1 = this.config.TimePeriodExtractor.Extract(text, refDate);
+
+            if (ers.Count == 1)
+            {
+                var afterStr = text.Replace(ers[0].Text, string.Empty);
+                var match = this.config.EachDayRegex.Match(afterStr);
+
+                if (match.Success)
+                {
+                    var pr = this.config.TimeParser.Parse(ers[0], DateObject.Now);
```
tellarin, line 235: Why passing "now" and not the existing reference time?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -190,20 +223,143 @@ private DateTimeResolutionResult ParserTimeEveryday(string text, DateObject refD
...
+                }
+            }
+            else if (ers1.Count == 1)
+            {
+                var afterStr = text.Replace(ers1[0].Text, string.Empty);
+                var match = this.config.EachDayRegex.Match(afterStr);
+
+                if (match.Success)
+                {
+                    var pr = this.config.TimePeriodParser.Parse(ers1[0], DateObject.Now);
```
tellarin, line 252: Same

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -230,12 +386,11 @@ private DateTimeResolutionResult ParseEach(IDateTimeExtractor extractor, IDateTi
             }
 
             // remove suffix 's' and "on" if existed and re-try
-            match = this.config.SetWeekDayRegex.Match(text);
-            if (match.Success)
+            var match1 = this.config.SetWeekDayRegex.Match(text);
```
tellarin, line 389: Rename to better variable names. To differentiate from other "match".

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/MergedParserUtil.cs
```diff
@@ -397,6 +398,16 @@ public static string DetermineSourceEntityType(string sourceType, string newType
             var resolutionPast = GenerateResolution(type, pastResolutionStr, mod);
             var resolutionFuture = GenerateResolution(type, futureResolutionStr, mod);
 
+            if (type.Equals(Constants.SYS_DATETIME_SET, StringComparison.Ordinal) && (config.Options & DateTimeOptions.TasksMode) != 0)
```
tellarin, line 401: It's better to check for the mode flag first, as much as possible. This makes the code easier to read and potentially saves a few cycles in the default mode skipping unnecessary checks there.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            return result;
+        }
+
+        public static Dictionary<string, string> TasksModeGenerateResolutionSetParser(Dictionary<string, string> resolutionDic, string mod, string timex)
+        {
+            var res = new Dictionary<string, string>();
+
+            TasksModeAddAltSingleDateTimeToResolution(resolutionDic, TimeTypeConstants.DATETIMEALT, mod, res);
+            if (timex.StartsWith("P") && res.Count > 0)
```
tellarin, line 42: Please avoid hardcoding values and timex-related terms. Reuse constants or methods that manipulate expressions.

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -16217,6 +16217,7 @@
     "Context": {
       "ReferenceDateTime": "2020-06-12T00:00:00"
     },
+    "NotSupported": "dotnet",
```
tellarin, line 16220: This should never happen. All .NET cases must continue to pass.
What happens here?

Specs/DateTime/English/DateTimeModel.json
```diff
@@ -18684,36 +18685,22 @@
...
+        "Text": "every day at 7:13 p.m.",
         "Start": 11,
-        "End": 19,
+        "End": 32,
         "TypeName": "datetimeV2.set",
         "Resolution": {
           "values": [
             {
-              "timex": "P1D",
+              "timex": "T19:13",
```
tellarin, line 18698: Hmmm. Shouldn't the new timex include both terms?

tellarin: I'm ok with it as the recurrence is kind of implicit. "every time it is 19:13".
But I worry about other cases like "every workday at 19:13".

Specs/DateTime/German/DateTimeModel.json
```diff
@@ -2744,6 +2744,7 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
+    "NotSupported": "dotnet",
```
tellarin, line 2747: This can't happen.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
Neelisha-saxena: removed unused code

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: removed hardcoded constants.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: added new variable in tasksmodeconstant file.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/MergedParserUtil.cs
Neelisha-saxena: code modified as per suggestion.

Specs/DateTime/German/DateTimeModel.json
Neelisha-saxena: A similar new test case for which dotnet is supported is added at line 5622, as dotnet output is modified.
particular dotnet change is not reflected to other languages.

Specs/DateTime/English/DateTimeModel.json
Neelisha-saxena: A similar new test case for which dotnet is supported is added at line 25754, as dotnet output is modified.
particular dotnet change is not reflected to other languages.

Specs/DateTime/English/DateTimeModel.json
Neelisha-saxena: Earlier version of set parser doesn't support cases like "every day at 5pm" as single entity, there were two different entities recognized with timex as set P1D and T19:30. So in default mode I have kept only T19:30 as timex for set resolution and for TASKSMODE timex type is "P1DT19:30".

.NET/Microsoft.Recognizers.Text.DataTypes.TimexExpression/TimexParsing.cs
Neelisha-saxena: assigned  indexOfP var with indexOfT. And pushed the change.

.NET/Microsoft.Recognizers.Definitions.Common/Dutch/DateTimeDefinitions.cs
Neelisha-saxena: To have the flexibility to recognize "every day at 5pm" as set type even if "every day" is not at starting of sentence.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
Neelisha-saxena: cases like each month, each year

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
Neelisha-saxena: DayEveryweek function is added to support cases like "19th of every month"

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: ParserDayEveryWeek is used for handling example such as 19th of every month. Added comment describing it.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: it's for parsing input text 15 of every month.

.NET/Microsoft.Recognizers.Text.DateTime/English/Extractors/EnglishSetExtractorConfiguration.cs
Neelisha-saxena: Only for the English set extractor do I want to use DayRegex instead of BeforeEachDayRegex.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: replaced DateObject.Now with refdate.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: Replaced DateObject.Now with refdate.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: It will parse the value under tasksmode for the input text, such as daily morning.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
```diff
@@ -91,25 +112,105 @@ public virtual List<Token> TimeEveryday(string text, DateObject reference)
...
+
+            return ret;
+        }
+
+        // Handle cases like 19th of every month
+        public virtual List<Token> DayEveryweek(string text, DateObject reference)
+        {
+            var ret = new List<Token>();
+            var ers = this.config.DateExtractor.Extract(text, reference);
+            if (NumberRecognizer.RecognizeOrdinal(text, config.Culture).Count > 0)
```
tellarin, line 151: Ideally this should use a number recognizer part of the same config.

tellarin: Also, I didn't notice this before, but I'm worried about the behaviour in languages other than English. Can we hide all of this behind the new flag and add a code comment to highlight it? Than it would be safer for now.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -32,6 +32,8 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
 
             if (er.Type.Equals(ParserName, StringComparison.Ordinal))
             {
+                er.Text = er.Text.Trim();
```
tellarin, line 35: Why do we need this trim now?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -190,20 +224,146 @@ private DateTimeResolutionResult ParserTimeEveryday(string text, DateObject refD
...
 
             if (ers.Count != 1)
             {
                 return ret;
             }
 
             var afterStr = text.Replace(ers[0].Text, string.Empty);
-            var match = this.config.EachDayRegex.Match(afterStr);
+            var timeErs = this.config.TimeExtractor.Extract(afterStr, refDate);
+            var timeErs1 = this.config.TimePeriodExtractor.Extract(afterStr, refDate);
```
tellarin, line 282: A better name would be timePeriodErs.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -190,20 +224,146 @@ private DateTimeResolutionResult ParserTimeEveryday(string text, DateObject refD
...
+            var ret = new DateTimeResolutionResult();
+
+            List<ExtractResult> ers = null;
+            var success = false;
+
+            // remove key words of set type from text
+            var match = config.SetEachRegex.Match(text);
+            if (match.Success)
+            {
+                var trimmedText = text.Replace(match.Value, "this ");
```
tellarin, line 335: "this " can't be hardcoded here. Also not with a whitespace. This should at least be a constant and define a method to process the value, instead of directly calling a replace here.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -248,7 +407,25 @@ private DateTimeResolutionResult ParseEach(IDateTimeExtractor extractor, IDateTi
...
+                if ((config.Options & DateTimeOptions.TasksMode) != 0)
+                {
+                    if (match.Success)
+                    {
+                        pr.TimexStr = TasksModeSetHandler.TasksModeTimexIntervalExt(pr.TimexStr);
+                    }
+
+                    if (match.Groups["other"].Success)
+                    {
+                        pr.TimexStr = pr.TimexStr.Replace("P1", "P2");
```
tellarin, line 419: What does this replace mean? This should be a method with a clear name, and probably P1 and P2 should be constants.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+
+namespace Microsoft.Recognizers.Text.DateTime.Utilities
+{
+    public static class TasksModeSetHandler
+    {
+        public static DateTimeResolutionResult TasksModeResolveSet(ref DateTimeResolutionResult result, string innerTimex, DateTimeParseResult pr = null)
+        {
+            result.Timex = innerTimex;
+
+            result.FutureValue = result.PastValue = "Set: " + innerTimex;
```
tellarin, line 17: "Set: " + innerTimex
should be the output of a well-named method instead of direct string manipulation here.
This will make it easirt for future changes that modify the value generation in a single place.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+
+            if (pr != null)
+            {
+                DateTimeResolutionResult value = (DateTimeResolutionResult)pr.Value;
+                if (value.FutureValue != null)
+                {
+                    if (pr.TimexStr.EndsWith("WE"))
```
tellarin, line 24: "WE" should be a constant.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+
+        public static Dictionary<string, string> TasksModeGenerateResolutionSetParser(Dictionary<string, string> resolutionDic, string mod, string timex)
+        {
+            var res = new Dictionary<string, string>();
+
+            TasksModeAddAltSingleDateTimeToResolution(resolutionDic, TimeTypeConstants.DATETIMEALT, mod, res);
+            if (timex.StartsWith(TasksModeConstants.GeneralPeriodPrefix) && res.Count > 0)
+            {
+                var extracted = new Dictionary<string, string>();
+                TimexRegex.Extract("period", timex, extracted);
```
tellarin, line 45: Al these hardcoded values should be constants.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+            }
+
+            return res;
+        }
+
+        public static string TasksModeTimexIntervalExt(string timex)
+        {
+            if (timex.Contains(Constants.TimexFuzzyWeek))
+            {
+                timex = timex + "P1W";
```
tellarin, line 69: Ideally all timex generations should be parametrized method calls, not direct string manipulations. Also hardcoded values should isntead re-use constants.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+                var resKey = TimeTypeConstants.DATETIME;
+
+                if (!result.Timex.Contains(TasksModeConstants.TimeTimexPrefix))
+                {
+                    resKey = TimeTypeConstants.DATE;
+                }
+
+                var futureValue = refDate.AddDays(7);
+
+                if (DateTimeFormatUtil.FormatDate(futureValue).Equals(value.Substring(0, 10)) && result.Timex.StartsWith(TasksModeConstants.FuzzyYearAndWeek))
```
tellarin, line 192: Why these exact indexes for substring? It would be better if they are constants defined somewhere or calculated based on some language config.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+                {
+                    if (result.Timex.Contains(TasksModeConstants.TimeTimexPrefix))
+                    {
+                        if (DateTimeFormatUtil.FormatTime(refDate).CompareTo(value.Substring(11)) <= 0)
+                        {
+                            value = DateTimeFormatUtil.FormatDate(refDate) + " " + value.Substring(11);
```
tellarin, line 198: Same comment on value generation. Define well-named parametrized values.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+
+                DateObject resDate = refDate;
+                if (DateTimeFormatUtil.FormatTime(resDate).CompareTo(value) > 0)
+                {
+                    resDate = resDate.AddDays(1);
+                }
+
+                result.FutureResolution = new Dictionary<string, string>
+                {
+                    { TimeTypeConstants.DATETIME, DateTimeFormatUtil.FormatDate((DateObject)resDate) + " " + (string)value },
```
tellarin, line 235: Same.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,292 @@
...
+            {
+                MergedParserUtil.AddSingleDateTimeToResolution(resolutionDic, TimeTypeConstants.TIME, mod, res);
+            }
+        }
+
+        internal static string GetStartValue(Resolution timexRes)
+        {
+            switch (timexRes.Values[0].Timex)
+            {
+                case "TMO": return "06:00:00";
```
tellarin, line 283: Same.

Specs/DateTime/German/DateTimeModel.json
tellarin: Not sure what you mean by this, but like I said, we can't have cases that worked before now not working. This is a blocker. Sorry.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: replaced default values with constants

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: missed to removed it earlier. corrected it now.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: Replaced with proper method and code comment.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: renamed variable with timePeriodErs

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: value is language independent; it is datetime string, and specific indexing is for extracting only date part from datetime

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: I followed the convention of the SetHandler.cs ResolveSet method, and this step is only used once. The following method is extension for ResolveSet for TasksMode.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: refactored function in more generalized manner.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: included method for specific task.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
Neelisha-saxena: replaced the following line with config specific method.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: replaced with parameterized method.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
Neelisha-saxena: I have tested the behavior of test cases in other languages and all of them are behaving the same as English. But to be on safer side for now move DayEveryweek method under flag.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Parsers/ArabicSetParserConfiguration.cs
```diff
@@ -168,5 +168,12 @@ public bool GetMatchedUnitTimex(string text, out string timex)
 
         public string WeekDayGroupMatchString(Match match) => SetHandler.WeekDayGroupMatchString(match);
 
+        public string ReplaceValueInTextWithFutTerm(string text, string value)
+        {
```
tellarin, line 172: As this behaviour is the same everywhere, the method should be declared in a shared place and receive the term or terms as a parameter. Also, there may be more than one term in the list, so the method should act accordingly.

.NET/Microsoft.Recognizers.Definitions.Common/Portuguese/DateTimeDefinitions.cs
```diff
@@ -703,5 +703,9 @@ public static class DateTimeDefinitions
       public const string QuarterTypeRegex = @"(trimestral(mente)?)$";
       public const string SemiAnnualTypeRegex = @"(semestral(mente)?)$";
       public const string YearTypeRegex = @"(anual(mente)?)$";
+      public static readonly IList<string> FutureTerms = new List<string>
+        {
+            @"esea"
```
tellarin, line 708: There are two issues here. One, this term is not a "future" term, so we need a better name. 
Second, this value is incorrect in PT, it should be "esse", "essa", "este", "esta".

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishSetParserConfiguration.cs
tellarin, line 136: Still many hardcoded values.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
tellarin: We should still need to change the call to the ordinal recognizer. But if this is the only issue, we can merge.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseSetParser.cs
```diff
@@ -32,6 +32,8 @@ public DateTimeParseResult Parse(ExtractResult er, DateObject refDate)
 
             if (er.Type.Equals(ParserName, StringComparison.Ordinal))
             {
+                er.Text = er.Text;
```
tellarin, line 35: Why is this needed?

.NET/Microsoft.Recognizers.Text.DateTime/TasksModeConstants.cs
```diff
@@ -49,5 +49,39 @@ public static class TasksModeConstants
         // tasksmode specific date parser constant
         public const string NextWeekGroupName = "next week";
 
+        // constants
+        public const string PeriodDaySuffix = "P1D";
```
tellarin, line 53: Do we need to redefine many of these constants? It seems to me that many could reuse the standard one in Constants.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
tellarin: Add at least a `//@TODO` comment here that this should be changed.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,316 @@
...
+        public static Dictionary<string, string> TasksModeGenerateResolutionSetParser(Dictionary<string, string> resolutionDic, string mod, string timex)
+        {
+            var res = new Dictionary<string, string>();
+
+            TasksModeAddAltSingleDateTimeToResolution(resolutionDic, TimeTypeConstants.DATETIMEALT, mod, res);
+            if (timex.StartsWith(TasksModeConstants.GeneralPeriodPrefix) && res.Count > 0)
+            {
+                var extracted = new Dictionary<string, string>();
+                TimexRegex.Extract(TasksModeConstants.PeriodString, timex, extracted);
+                res.Add("intervalSize", extracted.TryGetValue("amount", out var intervalSize) ? intervalSize : string.Empty);
```
tellarin, line 46: Still lots of hardcoded strings here. "amount", "dateUnit", "intervalSize", "intervalType", etc.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
tellarin: So we shoud define constants for them and add a comment describing their value and its calculation.

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
tellarin: At least add a `//@TODO` comment here

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseSetExtractor.cs
Neelisha-saxena: added //@TODO comment.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: Replace the follwing statement ("Set: " + innerTimex;) with the method call and constant name for set.

.NET/Microsoft.Recognizers.Text.DateTime/English/Parsers/EnglishSetParserConfiguration.cs
Neelisha-saxena: Replaced with DateTimeDefinitions Unit MAP.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: Replaced Additional string values with constant.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Parsers/ArabicSetParserConfiguration.cs
Neelisha-saxena: I added functionality to replace a value with only one term from a list; I didn't want to define a new pattern, so I used an existing variable.

Neelisha-saxena: Replace function with one common method.

.NET/Microsoft.Recognizers.Text.DateTime/TasksModeConstants.cs
Neelisha-saxena: Created a new variable in the TasksModeConstant file, with the intention of keeping the value specific to tasksmode if it is changed in the future.

Neelisha-saxena: But for now, derived it's value from Constant File.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: created constants for start and end value index and added comments for the same.

.NET/Microsoft.Recognizers.Text.DateTime/Portuguese/Parsers/PortugueseSetParserConfiguration.cs
```diff
@@ -194,5 +194,7 @@ public bool GetMatchedUnitTimex(string text, out string timex)
         }
 
         public string WeekDayGroupMatchString(Match match) => SetHandler.WeekDayGroupMatchString(match);
+
+        public string ReplaceValueInTextWithFutTerm(string text, string value) => TasksModeSetHandler.ReplaceValueInTextWithFutTerm(text, value, DateTimeDefinitions.ThisTerms[0]);
```
tellarin, line 198: Like I mentioned before, it can be a list of terms. Not only the first in the list. Pass the list as parameter and loop it inside the method.

tellarin (changes requested):
Almost there. Just some small issues left.

.NET/Microsoft.Recognizers.Text.DateTime/Portuguese/Parsers/PortugueseSetParserConfiguration.cs
Neelisha-saxena: TasksModeSetHandler.cs file method ReplaceValueInTextWithFutTerm now takes list as parameter.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Parsers/ArabicSetParserConfiguration.cs
```diff
@@ -38,6 +39,8 @@ public class ArabicSetParserConfiguration : BaseDateTimeOptionsConfiguration, IS
         private static readonly Regex YearTypeRegex =
             new Regex(DateTimeDefinitions.YearTypeRegex, RegexFlags);
 
+        private static readonly List<string> FutureTerms = (List<string>)DateTimeDefinitions.FutureTerms;
```
tellarin, line 42: Didn't we agree not to call it FutureTerms? Well, this is not a blocker to merge. But if there are other fixes to do, please rename this and the corresponding method.

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
```diff
@@ -0,0 +1,325 @@
...
+
+            }
+
+            return result;
+        }
+
+        public static string ReplaceValueInTextWithFutTerm(string text, string value, List<string> thisTermList)
+        {
+            value = value.Trim();
+            string thisTerm = thisTermList[0];
```
tellarin, line 270: Again, don't just use the first value in the list. If it has more than one, you should check more then one.

tellarin (changes requested):

.NET/Microsoft.Recognizers.Text.DateTime/Utilities/TasksModeSetHandler.cs
Neelisha-saxena: added comment as per our discussion.

.NET/Microsoft.Recognizers.Text.DateTime/Arabic/Parsers/ArabicSetParserConfiguration.cs
Neelisha-saxena: renamed in every file.

tellarin (approved):

PR #3015, opened 2022-08-02
[Python | EN DateTimeV2] Extend base date detection to more closely match .NET implementation

kevinwalshgen:
This is just a small update to match the .NET implementation and fix the detection of "Monday 21st" / "Monday 21" / etc.


tellarin (changes requested):
Looks reasonable. But can you also enable a relevant spec test in Python to show it behaves the same as in .NET?
All PRs with behaviour change should have corresponding spec cases to show the new implementation works as expected and to be used to avoid future regressions.
Thanks.

kevinwalshgen:
So went down a bit of a rabbit hole here as I noticed the original change didn't resolve anything at the Model test level, so fixed for extractor and went through running all the python tests to see which ones were actually passing now.

tellarin:
To be reviewed soon.

tellarin (approved):

Issue #3022, opened 2022-08-11
recognize datatime chinese


MarshallBruce-XC:
Hi , when i use DateTimeRecognizer by Culture-chinese version, there is type of sentence extracted sad result.
Ex: 
```
m = DateTimeRecognizer(Culture.Chinese).get_datetime_model()
text = "21年每月各物业公司的业扩报装量是多少"
res = m.parse(query = text )
for r in res:

    print(r.text)
    print(r.start)
    print(r.end)
    print(r.type_name)
    print(r.resolution)
output is :
21年
0
2
datetimeV2.duration
{'values': [{'timex': 'P21Y', 'type': 'duration', 'value': '662256000'}]}
每月
3
4
datetimeV2.set
{'values': [{'timex': 'P1M', 'type': 'set', 'value': 'not resolved'}]
```
That i thought is "21年" should means 2021 a daterange.  Just like this `{'values': [{'timex': '2021', 'type': 'daterange', 'start': '2021-01-01', 'end': '2022-01-01'}]}`
Can you optimize it or can you tell me where to change the code to achieve this effect.
Thanks a lot !!!!

PR #3044, opened 2022-11-04
[ZH Number] Add AmbiguityFiltersDict for cases like "伍"

felaray:
Add more case.


chexi-ms (approved):

felaray:
New commit for fix #3022

PR #3045, opened 2022-11-07
Adding a parametric matching timeout to all Regex instances  

fxs130430:
We are introducing an extra input parameter to specify a timeout value (in seconds) such that all RegEx objects return empty once the timeout event passes. We have seen a lot of cases (esp. with Japanese) where depending on the query, a RegEx object runs for as long as 30 hours before it eventually returns. To avoid meaningless computing and wasting resources, we allow each of the four main types (i.e., DateTime, Number, NumberWithUnit, Sequence) to be initialized with an optional time-out parameter.


.NET/Microsoft.Recognizers.Text.Choice/Arabic/Extractors/ArabicBooleanExtractorConfiguration.cs
```diff
@@ -43,5 +45,7 @@ public ArabicBooleanExtractorConfiguration(bool onlyTopMatch = true)
         public int MaxDistance => 2;
 
         public bool OnlyTopMatch { get; }
+
+        protected static TimeSpan RegexTimeOut => ChoiceRecognizer.GetTimeout(MethodBase.GetCurrentMethod().DeclaringType);
```
mhurst-microsoft, line 49: Would introducing an abstract class allow you to remove this line from all the classes? So, instead of IBooleanExtractorConfiguration you have AbstractBooleanExtractorConfiguration which implements the interface and provides the RegexTimeout.

fxs130430: Nice suggestion. The refactoring done.

aurghob:
Hi,

Initial comments:
1. Is it possible to add more description for the PR?
My take from the commits:
a. Removed dotnet 4.62
 b. Added base timeout functionality (via BaseBooleanExtractorConfiguration) and incorporated it into existing regex implementation.
3. Is the test failure(below) in the pipeline (https://ci.appveyor.com/project/tellarin/recognizers-text/builds/46020012) related to this change?
Failed!  - Failed:     1, Passed:   282, Skipped:     0, Total:   283, Duration: 710 ms - Microsoft.Recognizers.Text.DataTypes.DataDrivenTests.dll (netcoreapp3.1)
4. This will not be a breaking change for projects that use this dependency right?
5. For people working on this repo in the future,
Could you also add how you tested your change?
My assumption is you followed https://blog.botframework.com/2018/02/12/contributing-luis-microsoft-recognizers-text-part-3/, right?

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -189,5 +187,19 @@ protected override void InitializeConfiguration()
                         new KoreanMergedExtractorConfiguration(new BaseDateTimeOptionsConfiguration(Culture.Korean, options)))));
```
aurghob, line 187: Are the korean and arabic tests passing ? Or would you prefer to leave this out of scope?

aurghob (commented):
minor comments - I can see most of the code change is due to the refactor.

.NET/Microsoft.Recognizers.Text.DateTime/DateTimeRecognizer.cs
```diff
@@ -12,33 +11,32 @@
...
 using Microsoft.Recognizers.Text.DateTime.Turkish;
+using Microsoft.Recognizers.Text.DateTime.Utilities;
 
 namespace Microsoft.Recognizers.Text.DateTime
 {
     public class DateTimeRecognizer : Recognizer<DateTimeOptions>
     {
-        public DateTimeRecognizer(string targetCulture, DateTimeOptions options = DateTimeOptions.None, bool lazyInitialization = false)
-            : base(targetCulture, options, lazyInitialization)
+        public DateTimeRecognizer(string targetCulture, DateTimeOptions options = DateTimeOptions.None, bool lazyInitialization = false, int timeoutInSeconds = 0)
```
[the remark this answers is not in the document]
fxs130430: zero as default results in no timeout which is the existing behavior. As far as the scale, I think one minute is a fair game. That's why I thought milliseconds will be too small of a scale.

fxs130430:
> Hi,
> 
> Initial comments:
> 
> 1. Is it possible to add more description for the PR?
>    My take from the commits:
>    a. Removed dotnet 4.62
>    b. Added base timeout functionality (via BaseBooleanExtractorConfiguration) and incorporated it into existing regex implementation.
> 2. Is the test failure(below) in the pipeline (https://ci.appveyor.com/project/tellarin/recognizers-text/builds/46020012) related to this change?
>    Failed!  - Failed:     1, Passed:   282, Skipped:     0, Total:   283, Duration: 710 ms - Microsoft.Recognizers.Text.DataTypes.DataDrivenTests.dll (netcoreapp3.1)
> 3. This will not be a breaking change for projects that use this dependency right?
> 4. For people working on this repo in the future,
>    Could you also add how you tested your change?
>    My assumption is you followed https://blog.botframework.com/2018/02/12/contributing-luis-microsoft-recognizers-text-part-3/, right?

1 . Removing net462 was a mistake. We only remove it due to a technicality during the build process.
2. All tests passed.
3. It won't be breaking because the timeout is optional input and the default value zero means no timeout is enforced.
4. I did not provide any extra tests as the current test cases serve the purpose. I would suggest that we add an integration test with the problematic Japanese query that started all of this.

aurghob (approved):

PR #3048, opened 2022-11-23
[NL DateTimeV2] Dutch datetime support in Python

samhickey25:


tellarin (changes requested):
Looks mostly good. 
But why the change to regex PmRegex? Also, DateTimeModel.json must have its cases enabled in order to merge. The specs in this PR show that extractors/parsers work separately, but model-level tests must also pass.
BTW, any insights why some tests that pass in .NET fail in Python with these changes?

samhickey25:
> Looks mostly good. But why the change to regex PmRegex? Also, DateTimeModel.json must have its cases enabled in order to merge. The specs in this PR show that extractors/parsers work separately, but model-level tests must also pass. BTW, any insights why some tests that pass in .NET fail in Python with these changes?

I've updated the DateTimeModel tests to run against python now.
The PmRegex change was because it was matching on weekdays like Dinsdag which isn't right.

tellarin (approved):

PR #3061, opened 2022-12-22
[ES DateTimeV2] Support for <day> del <month>

samhickey25:


tellarin (changes requested):
Looks OK. But we need at least one test spec of the new case at the model level (DateTimeModel.json) to guarantee the final output is correct and check for regressions.

tellarin (approved):

PR #3063, opened 2023-01-03
[EN DateTimeV2] Support for [minutes] past [hour]

andrew-gradinari:
Support for the time expression “[minutes] past [hour]”.
This is a popular way of describing time in many English-speaking countries. For example, if someone says "It's 10 past 3" they are saying that the time is 3:10.


andrew-gradinari:
@microsoft-github-policy-service agree [company="Genesys"]

tellarin (changes requested):
Similar comments to PR #3067.

tellarin (approved):

PR #3057, opened 2022-12-15
[* DateTimeV2] Two digit year improvements

Conor-Keaney:
Fixes cases and added testing for "Month YY" e.g. "March 99" for English, Dutch, Spanish, German, French, Italian and Portuguese

PR #3070, opened 2023-01-27
Single quote before 2 digit year not being recognized

Conor-Keaney:


Conor-Keaney:
@tellarin Hi Borje, I'm having an issue with this fix and was wondering if you would be more familiar with .NET.

While the regexs I changed in https://github.com/microsoft/Recognizers-Text/pull/3057 are working for .NET and python, the changes in this PR to those same regexs don't seem to work for .NET. 
Testing this on a .NET regex tester show they should be working.

Any suggestions or direction would be greatly appreciated - thanks!

Conor-Keaney:
Hello @tellarin, would you be able to review this pull request whenever suits? Thanks!

Conor-Keaney:
Hi @tellarin, would you be able to give this a review whenever possible please, thanks!

PR #3083, opened 2023-03-24
[Infra] Fixing component governmence vulnarabilities and OneBranch changes

fxs130430:
- Updating some packages to address the Azure OneBranch component governance errors.
- Adding the packages metadata to the .csproj files. This was needed so that the dotnet pack command that will be used in the new onebranch pipeline adds the same metadata that the packages contain today.


tellarin (approved):
LGTM

PR #3086, opened 2023-03-28
[Infra] Adding .net6.0 support to the packages

fxs130430:
In this PR I'm adding the support for .NET6.0 target framework and make the necessary changes to the AppVeyor (PR build pipeline environment) yml file to accommodate .NET6.0 build and also to speed up the pipeline execution. There is a 1-hour limitation; the pipeline should finish building all languages (.net, java, js, python) in less than that. Addition of .NET6.0 would prolong the execution and fail. Previously, the pipeline was using Python3.7 to run the tests. I upgraded it to Python3.11 which significantly performs better. But, we got 2 failures in the tests due to different handling of implicit type casts (i.e., decimal to int) in Python3.7 vs. Python3.11 (or something in pytest).   

Also, adding .NET6.0 broke a few unit tests due to using different characters in Swedish and Arabic for Negative number sign (i.e., "-").

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,7 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <packageSources>
-    <clear />
-    <add key="luis-common-upstream" value="https://pkgs.dev.azure.com/msazure/a531312f-ba7b-4573-ab4f-d468b740a319/_packaging/luis-common-upstream/nuget/v3/index.json" />
-  </packageSources>
-</configuration>
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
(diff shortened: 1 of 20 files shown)


NuGet.Config
fxs130430, line 1: Needed to address OneBranch secure supply chain policy. It is removed in the build script of the PR pipeline to allow the default nuget feed to be used.

Python/libraries/datatypes-timex-expression/datatypes_timex_expression/timex_helpers.py
```diff
@@ -164,9 +164,9 @@ def timex_datetime_add(start, duration):
     @staticmethod
     def date_from_timex(timex):
         return date(
-            timex.year if timex.year is not None else 2001,
-            timex.month if timex.month is not None else 1,
-            timex.day_of_month if timex.day_of_month is not None else 1
+            int(timex.year) if timex.year is not None else 2001,
+            int(timex.month) if timex.month is not None else 1,
+            int(timex.day_of_month) if timex.day_of_month is not None else 1
```
fxs130430, line 169: This is needed to avoid python3.11 interpreter type errors

aurghob (approved):
Approving - all changes seem necessary here. 
Discussed offline that framework changes will be made in the package branch.

PR #3147, opened 2024-08-09
Fix Recognizers cannot recognize next next week/weekend in Chinese

MichaelMWW:
Fix Recognizers cannot recognize 下下周(next next week), 下下周末(next next weekend) in Chinese.


.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
```diff
@@ -77,7 +77,7 @@ public static class DateTimeDefinitions
       public const string DatePeriodDayRegexInCJK = @"(?<day>初一|三十|一日|十一日|二十一日|三十一日|二日|三日|四日|五日|六日|七日|八日|九日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|一日|十一日|十日|二十一日|二十日|三十一日|三十日|二日|三日|四日|五日|六日|七日|八日|九日|十二日|十三日|十四日|十五日|十六日|十七日|十八日|十九日|二十二日|二十三日|二十四日|二十五日|二十六日|二十七日|二十八日|二十九日|十日|二十日|三十日|10日|11日|12日|13日|14日|15日|16日|17日|18日|19日|1日|20日|21日|22日|23日|24日|25日|26日|27日|28日|29日|2日|30日|31日|3日|4日|5日|6日|7日|8日|9日|一号|十一号|二十一号|三十一号|二号|三号|四号|五号|六号|七号|八号|九号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|一号|十一号|十号|二十一号|二十号|三十一号|三十号|二号|三号|四号|五号|六号|七号|八号|九号|十二号|十三号|十四号|十五号|十六号|十七号|十八号|十九号|二十二号|二十三号|二十四号|二十五号|二十六号|二十七号|二十八号|二十九号|十号|二十号|三十号|10号|11号|12号|13号|14号|15号|16号|17号|18号|19号|1号|20号|21号|22号|23号|24号|25号|26号|27号|28号|29号|2号|30号|31号|3号|4号|5号|6号|7号|8号|9号|一|十一|二十一|三十一|二|三|四|五|六|七|八|九|十二|十三|十四|十五|十六|十七|十八|十九|二十二|二十三|二十四|二十五|二十六|二十七|二十八|二十九|一|十一|十|二十一|二十|三十一|三十|二|三|四|五|六|七|八|九|十二|十三|十四|十五|十六|十七|十八|十九|二十二|二十三|二十四|二十五|二十六|二十七|二十八|二十九|十|二十|三十|廿|卅)";
       public const string DatePeriodThisRegex = @"这个|这一个|这|这一|本";
       public const string DatePeriodLastRegex = @"上个|上一个|上|上一";
-      public const string DatePeriodNextRegex = @"下个|下一个|下|下一";
```
aurghob, line 80: Can you please not remove this regex and add a new regex for the new one instead?

Patterns/Chinese/Chinese-DateTime.yaml
```diff
@@ -148,7 +148,7 @@ DatePeriodThisRegex: !simpleRegex
 DatePeriodLastRegex: !simpleRegex
   def: 上个|上一个|上|上一
 DatePeriodNextRegex: !simpleRegex
-  def: 下个|下一个|下|下一
+  def: 下下|下个|下一个|下|下一
```
aurghob, line 151: Will "下个|下一个|下|下一" detection fail with this change ?

MichaelMWW: No, they will not fail since it will try to match "下下" first.

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
MichaelMWW: Do you mean add a new separate regex for "下下"? This cs file is auto generated from the yaml file, we need to add it in yaml file.

Patterns/Chinese/Chinese-DateTime.yaml
tellarin: Can you add a spec with a test case for it too, to make sure no regression will happen?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
```diff
@@ -1073,7 +1073,7 @@ private DateTimeResolutionResult ParseOneWordPeriod(string text, DateObject refe
                     var swift = 0;
                     if (nextMatch.Success)
                     {
-                        if (nextMatch.Groups[Constants.AfterGroupName].Success)
+                        if (nextMatch.Groups[Constants.AfterGroupName].Success || nextMatch.Value.Equals("下下", StringComparison.Ordinal))
```
tellarin, line 1076: Why is this needed? It would be better to define a constant in the YAML for the value with a good name, instead of hardcoding a substring here.

tellarin: A code comment here explaining why this is needed would also be very useful.

Patterns/Chinese/Chinese-DateTime.yaml
MichaelMWW: > Can you add a spec with a test case for some of the other cases too, if they don't exist, to make sure no regression will happen?

Yes, I have added some test cases, all the new tests and the old test cases in TestDateTime_Chinese passed.

.NET/Microsoft.Recognizers.Definitions.Common/Chinese/DateTimeDefinitions.cs
MichaelMWW: Add new regex for this scenario.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/CJK/BaseCJKDatePeriodParser.cs
MichaelMWW: > Why is this needed? It would be better to define a constant in the YAML for the value with a good name, instead of hardcoding a substring here.

Thanks for the comments, finish updating the code.

tellarin (changes requested):
This looks reasonable. But to make sure all works correctly, you also need to add specs at the Model level in DateTimeModel.json

BTW, do you know if similar cases happen in Japanese or Korean?

tellarin:
@MichaelMWW Have you signed the CLA? The auto check says it's still pending too. Thanks.

MichaelMWW:
> @microsoft-github-policy-service agree company="Microsoft"

> @MichaelMWW please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

MichaelMWW:
@microsoft-github-policy-service agree company="Microsoft"

MichaelMWW:
> This looks reasonable. But to make sure all works correctly, you also need to add specs at the Model level in DateTimeModel.json
> I notice that in DateTimeModel.json, there are also some other specs lacking, like 下周(next week), are all test specs for datatime need to add specs in Model level?
> BTW, do you know if similar cases happen in Japanese or Korean?
> The bug is just for Chinese. I have done some research, in Japanese or Korean, they both have similar saying which we may handle in the future.

MichaelMWW:
> @MichaelMWW Have you signed the CLA? The auto check says it's still pending too. Thanks.
Finished now.

aurghob (approved):

tellarin (approved):

PR #3151, opened 2024-08-15
Fix wrong resolution for some fractions

MichaelMWW:
Fix wrong resolution for some fractions, scenario as below.
“two hundred and fifty-four hundredths” -> 200.54 (correctly found)
“two and fifty-four hundredths” -> 0.0037037037… (it calculates (2 / 54) div by 100) instead of 2.54

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.xml
+++ b/.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.xml
@@ -167,6 +167,15 @@
             <param name="extResult">input arabic number.</param>
             <returns>parsed result.</returns>
         </member>
+        <member name="M:Microsoft.Recognizers.Text.Number.BaseNumberParser.GetSplitIndex(System.Collections.Generic.List{System.String})">
+            <summary>
+            Get the split index for a fraction word list, split index used to separate the numerator and the denominator.
+            Ex: A fraction is "three fifth", it will be joined as a list which 1st item is "three" and 2nd item is "fifth", the split index is 1 (index of fifth).
+            Ex: A fraction is "two and fifty-four hundredths", the split index is 3 (index of hundredths).
+            </summary>
+            <param name="fracWords">fraction words list.</param>
+            <returns>split index.</returns>
+        </member>
         <member name="M:Microsoft.Recognizers.Text.Number.INumberParserConfiguration.NormalizeTokenSet(System.Collections.Generic.IEnumerable{System.String},Microsoft.Recognizers.Text.ParseResult)">
             <summary>
             Used when requiring to normalize a token to a valid expression supported by the ImmutableDictionaries (language dictionaries).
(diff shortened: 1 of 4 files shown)


.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -999,6 +934,85 @@ protected double GetPointValue(List<string> matchStrs)
...
+                {
+                    continue;
+                }
+
+                var previousValue = currentValue;
+                currentValue = Config.ResolveCompositeNumber(fracWords[splitIndex]);
+
+                var hundredsSM = 100;
+
+                // Below flag isUncomposobleWithSeparator is used to handle one existing bug for handling fraction input like "two and fifty-four hundredths".
```
aurghob, line 955: Can you change explanation to what it currently is rather than ask to compare with previous code ? Explanation should be about code and not ask the person reading code to compare with previous code according to me.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
```diff
@@ -999,6 +934,85 @@ protected double GetPointValue(List<string> matchStrs)
             return ret;
         }
 
+        protected int GetSplitIndex(List<string> fracWords)
```
aurghob, line 937: Change to private please

aurghob: Could you also add documentation to this method instead? Will help explain what split index is

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
MichaelMWW: Updated comments, please help to check again.

.NET/Microsoft.Recognizers.Text.Number/Parsers/BaseNumberParser.cs
MichaelMWW: Done.

.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.xml
```diff
@@ -167,6 +167,15 @@
             <param name="extResult">input arabic number.</param>
             <returns>parsed result.</returns>
         </member>
+        <member name="M:Microsoft.Recognizers.Text.Number.BaseNumberParser.GetSplitIndex(System.Collections.Generic.List{System.String})">
+            <summary>
+            Get the split index for a fraction word list, split index used to separate the numerator and the denominator.
```
aurghob, line 172: Is this still needed in the xml now that the method is private?

aurghob (approved):
Minor comment. But LGTM

.NET/Microsoft.Recognizers.Text.Number/Microsoft.Recognizers.Text.Number.xml
MichaelMWW: If your method has documentation comments, it will automatically add member in the xml file when build.

PR #3152, opened 2024-08-20
Fix 2 years ago in French not recognized

MichaelMWW:
Fix 2 years ago in French not recognized.
"il y a deux ans" e.g. "two years ago" is not being detected in French.

### DIFF
--- a/Patterns/French/French-DateTime.yaml
+++ b/Patterns/French/French-DateTime.yaml
@@ -76,7 +76,7 @@ MonthSuffixRegex: !nestedRegex
   def: (?<msuf>(en\s*|le\s*|de\s*|dans\s*)?({RelativeMonthRegex}|{WrittenMonthRegex}))
   references: [ RelativeMonthRegex, WrittenMonthRegex ]
 DateUnitRegex: !simpleRegex
-  def: (?<unit>an(s)?|(?<plural>mois)|((l')?ann[eé]e|semaine|journ[eé]e|jour)(?<plural>s)?)\b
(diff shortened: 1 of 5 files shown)


Patterns/French/French-DateTime.yaml
```diff
@@ -479,12 +479,12 @@ AgoPrefixRegex: !simpleRegex
 LaterRegex: !simpleRegex
   def: \b(plus\s+tard|à\s+partir\s+(de\s+(maintenant|demain)|d'aujourd'hui)|après\s+(aujourd'hui|demain))\b
 AgoRegex: !nestedRegex
-  def: \b((depuis|il\s+y\s*a)(\s+{AroundRegex})?|auparavant|avant\s+(?<day>hier|aujourd'hui))\b
+  def: \b((il\s+y\s*a)(\s+{AroundRegex})?|auparavant|avant\s+(?<day>hier|aujourd'hui))\b
   references: [ AroundRegex ]
 BeforeAfterRegex: !simpleRegex
   def: ^\b$
 InConnectorRegex: !simpleRegex
-  def: \b(dans|en|sur)\b
```
aurghob, line 487: Why did we remove sur?

MichaelMWW: > Why did we remove sur?

InConnectorRegex is used to extract the connector "in" in cases like "In 3 days/weeks/months/years" which will be recognized as date. 
"sur 3 ans" means "over 3 year" and it expected to be a duration, "sur" means "over" and not in. If we keep the sur here, it will be recognized as date.

aurghob (approved):

PR #3153, opened 2024-08-30
Add support for French decade with century

MichaelMWW:
Add support for recognizing French decade with century. It now supports Arabic numbers and language specific number words, examples as below.
1. dans les années 90
2. les années quatre-vingt-dix


tellarin (changes requested):
Don't forget to always add (or remove the "NotSupportted" attribute) samples for the cases directly affected by the PR to the corresponding DateTimeModel.json. This will ensure the final output is indeed correct and in the correct format.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -43,9 +43,9 @@ public static class DateTimeDefinitions
       public const string WrittenOneToNineRegex = @"(?:une?|deux|trois|quatre|cinq|six|sept|huit|neuf)";
       public const string WrittenElevenToNineteenRegex = @"(?:(seize|quinze|quatorze|treize|douze|onze)|dix\W(neuf|huit|sept))";
       public const string WrittenTensRegex = @"(?:quatre\Wvingt(s|\Wdix)?|soixante(\Wdix)?|dix|vingt|trente|quarante|cinquante|septante|octante|huitante|nonante)";
-      public static readonly string WrittenCenturyFullYearRegex = $@"(?:(deux\s+)?mille((\s+{WrittenOneToNineRegex})?\s+cents?)?)";
+      public static readonly string WrittenCenturyFullYearRegex = $@"(?:(deux\s+)?mille((\s+{WrittenOneToNineRegex})?\s+cent?)?)";
```
tellarin, line 46: Why this change? The s? is already saying the s is optional, useful for cases where you could have a duration of multiple centuries.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -720,11 +721,20 @@ public static class DateTimeDefinitions
       public const string NightRegex = @"\b(minuit|nuit)\b";
       public static readonly Dictionary<string, int> WrittenDecades = new Dictionary<string, int>
         {
-            { @"", 0 }
+            { @"cent", 0 },
```
tellarin, line 724: Why is this falling in the decades config?

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
MichaelMWW: > Why this change? The s? is already saying the s is optional, useful for cases where you could have a duration of multiple centuries.

In French, when express century and decade, cents is incorrect, it should be cent.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
MichaelMWW: In other language like English, the hundreds is also in the WrittenDecades.

MichaelMWW:
> Don't forget to always add (or remove the "NotSupportted" attribute) samples for the cases directly affected by the PR to the corresponding DateTimeModel.json. This will ensure the final output is indeed correct and in the correct format.

OK.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
MichaelMWW: > > Why this change? The s? is already saying the s is optional, useful for cases where you could have a duration of multiple centuries.
> 
> In French, when express century and decade, cents is incorrect, it should be cent.

Since s is optional, it should be fine to keep it, I will revert the change.

.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
MichaelMWW: The "cent" entry is unnecessary, I will remove it.

MichaelMWW: > In other language like English, the hundreds is also in the WrittenDecades.

The "cent" entry is unnecessary, I will remove it.

aurghob (approved):
LGTM

tellarin (approved):

PR #3154, opened 2024-09-04
Fix the issue with parsing sentences expressing the decade with century and number with unit in Spanish

MichaelMWW:
Fix the issue with parsing sentences expressing the decade with century in Spanish.
Below sentences cannot be parsed correctly before, now it works with accents or without accents.
1. escrito en los años/anos 90
2. los años/anos setenta
3. ¿Cuáles fueron los presidentes de Estados Unidos en la década/decada de 1990?

Fix the issue that it cannot recognize number with unit without accents, both "Tengo diez anos" and "Tengo diez años" work now.

### DIFF
--- a/Specs/DateTime/Spanish/DatePeriodParser.json
+++ b/Specs/DateTime/Spanish/DatePeriodParser.json
@@ -3908,11 +3908,11 @@
     "Context": {
       "ReferenceDateTime": "2016-11-07T00:00:00"
     },
-    "NotSupported": "dotnet, java, python",
+    "NotSupported": "java, python",
     "NotSupportedByDesign": "javascript, python",
     "Results": [
       {
-        "Text": "años 1970",
+        "Text": "los años 1970",
         "Type": "daterange",
         "Value": {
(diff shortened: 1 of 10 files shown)


Specs/DateTime/Spanish/DatePeriodParser.json
```diff
@@ -4070,11 +4070,11 @@
     "Context": {
```
aurghob, line 4070: Does this also work with los anos?

aurghob (commented):
Please make sure to add tests that work without some of the accents as well please. Could you please go back and make the fix. I understand there might be many - lets start with the few we want fixed and a few more for now. We can come back and fix more if needed.

aurghob (changes requested):
self requested review by mistake

tellarin (approved):

Specs/DateTime/Spanish/DatePeriodParser.json
MichaelMWW: > Does this also work with los anos?

Adding support for words without accents, please review again.

aurghob (approved):

PR #3174, opened 2024-10-16
[EN DateTimeV2] Add support for time range with duration

MichaelMWW:
Add support for time range with duration, supported patterns as below.
- from 6 am for 2 hours
- for 2 hours from 6 am
- 16th Oct from 6 am for 2 hours
- 16th Oct for 2 hours from 6 am


.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
```diff
@@ -153,6 +154,22 @@ private List<Token> MatchSimpleCases(string text)
             return ret;
         }
 
+        // Cases like "from 6am for 3 hours" and "for 3 hours from 6 am" are extracted as timerange here.
+        private List<Token> MatchTimePeriodWithDurationCases(string text)
```
aurghob, line 158: Is this code being used by an english implementation class ?

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
```diff
@@ -476,6 +477,12 @@ private DateTimeResolutionResult ParseBaseDatePeriod(string text, DateObject ref
                 innerResult = ParseDatePointWithAgoAndLater(text, referenceDate);
             }
 
+            // Cases like "for x weeks/days from today/12 sep etc."
+            if (!innerResult.Success)
+            {
+                innerResult = ParseDatePointWithForPrefix(text, referenceDate);
```
aurghob, line 483: Can this be moved to the english implementation if it exists ?

aurghob (approved):

.NET/Microsoft.Recognizers.Text.DateTime/Extractors/BaseTimePeriodExtractor.cs
MichaelMWW: As we talked, will consider it in next sprint.

.NET/Microsoft.Recognizers.Text.DateTime/Parsers/BaseDatePeriodParser.cs
MichaelMWW: As we talked, will consider it in next sprint.

PR #3175, opened 2024-10-24
[DE DateTimeV2] Fix occasionally NullReferenceException in German DateTimeModel test when input is tag der Arbeit

MichaelMWW:
Fix occasionally NullReferenceException in German DateTimeModel test when input is tag der Arbeit, the test and message as below.

Test:
Microsoft.Recognizers.Text.DateTime.Tests.TestDateTime_German DateTimeModel 

Message: 
Test method Microsoft.Recognizers.Text.DateTime.Tests.TestDateTime_German.DateTimeModel threw exception: 
System.ApplicationException: Input: "Tag der Arbeit, ersten Weihnachtsfeiertag, zweiten Weihnachtstag" ---> System.NullReferenceException: Object reference not set to an instance of an object.

### DIFF
--- a/.NET/Microsoft.Recognizers.Text.DateTime/German/Parsers/HolidayParserGer.cs
+++ b/.NET/Microsoft.Recognizers.Text.DateTime/German/Parsers/HolidayParserGer.cs
@@ -64,6 +64,7 @@ namespace Microsoft.Recognizers.Text.DateTime.German
                 { "stgeorgeday", StGeorgeDay },
                 { "mayday", Mayday },
                 { "labour", LaborDay },
+                { "laborday", LaborDay },
                 { "cincodemayoday", CincoDeMayo },
                 { "baptisteday", BaptisteDay },
                 { "usindependenceday", UsaIndependenceDay },


tellarin (commented):
This issue only happened sometimes and this fixes it? That's strange.
The corresponding test case is already in DateTimeModel?

MichaelMWW:
> This issue only happened sometimes and this fixes it? That's strange. The corresponding test case is already in DateTimeModel?

The reason for the occasionally fail is that in line of code https://github.com/microsoft/Recognizers-Text/blob/master/.NET/Microsoft.Recognizers.Text.DateTime/German/Parsers/HolidayParserGer.cs#L578, there is a dictionary **this.config.HolidayNames**, it has  below 3 items that contains **tagderarbeit**.

- labour: [ tagderarbeit ]
- mayday: [ mayday, maitag, tagderarbeit, maifeiertag ]
- laborday: [ mayday, maitag, tagderarbeit, maifeiertag ]

Then it will check the value which contains tagderarbeit from **this.config.HolidayNames** and return the key as holidayKey. So the holidayKey value can be labour/mayday/laborday, it will use the key to map from another dictionary FixedHolidaysDict. In FixedHolidaysDict, it lacks laborday, so it will leads to failure when holidayKey is laborday. When holidayKey is labour/mayday, it works fine.

aurghob (approved):

PR #3178, opened 2024-11-08
[EN DateTimeV2] Support [duration]+starting+[datetime] pattern as date range or date time range

MichaelMWW:
Support duration starting date time pattern as date range or date time range.

- 2 weeks starting date/datetime
- 2 days starting date/datetime


aurghob (approved):
Please look into these ?

MichaelMWW:
> Please look into open comments next ?

Which comments?

PR #3185, opened 2024-11-21
[ZH Email] Fix email address not be recognized when there are some Chinese characters in the sentence.

MichaelMWW:
Fix email address not be recognized when there are some Chinese characters in the sentence.
Originally, the recognizer cannot recognize the email address "test@test.com" in sentence "1邮件地址test@test.com", now fix it.

### DIFF
--- a/Specs/Sequence/English/EmailModel.json
+++ b/Specs/Sequence/English/EmailModel.json
@@ -72,6 +72,18 @@
     ]
   },
   {
+    "Input": "a邮箱地址test@test.com",
+    "Results": [
+      {
+        "Text": "test@test.com",
+        "TypeName": "email",
+        "Resolution": {
+          "value": "test@test.com"
+        }
+      }
+    ]
+  },
+  {
     "Input": "Hello, @Carol, please write to me at Dave@abc.com for more information on task #A1",
     "Results": [
       {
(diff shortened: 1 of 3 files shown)


Specs/Sequence/English/EmailModel.json
```diff
@@ -142,13 +154,13 @@
   },
   {
     "Input": "Both a..bc@outlook.com and .abc@hotmail.com are not valid e-mail addresses.",
-	"Comment": "By default the current system is strict. If a relaxed match is needed (to catch these), enable the Relaxed option.",
+    "Comment": "By default the current system is strict. If a relaxed match is needed (to catch these), enable the Relaxed option.",
```
bidisha-c, line 157: what is this "Relaxed" option? I am thinking if we should enable this with the relaxed option itself?

bidisha-c: My proposal was not to enable these settings. Instead to enable the detection of email address not be recognized when there are some Chinese characters in the sentence as a relaxed option instead of by default

MichaelMWW: > My proposal was not to enable these settings. Instead to enable the detection of email address not be recognized when there are some Chinese characters in the sentence as a relaxed option instead of by default

Relaxed option will not solve the issue, the issue is caused by the regex that I am updated in this PR.
There are 2 regex used to extract email, in case there is a Chinese sentence **"1邮件地址test@test.com"**, the first regex matches **"test@test.com"**, the 2nd regex (the regex I have fixed in the PR) matches **"1邮件地址test@test.com"**, so the results merged as **"1邮件地址test@test.com"**, if relaxed option is enabled here, it will directly return **"1邮件地址test@test.com"** which is incorrect, the correct email should be "test@test.com"

MichaelMWW: Need to mention that currently we don't support RFC 6530 which allow non-ASCII characters in email address, such as Chinese, Arabic.

bidisha-c: I am not very familiar with this codebase but my idea was 

Using the relaxed option - run the regex without the word separator to return the correct detection for 邮件地址test@test.com along with all other existing anomalous behaviors like ending with a . or having multiple dots (..) in the email address and keep existing email detection intact so we do not break existing customers. 

That said, if this is indeed a bug fix, it should be ok to fix in the regular regex.

bidisha-c: The more that I think, I am ok with your approach.

bidisha-c (approved):

PR #3189, opened 2025-02-10
[FR DateTimeV2] Fix the issue of 'duration' in French being parsed as a time

MichaelMWW:
Fix the issue of 'duration' in French being parsed as a time.
In French, "2 heures" (two hours) can be used in different contexts, mainly for telling time and indicating duration.
For example, "2 heures" in "dure 2 heures" was originally detected as time "02:00", it is incorrect, it should be a 2 hours duration.

### DIFF
--- a/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
+++ b/.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
@@ -250,7 +250,7 @@ namespace Microsoft.Recognizers.Definitions.French
       public const string WeekWithWeekDayRangeRegex = @"^\b$";
       public const string GeneralEndingRegex = @"^\b$";
       public const string MiddlePauseRegex = @"^\b$";
-      public const string DurationConnectorRegex = @"^\b$";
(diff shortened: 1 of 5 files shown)


.NET/Microsoft.Recognizers.Definitions.Common/French/DateTimeDefinitions.cs
```diff
@@ -250,7 +250,7 @@ public static class DateTimeDefinitions
       public const string WeekWithWeekDayRangeRegex = @"^\b$";
       public const string GeneralEndingRegex = @"^\b$";
       public const string MiddlePauseRegex = @"^\b$";
-      public const string DurationConnectorRegex = @"^\b$";
+      public const string DurationConnectorRegex = @"^\s*(?<connector>\s+|et|,)\s*$";
```
bidisha-c, line 253: Why do we need to update the regex here?

MichaelMWW: The DurationConnectorRegex regex is used to handle hours followed with minutes and seconds.

aurghob (approved):

